PR #44, opened 2021-11-25
Fix/ Cluster Node Error

ArtyomPodymov:
#RI-2003 - catch the lastNodeError for the cluster if it exists


arthosofteq (approved):

arthosofteq:
@ArtyomPodymov Could you add unit tests for this changes?

PR #45, opened 2021-11-25
Fix/ Folder name for settings on prod

ArtyomPodymov:
allow defining APP FOLDER NAME by using env variable only for production

### DIFF
--- a/redisinsight/api/config/staging.ts
+++ b/redisinsight/api/config/staging.ts
@@ -1,6 +1,6 @@
 import { join } from 'path';
 
-const homedir = join(require('os').homedir(), '.redisinsight-v2.0-stage');
+const homedir = join(require('os').homedir(), process.env.APP_FOLDER_NAME || '.redisinsight-v2.0-stage');
 
 export default {
   dir_path: {
--- a/redisinsight/api/config/staging.ts
+++ b/redisinsight/api/config/staging.ts
@@ -1,6 +1,6 @@
 import { join } from 'path';
 
-const homedir = join(require('os').homedir(), process.env.APP_FOLDER_NAME || '.redisinsight-v2.0-stage');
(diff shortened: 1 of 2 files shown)


redisinsight/api/config/staging.ts
```diff
@@ -1,6 +1,6 @@
 import { join } from 'path';
 
-const homedir = join(require('os').homedir(), process.env.APP_FOLDER_NAME || '.redisinsight-v2.0-stage');
+const homedir = join(require('os').homedir(), '.redisinsight-v2.0-stage');
```
arthosofteq, line 3: I'm afraid that integration tests on docker images will fail then. 
Why we need such changes?

arthosofteq (changes requested):

redisinsight/api/config/staging.ts
ArtyomPodymov: Because at the moment the stage and prod builds both refer to the same folder (.redisinsight-preview) inside homedir.  Or we can add the additional env variable for stage (APP_STAGE_FOLDER_NAME)

arthosofteq: > 

We should handle it elsewhere. 
For example: 
- in the circleci.yml (like it is done for autoupdate)
- or we can remove environment variable from the circleci and change defaults for each config file (dev, stage, prod)

ArtyomPodymov: For now, I just updated the default homedir for production env

arthosofteq: It is not enough to change default value for production only. We need to remove circleci  APP_FOLDER_NAME env then

arthosofteq: Never mind. I'll remove it myself

arthosofteq (approved):

PR #67, opened 2021-12-01
Create codeql-analysis.yml

gkorland:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main, latest, release/* ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main ]
+  schedule:
+    - cron: '37 11 * * 3'
```
arthosofteq, line 21: @gkorland Do we need to run jobs by cron and why 11:37 each Wednesday then?

arthosofteq (approved):

PR #71, opened 2021-12-01
#RI-2100 - fix links for commands

romansergeenkosofteq:


redisinsight/ui/src/constants/commands.ts
```diff
@@ -62,6 +62,14 @@ export enum CommandGroup {
   AI = 'ai'
 }
 
+export enum CommandGroupsStartsWith {
```
zalenskiSofteq, line 65: May be better name for this enum is `CommandPrefix`?

romansergeenkosofteq: yes, renamed

romansergeenkosofteq: yes, renamed

zalenskiSofteq (approved):

PR #85, opened 2021-12-07
There are 25 dependabot security alerts 

mmorel-35:
Hi! I love redisinsight! Thanks you very much for such a great tool! 
Once I forked the repository, I saw that there are 25 dependabot security alerts.
Here you'll find a configuration to set up Dependabot for npm and docker dependencies.
This is checking the available updates every week and allows up to 100 PR for npm.
This will be a lot at first but once treated there will be just the delta between two weeks. From what I have seen you'll receive 96 npm and one docker PR.
You can find more documention about dependabot [here](https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates)


mmorel-35:
Friendly ping @arthosofteq, what do you think ?

mmorel-35:
I have updated every dependency I could. I released the constraint on node version being strictly under 16.
I updated node version to 16.13.1 as 15.x is not lts and was causing trouble to update several dependencies.

PR #93, opened 2021-12-08
#RI-1841 - add support for auto updates to stage builds

arthosofteq:
Also possibility for updates for any build environment if "manual" environment variable defined

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -432,7 +432,7 @@ jobs:
               exit 0;
             fi
 
-            UPGRADES_LINK=UPGRADES_LINK_STAGE SEGMENT_WRITE_KEY=$SEGMENT_WRITE_KEY_STAGE yarn package:<< parameters.env >>
+            UPGRADES_LINK=$UPGRADES_LINK_STAGE SEGMENT_WRITE_KEY=$SEGMENT_WRITE_KEY_STAGE yarn package:<< parameters.env >>
       - persist_to_workspace:
           root: .
           paths:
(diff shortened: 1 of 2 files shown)


redisinsight/main.dev.ts
```diff
@@ -119,8 +119,20 @@ const bootstrap = async () => {
     trayInstance = tray.buildTray();
   }
 
-  if (process.env.NODE_ENV === 'production') {
-    new AppUpdater();
+  let upgradeUrl;
+  switch (process.env.NODE_ENV) {
```
zalenskiSofteq, line 123: I think better to replace this `switch` to rewrite env variable in the config for circle.ci

.circleci/config.yml
```diff
@@ -432,7 +432,7 @@ jobs:
               exit 0;
             fi
 
-            SEGMENT_WRITE_KEY=$SEGMENT_WRITE_KEY_STAGE yarn package:<< parameters.env >>
+            UPGRADES_LINK=UPGRADES_LINK_STAGE SEGMENT_WRITE_KEY=$SEGMENT_WRITE_KEY_STAGE yarn package:<< parameters.env >>
```
zalenskiSofteq, line 435: I think circle ci env variable should start with symbol `$`, isn't it?
`UPGRADES_LINK=$UPGRADES_LINK_STAGE`

arthosofteq: Good catch! Thanks!

zalenskiSofteq (approved):

PR #114, opened 2021-12-14
#RI-2169, #RI-2135

zalenskiSofteq:
* #RI-2135 - Request POST /api/instance/{id}/cli//send-command is sent for command execution in Workbench
* #RI-2169 - Area's layout has shifted

### DIFF
--- a/redisinsight/menu.ts
+++ b/redisinsight/menu.ts
@@ -120,7 +120,7 @@ export default class MenuBuilder {
           label: 'New Window',
           accelerator: 'Command+N',
           click: () => {
-            createWindow(null);
+            createWindow();
           },
         },
         {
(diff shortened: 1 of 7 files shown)


redisinsight/menu.ts
```diff
@@ -120,7 +120,7 @@ export default class MenuBuilder {
           label: 'New Window',
           accelerator: 'Command+N',
           click: () => {
-            createWindow();
+            createWindow(null);
```
romansergeenkosofteq, line 123: Maybe better to set default value for splash?
`splash: BrowserWindow | null = null`

zalenskiSofteq: yes thanks
changed

romansergeenkosofteq (approved):

PR #153, opened 2021-12-24
[E2E] Workbench tests are added

elenaNaboko:
The following tests are added:
 - Verify that user can find (using right click) "Run Commands" custom shortcut option in monaco menu and run a command;
 - Verify that user can see saved scroll position in Enablement area when he leaves Workbench page and goes back again.


tests/e2e/tests/regression/workbench/default-scripts-area.e2e.ts
```diff
@@ -72,3 +72,26 @@ test('Verify that user can see saved article in Enablement area when he leaves W
...
+    //Scroll to the very bottom of the page
+    await t.scrollIntoView(lastButton);
+    //Check the scroll position
+    const scrollPosition = await workbenchPage.scrolledEnablementArea.scrollTop;
+    //Go to Browser page
+    await t.click(myRedisDatabasePage.browserButton);
+    //Go back to Workbench page
+    await t.click(myRedisDatabasePage.workbenchButton);
+    //Check that scroll position is saved
+    await t.expect(workbenchPage.scrolledEnablementArea.scrollTop).eql(scrollPosition, 'Scroll position is correct');
```
ofersteinberg, line 90: Regarding the error message, the more correct is `The scroll position status`

tests/e2e/tests/regression/workbench/scripting-area.e2e.ts
```diff
@@ -80,3 +80,19 @@ test('Verify that user can see an indication (green triangle) of commands from t
...
+test('Verify that user can find (using right click) "Run Commands" custom shortcut option in monaco menu and run a command', async t => {
+    const command = 'HSET key field value';
+    //Put a command in Editing Area
+    await t.typeText(workbenchPage.queryInput, command);
+    //Right click to get context menu
+    await t.rightClick(workbenchPage.queryInput);
+    //Select Command Palette option
+    await t.click(workbenchPage.monacoContextMenu.find(workbenchPage.cssMonacoCommandPaletteLine));
+    //Print "Run Commands" shortcut
+    await t.typeText(workbenchPage.monacoShortcutInput, 'Run Commands');
```
ofersteinberg, line 92: I recommend using with `{ replace: true, paste: true }` on `typeText`

tests/e2e/tests/regression/workbench/scripting-area.e2e.ts
```diff
@@ -80,3 +80,19 @@ test('Verify that user can see an indication (green triangle) of commands from t
...
+    //Select "Run Commands" from menu
+    await t.click(workbenchPage.monacoSuggestionOption);
+    //Check the result with sent command
+    const commandTextInResult = await workbenchPage.queryCardCommand.withExactText(command);
+    await t.expect(commandTextInResult.exists).ok('Result of sent command exists');
```
ofersteinberg, line 97: More accurate error message: `The result of sent command`

ofersteinberg (changes requested):

tests/e2e/tests/regression/workbench/default-scripts-area.e2e.ts
elenaNaboko: Fixed

tests/e2e/tests/regression/workbench/scripting-area.e2e.ts
elenaNaboko: Fixed

tests/e2e/tests/regression/workbench/scripting-area.e2e.ts
elenaNaboko: Did not add replace and paste options because monaco context menu doesn't work with them and test fails. Need "typeahead" behavior in this test

ofersteinberg (approved):

PR #181, opened 2022-01-04
#RI-2146 add possibility to save plugin state

arthosofteq:


redisinsight/api/src/modules/workbench/providers/command-execution.provider.ts
```diff
@@ -28,16 +32,37 @@ export class CommandExecutionProvider {
   async create(commandExecution: Partial<CommandExecution>): Promise<CommandExecution> {
     const entity = plainToClass(CommandExecutionEntity, commandExecution);
 
-    return classToClass(
+    // Do not store command execution result that exceeded limitation
+    if (JSON.stringify(entity.result).length > WORKBENCH_CONFIG.maxResultSize) {
+      entity.result = JSON.stringify([
+        {
+          status: CommandExecutionStatus.Success,
```
ArtyomPodymov, line 39: Are you sure the status should be Success?

arthosofteq: yes. it should be. here is command execution status that is mean that redis command was successfully execute. It is not about BE limitation on storing results

ArtyomPodymov (approved):

PR #188, opened 2022-01-10
[E2E] Test with TTL value is added

elenaNaboko:


tests/e2e/helpers/keys.ts
```diff
@@ -0,0 +1,23 @@
+import { KeyTypesTexts } from './constants';
+
+export function getRandomKeyName(keyNameLength: number): string {
+    let result = '';
+    const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+    const charactersLength = characters.length;
+    for (let i = 0; i < keyNameLength; i++) {
+        result += characters.charAt(Math.floor(Math.random() * charactersLength));
+    }
```
ofersteinberg, line 9: IMO You can use with [chance](https://www.npmjs.com/package/chance) instead of this

elenaNaboko: Great library, thanks! Fix test with 'chance'

tests/e2e/package.json
```diff
@@ -23,8 +23,10 @@
     "cli-argument-parser": "0.4.5"
   },
   "devDependencies": {
+    "@types/chance": "^1.1.3",
```
ofersteinberg, line 26: Please pin the versions

tests/e2e/helpers/keys.ts
```diff
@@ -1,11 +1,11 @@
...
 import { KeyTypesTexts } from './constants';
 
 export function getRandomKeyName(keyNameLength: number): string {
+    const chance = new Chance();
     let result = '';
-    const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
-    const charactersLength = characters.length;
     for (let i = 0; i < keyNameLength; i++) {
-        result += characters.charAt(Math.floor(Math.random() * charactersLength));
+        result += chance.character();
```
ofersteinberg, line 8: It means you use the default, are `!@#$%^&*()` valid characters for your use cases?

ofersteinberg (changes requested):
I see there is some error on `redisinsight/api/test/helpers/local-db.ts` can you fix it in addition?

tests/e2e/package.json
elenaNaboko: Fixed. Why we should use exact version and “Compatible with version” is not valid?

tests/e2e/helpers/keys.ts
elenaNaboko: Yes, it will be better to use whole symbol range. This function is used for key name generation which can contain any symbols

elenaNaboko:
> I see there is some error on `redisinsight/api/test/helpers/local-db.ts` can you fix it in addition?

This is the integration tests issue, connected with weak password encryption. This is can not be fixed, discussed with @arthosofteq

arthosofteq:
> I see there is some error on `redisinsight/api/test/helpers/local-db.ts` can you fix it in addition?

Yeah, an error says: `Storing cryptographic hashes of passwords is standard security practice, but it is equally important to select the right hashing scheme. If an attacker obtains the hashed passwords of an application, the password hashing scheme should still prevent the attacker from easily obtaining the original cleartext passwords.`
We are not storing this hashes anywhere. We used it just to create cipher key with fixed length from the password (that stored securely). So we are safe at the moment. However even if we decide to change hash schema it might bring us problems since users already have encrypted data with such cipher key (created using current hash schema)

tests/e2e/package.json
arthosofteq: @ofersteinberg Why do we need to pin the version?

ofersteinberg: We want to control it. It avoids failures in the future if they will have breaking changes.

ofersteinberg (approved):

PR #211, opened 2022-01-19
mechanism for run rte with predefined data

arthosofteq:


tests/e2e/rte/oss-standalone-big/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+echo 'Downloading .rdb file...'
+curl --request GET -sL \
+     --url 'https://s3.amazonaws.com/redisinsight.test/public/rte/big_db.tar.gz'\
```
zalenskiSofteq, line 5: 1. May be better use Cloud front for fast download? link for tests s3 is `https://download-test.redisinsight.redis.com/` tests will pass faster)
2. No reason for use env variable for path?

arthosofteq: @zalenskiSofteq ok. probably you are right. 
Moved dump url to docker file arg and made it optional. So if you build image with TEST_BIG_DB_DUMP arg we will try to extract dump from archive and run redis with it. If it is not defined we will run redis with empty db. As we moved optinal dump option to dockerfile we will download dump on image build only which will save time to run container.

zalenskiSofteq (approved):

PR #241, opened 2022-01-28
Fix/xhr error message

ArtyomPodymov:
#RI-2356 - specify a human-readable error message

### DIFF
--- a/redisinsight/ui/src/components/monitor-config/MonitorConfig.tsx
+++ b/redisinsight/ui/src/components/monitor-config/MonitorConfig.tsx
@@ -11,7 +11,7 @@ import {
   stopMonitor,
 } from 'uiSrc/slices/cli/monitor'
 import { getBaseApiUrl } from 'uiSrc/utils'
-import { MonitorEvent, SocketEvent } from 'uiSrc/constants'
+import { MonitorEvent, SocketErrors, SocketEvent } from 'uiSrc/constants'
 import { IMonitorDataPayload } from 'uiSrc/slices/interfaces'
 import { connectedInstanceSelector } from 'uiSrc/slices/instances'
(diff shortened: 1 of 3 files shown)


redisinsight/ui/src/components/monitor-config/MonitorConfig.tsx
```diff
@@ -87,8 +87,12 @@ const MonitorConfig = ({ retryDelay = 10000 } : IProps) => {
     })
 
     // Catch connect error
-    newSocket.on(SocketEvent.ConnectionError, (error: Error) => {
-      payloads.push({ isError: true, time: `${Date.now()}`, message: `${error?.name}: ${error?.message}` })
+    newSocket.on(SocketEvent.ConnectionError, (error: { type: string; name: any; message: any }) => {
+      if (error?.type === 'TransportError') {
```
zalenskiSofteq, line 91: Could you create a enum for constant string? May be in the future any type will added

romansergeenkosofteq: and better to use switch case insted of `if else` in the future

ArtyomPodymov: > Could you create a enum for constant string? May be in the future any type will added

done

zalenskiSofteq (approved):

redisinsight/ui/src/components/monitor-config/MonitorConfig.tsx
ArtyomPodymov: > and better to use switch case instead of `else if` in the future

according to the eslint rules, we must have at least one `else if`, otherwise the `swith` construct is useless (as in current case)

romansergeenkosofteq (approved):

PR #273, opened 2022-02-07
e2e - cli logical db tests

tanyatrayanava:


tests/e2e/tests/regression/cli/cli.e2e.ts
```diff
@@ -91,3 +101,81 @@ test
         //Verify result
         await t.expect(cliPage.cliOutputResponseSuccess.count).eql(repeats, `CLI contains ${repeats} results`);
 });
+test
+    .meta({ rte: rte.standalone })
+    .before(async () => {
+        await acceptLicenseTerms();
+    })
```
ofersteinberg, line 108: FYI This overrides the beforeEach fixture hook

tests/e2e/tests/regression/cli/cli.e2e.ts
```diff
@@ -91,3 +101,81 @@ test
...
+        await t.expect(cliPage.cliEndpoint.textContent).eql(databaseEndpoint, 'Database index 0 is not displayed in the CLI endpoint');
+});
+test
+    .meta({ rte: rte.standalone })
+    .before(async () => {
+        await acceptLicenseTerms();
+    })
+    .after(async () => {
+        //Delete database
+        await deleteDatabase(ossStandaloneConfig.databaseName + ` [${index}]`);
```
ofersteinberg, line 133: Could you adapt the beforeEach work with compatible all cases?

tests/e2e/tests/regression/cli/cli.e2e.ts
tanyatrayanava: in this hook the db is not added, because it should be added inside test 
and in the beforeEach hook the standalone db without any index is added

tests/e2e/tests/regression/cli/cli.e2e.ts
tanyatrayanava: removed tests with logical db in the different file

ofersteinberg (approved):

PR #277, opened 2022-02-08
#RI-2386 - Cover cases when user does not have permissions for Monitor command

romansergeenkosofteq:


redisinsight/ui/src/components/monitor/Monitor/Monitor.tsx
```diff
@@ -10,12 +11,15 @@ import {
 import { AutoSizer } from 'react-virtualized'
 
 import { IMonitorDataPayload } from 'uiSrc/slices/interfaces'
+import { ReactComponent as NoPermissionsIcon } from 'uiSrc/assets/img/monitor/no_permissions.svg'
```
ArtyomPodymov, line 14: It might be better to name this icon `ban,svg`. This is a very common name https://fontawesome.com/icons/ban?s=regular

redisinsight/ui/src/constants/apiStatusCode.ts
```diff
@@ -1,6 +1,7 @@
 enum ApiStatusCode {
   Unauthorized = 401,
   BadRequest = 400,
+  NoPermission = 403
```
ArtyomPodymov, line 4: Better to use `Forbidden = 403`

romansergeenkosofteq: done

ArtyomPodymov (approved):

ArtyomPodymov (approved):

PR #374, opened 2022-03-14
disable autoupdate for Redis Stack

zalenskiSofteq:


redisinsight/main.dev.ts
```diff
@@ -121,7 +121,7 @@ const bootstrap = async () => {
 
   const upgradeUrl = process.env.MANUAL_UPGRADES_LINK || process.env.UPGRADES_LINK;
 
-  if (upgradeUrl) {
+  if (upgradeUrl && process.env.BUILD_TYPE !== 'REDIS_STACK') {
```
arthosofteq, line 124: It will not work this way
We are creating regular ELECTRON build for stack app
We need to handle it on build in the CI

arthosofteq (changes requested):

PR #393, opened 2022-03-17
#RI-2494-add script to download default content

AmirAllayarovSofteq:


redisinsight/api/scripts/default-content.ts
```diff
@@ -0,0 +1,54 @@
...
+      buildInfo,
+    );
+
+    process.exit(0);
+  } catch (e) {
+    console.error('Something went wrong trying to get default commands jsons', e);
+    process.exit(1);
+  }
+}
+
```
arthosofteq, line 53: Do not forget to rewrite this script using new methods that you've created for guides and tutorials

arthosofteq:
Did you test it on your local machine because I don't see any changes in the provider that should copy defaults to home directory?

arthosofteq (approved):

redisinsight/api/scripts/default-content.ts
AmirAllayarovSofteq: Yep. Remember it

PR #455, opened 2022-03-25
#RI-2568 - improve dedicated editor positioning

romansergeenkosofteq:


redisinsight/ui/src/components/query/Query/Query.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { useContext, useEffect, useRef, useState } from 'react'
+import { AutoSizer } from 'react-virtualized'
```
zalenskiSofteq, line 2: better use separated AutoSizer from 'react-virtualized-auto-sizer'
`import AutoSizer from 'react-virtualized-auto-sizer'`

romansergeenkosofteq: fixed

zalenskiSofteq (approved):

PR #477, opened 2022-03-30
e2e - automatically update

tanyatrayanava:


tests/e2e/tests/critical-path/database/autoupdate-of-json.e2e.ts
```diff
@@ -0,0 +1,49 @@
...
+        const createRedisFilePathNew = editJsonFile(createRedisPath);
+        //Check the promo button
+        await t.expect(myRedisDatabasePage.promoButton.textContent).notContains(cloudValueForEdit, 'Promo button text is updated');
+        //Check the json files are automatically updated
+        const timestampAfterUpdate = await buildFilePathNew.get('timestamp');
+        const cloudTitle = await createRedisFilePathNew.get('cloud.title');
+        const cloudDescription = await createRedisFilePathNew.get('cloud.description');
+        await t.expect(timestampAfterUpdate).notEql(timestampForEdit);
+        await t.expect(cloudTitle).notEql(cloudValueForEdit);
+        await t.expect(cloudDescription).notEql(cloudValueForEdit);
```
ofersteinberg, line 48: Please add error message

ofersteinberg, line 48: Something here isn't clear enough.
Do the first two lines are apart of the test or pre requisite? 
It's hard to understand from reading the code the test case, and I feel that in case the test will be fail, shouldn't be easy understand the bug (and if the bug is on the product or test).

tests/e2e/package.json
```diff
@@ -26,11 +26,13 @@
   },
   "devDependencies": {
     "@types/chance": "1.1.3",
+    "@types/edit-json-file": "^1.7.0",
```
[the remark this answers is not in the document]
tanyatrayanava: done

tests/e2e/tests/critical-path/database/autoupdate-of-json.e2e.ts
tanyatrayanava: added

tanyatrayanava: these first two lines are necessary because the data that is in these files is cached and needs to be overwritten to get the updated information
I also added more comments to this test

ofersteinberg (dismissed):

ofersteinberg (approved):

PR #482, opened 2022-04-01
[E2E] Sorting modules test added

elenaNaboko:


tests/e2e/tests/critical-path/database/modules.e2e.ts
```diff
@@ -29,8 +29,14 @@ test
     ('Verify that user can see DB modules on DB list page for Standalone DB', async t => {
         //Check module column on DB list page
         await t.expect(myRedisDatabasePage.moduleColumn.exists).ok('Module column');
-        //Check that module icons are displayed
-        await myRedisDatabasePage.checkModulesOnPage(moduleList);
+        //Verify that user can see the following sorting order: Search, JSON, Graph, TimeSeries, Bloom, Gears, AI for modules
+        const databaseLine = await myRedisDatabasePage.dbNameList.withExactText(ossStandaloneRedisearch.databaseName).parent('tr');
```
ofersteinberg, line 33: So if the order will be changed, the test will fail?
Can you show maybe a screenshot of what you test here?

elenaNaboko: Yes, with different order the test will failed. We had the story about sorting modules order (RI-2491) so I have added one check inside of existed test. 
image

ofersteinberg: But why check the order? or not checking only the existence?

elenaNaboko: Because we have the story according to which we need to display all these modules in specific order. RediSearch firstly, then RedisJSON, then Graph, TimeSeries, Bloom, Gears and AI. I added this validating in already existing test just to reduce tests time executions. If order change in the application, test will fail (what is expected).

ofersteinberg (approved):
Approving although I less agree with the test approach.

PR #484, opened 2022-04-02
e2e - RedisGraph and TimeSeries tests

tanyatrayanava:


tests/e2e/pageObjects/workbench-page.ts
```diff
@@ -199,13 +219,19 @@ export class WorkbenchPage {
...
+    }
+
     /**
      * Send a command in Workbench
      * @param command The command
      * @param speed The speed in seconds. Default is 1
      * @param paste
      */
-    async sendCommandInWorkbench(command: string, speed = 1, paste = false): Promise<void> {
+    async sendCommandInWorkbench(command: string, speed = 1, paste = true): Promise<void> {
```
ofersteinberg, line 234: I don't think this should be a parameter (also the speed). IMO you can do it built the function.
Do you have tests that use different speed and paste configurations?

tests/e2e/tests/regression/workbench/redis-stack-commands.e2e.ts
```diff
@@ -0,0 +1,74 @@
...
+        await t.expect(await workbenchPage.queryCardContainer.nth(0).find(workbenchPage.queryGraphContainer).exists).ok('The Graph view is switched for GRAPH command');
+    });
+test
+    .meta({ rte: rte.standalone })
+    ('Verify that user can see "No data to visualize" message for Graph command', async t => {
+        //Send Graph command
+        await t.click(workbenchPage.redisStackTutorialsButton);
+        await t.click(workbenchPage.workingWithGraphLink);
+        await t.click(workbenchPage.preselectModelBikeSalesButton);
+        await t.click(workbenchPage.submitCommandButton);
```
ofersteinberg, line 52: Make a function within the file if this action is repeating.

ofersteinberg (changes requested):

tests/e2e/tests/regression/workbench/redis-stack-commands.e2e.ts
tanyatrayanava: this steps are bot repeating

tests/e2e/pageObjects/workbench-page.ts
tanyatrayanava: yes, in some tests are set different values for this parameters

ofersteinberg: Ok, at least add the description above. Also, since you changed it from false to true, don't you update the other tests?

tanyatrayanava: yes, thanks, removed the parameters, for the rest of the tests, I did not change the paste parameter to false, because this will slow down the test

ofersteinberg (approved):

PR #552, opened 2022-04-20
remove files after shutting down

arthosofteq:


redisinsight/main.dev.ts
```diff
@@ -387,6 +389,15 @@ app.on('certificate-error', (event, _webContents, _url, _error, _certificate, ca
   callback(true);
 });
 
+app.on('quit', () => {
+  try {
+    if (backendGracefulShutdown) {
```
zalenskiSofteq, line 394: Could you rewrite with optional chain? instead of `if` conditional just:
`backendGracefulShutdown?.();`

zalenskiSofteq (approved):

PR #620, opened 2022-05-11
e2e - Show Stream in a table

tanyatrayanava:


tests/e2e/tests/regression/browser/stream-key.e2e.ts
```diff
@@ -0,0 +1,110 @@
...
+import { BrowserPage, CliPage } from '../../../pageObjects';
+import { commonUrl, ossStandaloneConfig } from '../../../helpers/conf';
+
+const browserPage = new BrowserPage();
+const cliPage = new CliPage();
+const chance = new Chance();
+
+const field = chance.word({length: 5});
+const value = chance.word({length: 5});
+let keyName = chance.word({length: 20});
```
ofersteinberg, line 13: I see you like to use it, you can create a function in helped which generates it and send only the length. Would be essentials for you

tests/e2e/tests/regression/browser/stream-key.e2e.ts
```diff
@@ -0,0 +1,110 @@
...
+    await t.expect(browserPage.virtualTableContainer.visible).ok('The Stream is displayed in a table format');
+    for(const field of streamFields){
+        await t.expect(browserPage.streamEntriesContainer.textContent).contains(field, 'The Stream fields are in the table');
+    }
+    await t.expect(browserPage.streamFieldsValues.textContent).contains(value, 'The Stream field value is in the table');
+});
+test('Verify that user can sort ASC/DESC by Entry ID', async t => {
+    keyName = chance.word({length: 20});
+    const command = `XADD ${keyName} * '${field}' '${value}'`;
+    //Add new Stream key
```
ofersteinberg, line 50: Mention you do that 5 times. Also, (just suggestion) can be nice to create a function of sending command that cand do that multiple time instead of use this loop each time

tanyatrayanava: done

PR #646, opened 2022-05-17
#RI-2855-update stream get entri controller

AmirAllayarovSofteq:


arthosofteq (commented):
WHy we are using `""` empty string and not `null`?

AmirAllayarovSofteq:
Made it same as initial value in FE

arthosofteq (approved):

PR #688, opened 2022-05-23
#RI-2931 BE implementation for create new consumer group(s)

arthosofteq:
#RI-2933 BE implementation for consumer groups list


AmirAllayarovSofteq (approved):
LGTM

PR #707, opened 2022-05-25
#RI-2987 - Slow Log page doesn't open, no requests are sent for Redis…

zalenskiSofteq:
#RI-2987 - Slow Log page doesn't open, no requests are sent for RedisStack DB


AmirAllayarovSofteq (approved):
LGTM

PR #761, opened 2022-06-09
#RI-3038 BE return total keys per database if dbN > 1

arthosofteq:


redisinsight/api/src/modules/shared/services/instances-business/overview.service.ts
```diff
@@ -192,27 +194,44 @@ export class OverviewService {
...
+
+      // filter 0 values
+      Object.keys(totalKeysPerDb).forEach((key) => {
+        if (!totalKeysPerDb[key]) {
+          delete totalKeysPerDb[key];
+        }
+      });
+
+      const totalKeys = totalKeysPerDb ? sum(Object.values(totalKeysPerDb)) : undefined;
+      return [totalKeys, Object.keys(totalKeysPerDb).length > 1 ? totalKeysPerDb : undefined];
```
AmirAllayarovSofteq, line 232: May be will be clearer if move Object.keys(totalKeysPerDb).length > 1 ? totalKeysPerDb : undefined] to variable ?

AmirAllayarovSofteq (approved):
LGTM!

PR #781, opened 2022-06-15
#RI-3037-add spublish check

AmirAllayarovSofteq:


redisinsight/ui/src/constants/redisVersions.ts
```diff
@@ -0,0 +1,5 @@
+export const RedisVersions = {
```
zalenskiSofteq, line 1: We have a file with constants for different version. Please move there
`commandsVersions.ts`

zalenskiSofteq (approved):

PR #853, opened 2022-07-05
e2e - Search the list of db tests

vlad-dargel:


tests/e2e/helpers/api/api-database.ts
```diff
@@ -0,0 +1,73 @@
...
+    return await response.body;
+}
+
+/**
+ * Get database through api using database name
+ * @param databaseName The database name
+ */
+export async function getDatabaseByName(databaseName: string): Promise<void> {
+    const results = await getAllDatabases();
+    let res: object = {};
```
ofersteinberg, line 54: Call it `response` or use other better name

tests/e2e/helpers/api/api-database.ts
```diff
@@ -0,0 +1,73 @@
...
+ * Delete database through api using database name
+ * @param databaseName The database name
+ */
+export async function deleteDatabaseByName(databaseName: string): Promise<void> {
+    const databaseId = await getDatabaseByName(databaseName);
+    const response = await request(endpoint).delete(`/instance`)
+        .send({ "ids": [`${databaseId}`] }).set('Accept', 'application/json');
+
+    await t.expect(await response.status).eql(200, 'deleteDatabaseByName request failed');
+}
```
ofersteinberg, line 73: General comment - I think better to use HTTP client for API requests

tests/e2e/helpers/async-helper.ts
```diff
@@ -0,0 +1,15 @@
...
+    const fail = Symbol();
+    return (await Promise.all(arr.map(async item => (await callback(item)) ? item : fail))).filter(i => i !== fail);
+};
+
+async function asyncFind(arr: any, asyncCallback: any) {
+    const index = (await Promise.all(arr.map(asyncCallback))).findIndex(result => result);
+    return arr[index];
+}
+
+function doAsyncStuff() {
```
ofersteinberg, line 11: Descriptions etc are missed

tests/e2e/tests/regression/browser/database-list-search.e2e.ts
```diff
@@ -0,0 +1,104 @@
...
+fixture`Database list search`
+    .meta({ type: 'regression' })
+    .page(commonUrl)
+    .beforeEach(async () => {
+        // Add new databases using API
+        await acceptLicenseTerms();
+        await addNewStandaloneDatabase(ossStandaloneConfig);
+        await addNewStandaloneDatabase(databasesForSearch[0]);
+        await addNewStandaloneDatabase(databasesForSearch[1]);
+        await addNewStandaloneDatabase(databasesForSearch[2]);
```
ofersteinberg, line 25: You could use a loop, or, consider creating a function of `addNewStandaloneDatabases`

tests/e2e/tests/regression/browser/database-list-search.e2e.ts
```diff
@@ -0,0 +1,104 @@
...
+        // Reload Page
+        await t.eval(() => location.reload());
+    })
+    .afterEach(async () => {
+        //Clear and delete databases
+        await deleteDatabaseByName(ossStandaloneConfig.databaseName);
+        await deleteDatabaseByName(databasesForSearch[0].databaseName);
+        await deleteDatabaseByName(databasesForSearch[1].databaseName);
+        await deleteDatabaseByName(databasesForSearch[2].databaseName);
+        await deleteDatabaseByName(ossClusterConfig.ossClusterDatabaseName);
```
ofersteinberg, line 36: Same comment

ofersteinberg (changes requested):

tests/e2e/helpers/api/api-database.ts
```diff
@@ -0,0 +1,73 @@
+
+import { t } from 'testcafe';
+import * as request from 'supertest';
+import { asyncFilter, doAsyncStuff } from '../async-helper'
+import { AddNewDatabaseParameters, OSSClusterParameters } from '../../pageObjects/add-redis-database-page';
+import { apiUrl } from '../../helpers/conf';
+
+const endpoint = apiUrl;
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'; // lgtm[js/disabling-certificate-validation]
```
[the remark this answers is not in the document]
vlad-dargel: it is an environment variable for avoiding an error  'unable to verify the first certificate'. This works to ignore the certificate verification when testing HTTPS against localhost. 
I've took this way from integration tests
RedisInsight\redisinsight\api\test\helpers\server.ts

tests/e2e/tests/regression/browser/database-list-search.e2e.ts
```diff
@@ -0,0 +1,104 @@
...
+        const searchedDBHost = 'Standalone'
+        await t.typeText(myRedisDatabasePage.searchInput, searchedDBHost, { replace: true });
+        //Verify that database found on the list search by connection type
+        await t.expect(myRedisDatabasePage.dbNameList.withExactText(databasesForSearch[0].databaseName).exists).ok('The database with connection type is found', { timeout: 10000 });
+        await t.expect(myRedisDatabasePage.dbNameList.withExactText(databasesForSearch[1].databaseName).exists).ok('The database with connection type is found', { timeout: 10000 });
+        await t.expect(myRedisDatabasePage.dbNameList.withExactText(ossClusterConfig.ossClusterDatabaseName).exists).notOk('The database with other connection type is not found', { timeout: 10000 });
+    });
+test
+    .before(async () => {
+        await t.wait(60000);
```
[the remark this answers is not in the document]
vlad-dargel: This is needed here to wait 1 min until "Last connection" value will be changed from "less than a minute ago" to "1 minute ago". There is no other way to change this falue, it is calculated on FE.

tests/e2e/helpers/async-helper.ts
vlad-dargel: fixed, thanks

tests/e2e/helpers/api/api-database.ts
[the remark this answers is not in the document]
ofersteinberg: I see it throws an alert. Could you think of an alternative solution?

tests/e2e/tests/regression/browser/database-list-search.e2e.ts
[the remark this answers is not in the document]
ofersteinberg: Can't you implement a while loop that waits for this status?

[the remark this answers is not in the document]
vlad-dargel: looks good, will try this, thanks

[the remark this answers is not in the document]
vlad-dargel: fixed

tests/e2e/tests/regression/browser/database-list-search.e2e.ts
vlad-dargel: fixed, function addNewStandaloneDatabases added

tests/e2e/tests/regression/browser/database-list-search.e2e.ts
vlad-dargel: fixed, function deleteStandaloneDatabases added

tests/e2e/helpers/api/api-database.ts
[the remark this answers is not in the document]
ofersteinberg: Can't you call it via the CI system instead of within the test?

tests/e2e/helpers/api/api-database.ts
```diff
@@ -0,0 +1,107 @@
+
```
ofersteinberg, line 1: Redundant line

tests/e2e/helpers/api/api-database.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Add a new Standalone database through api using host and port
+ * @param databaseParameters The database parameters
+ */
+export async function addNewStandaloneDatabase(databaseParameters: AddNewDatabaseParameters): Promise<void> {
+    const response = await request(endpoint).post(`/instance`)
+        .send({ "name": databaseParameters.databaseName, "host": databaseParameters.host, "port": databaseParameters.port })
+        .set('Accept', 'application/json');
+
+    await t.expect(await response.status).eql(201);
+    await t.expect(await response.body.name).eql(databaseParameters.databaseName);
```
ofersteinberg, line 21: Please add an error message (also bellow)

tests/e2e/helpers/api/api-database.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/**
+ * Add a new Standalone databases through api using host and port
+ * @param databasesParameters The databases parameters array
+ */
+export async function addNewStandaloneDatabases(databasesParameters: any): Promise<void> {
```
ofersteinberg, line 28: Why type `any`? You use with typescript, you can create your own type

tests/e2e/helpers/async-helper.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Helper function to work with arr.filter() method with async functions
+ * @param array The array
+ * @param callback The callback function need to be processed
+ */
+async function asyncFilter(array: any, callback: any) {
```
ofersteinberg, line 6: Couldn't you declare a dedicated type these both parameters?

tests/e2e/tests/regression/browser/database-list-search.e2e.ts
```diff
@@ -0,0 +1,110 @@
...
+    })
+    .afterEach(async () => {
+        //Clear and delete databases
+        await deleteStandaloneDatabases(databasesForAdding);
+        await deleteOSSClusterDatabase(ossClusterConfig);
+    });
+test
+    .meta({ rte: rte.standalone })('Verify that user can search DB by database name on the List of databases', async t => {
+        //Search for DB by name
+        const searchedDBName = 'Search'
```
ofersteinberg, line 46: Please add `;` (don't you have eslint here? @elenaNaboko )

ofersteinberg (changes requested):

tests/e2e/helpers/api/api-database.ts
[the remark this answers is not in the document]
arthosofteq: If you mean to use CircleCI env variables for it - we will need to do the same on local machines for all devs then which is incorrect. Also, it is incorrect to set it on CircleCI env variables because this rule will be applied to all other parts of application
If you suggest to invoke tests with inline env variable like `ENV=val ./something` in circleci.yml or even using .env file we have to store it in the source code anyway and code analysis tool will still find this.

tests/e2e/tests/regression/browser/database-list-search.e2e.ts
AlenaNaboka: We have, but for some reason VS Code doesn't highlight some issue. WebStorm does 
image

tests/e2e/helpers/async-helper.ts
vlad-dargel: fixed, thanks

tests/e2e/helpers/api/api-database.ts
vlad-dargel: fixed, thanks

tests/e2e/helpers/api/api-database.ts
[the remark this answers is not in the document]
arthosofteq: @ofersteinberg It has. This is what I was talking about. 
In this case we will remove this row in the code and move it to the vault. As a result in my local PC tests will fail. I have to add such env to my local machine too then (each dev and qa have to do it). But I don't want to care about it. I want all tests are running out-of-the box without additional steps to do.

tests/e2e/helpers/api/api-database.ts
```diff
@@ -17,8 +16,20 @@ export async function addNewStandaloneDatabase(databaseParameters: AddNewDatabas
         .send({ "name": databaseParameters.databaseName, "host": databaseParameters.host, "port": databaseParameters.port })
         .set('Accept', 'application/json');
 
-    await t.expect(await response.status).eql(201);
-    await t.expect(await response.body.name).eql(databaseParameters.databaseName);
+    await t.expect(await response.status).eql(201, 'addNewStandaloneDatabase request failed');
```
ofersteinberg, line 19: Instead of `addNewStandaloneDatabase request failed` please write something such as `The creation of standalone database`
Same bellow. Do not use synonymous such `addNewStandaloneDatabase` in the error message. It should be readable for anyone.

vlad-dargel: Thanks, fixed

tests/e2e/tests/regression/database/database-list-search.e2e.ts
```diff
@@ -0,0 +1,110 @@
...
+    });
+test
+    .meta({ rte: rte.standalone })('Verify that user sees "No results found" message when pattern doesn`t match any database', async t => {
+        //Search for DB by Invalid search
+        const searchedDBHost = 'invalid';
+        await t.typeText(myRedisDatabasePage.searchInput, searchedDBHost, { replace: true });
+        //Verify that "No results found" message is displayed in case of invalid search
+        await t.expect(myRedisDatabasePage.noResultsFoundMessage.exists).ok('"No results found message" not displayed');
+        await t.expect(myRedisDatabasePage.noResultsFoundText.exists).ok('"No databases matched your search" message not displayed');
+    });
```
ofersteinberg, line 110: @ViktarStarastsenka - here is an example that most tests could cover by JEST.

ofersteinberg (approved):

ofersteinberg (approved):

PR #893, opened 2022-07-14
#RI-3246, #RI-3247

zalenskiSofteq:
#RI-3246 - Value is cut when user edits key value
#RI-3247 - Next value has the same cell size when user deletes previous hash value


AmirAllayarovSofteq (approved):
LGTM

PR #911, opened 2022-07-15
#RI-3258 - After key value editing and saving, user should see previous state of the value

zalenskiSofteq:


AmirAllayarovSofteq (approved):
LGTM

PR #919, opened 2022-07-18
#RI-3292, #RI-3287, #RI-3288

zalenskiSofteq:
* #RI-3292 - Error messages overlappses if key name is too long
* #RI-3287 - Switching between tree view and browser, doesn't apply filters
* #RI-3288 - New part of keys are not displayed in Tree View when user scans more


AmirAllayarovSofteq (approved):
LGTM

PR #932, opened 2022-07-20
#RI-3304, #RI-3308, #RI-3170, #RI-3304, #RI-3311

zalenskiSofteq:
* #RI-3304 - JSON key is not updated when user changes value in CLI
* #RI-3308 - Clicking on search keys button user gets "Something went wrong" 
* #RI-3170 - Stream grid is brocken when user opens full screen
* #RI-3304 - JSON key is not updated when user changes value in CLI
* #RI-3311 - Promotion message should be displayed only once when user opens CLI


AmirAllayarovSofteq (approved):
LGTM

PR #948, opened 2022-07-20
#RI-3322, #RI-3317

egor-zalenski:
* #RI-3322 - Stream page crashes when user deletes field with big field name
* #RI-3317 - Stream value is not visible if field name contains two "-" symbols


AmirAllayarovSofteq (approved):
LGTM

PR #957, opened 2022-07-22
delete e2etest exe job

vlad-dargel:


.circleci/config.yml
```diff
@@ -1122,10 +1122,3 @@ workflows:
           report: true
           requires:
             - Build app - Linux (stage)
-      # e2e desktop tests on exe build
-      - e2e-exe:
```
zalenskiSofteq, line 1126: I think better comment this test
Someday it will return

.circleci/config.yml
```diff
@@ -1122,10 +1122,3 @@ workflows:
...
-          name: E2ETest (exe) - Nightly
-          parallelism: 4
-          report: true
-          requires:
-            - Build app - Windows (stage)
```
AlenaNaboka, line 1131: Can you  please just comment exe job? We are going to add windows tests when devops set environment for them. 
Like here: https://github.com/RedisInsight/RedisInsight/pull/927/files

vlad-dargel: sure

.circleci/config.yml
vlad-dargel: agree. thanks!

zalenskiSofteq (approved):

PR #961, opened 2022-07-25
e2e expand collapse large values

vlad-dargel:


tests/e2e/helpers/api/api-keys.ts
```diff
@@ -0,0 +1,128 @@
...
+    HashKeyParameters,
+    ListKeyParameters,
+    SetKeyParameters,
+    SortedSetKeyParameters,
+    StreamKeyParameters
+} from '../../pageObjects/browser-page';
+import { getDatabaseByName } from './api-database';
+
+const endpoint = apiUrl;
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'; // lgtm[js/disabling-certificate-validation]
```
ofersteinberg, line 15: Don't you already use it in past? Maybe you come to move it to a common place and import it?

tests/e2e/tests/regression/browser/large-key-details-values.e2e.ts
```diff
@@ -0,0 +1,129 @@
...
+        await addListKeyApi(listKeyParameters, ossStandaloneConfig);
+    })
+    .afterEach(async() => {
+        //Delete database
+        await deleteKeysApi(keyNames, ossStandaloneConfig);
+        await deleteStandaloneDatabaseApi(ossStandaloneConfig);
+    });
+test('Verify that user can click on a row to expand it if any of its cells contains a value which is truncated.', async t => {
+    const entryFieldLong = browserPage.streamEntryFields.nth(1).parent(1);
+    const entryFieldSmall = browserPage.streamEntryFields.nth(0).parent(1);
```
ofersteinberg, line 47: Can't you use data-id?

vlad-dargel: this is a component common to all virtualized table cells, and for all other tables, it is better not to add data-test-id
this component does not know the context of where it is being drawn

tests/e2e/helpers/api/api-keys.ts
vlad-dargel: moved to conf.ts
thanks!

tests/e2e/tests/regression/browser/large-key-details-values.e2e.ts
ofersteinberg: Notice it can be break

tests/e2e/helpers/api/api-keys.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/**
+ * Add Stream key
+ * @param keyParameters The key parameters
+ * @param databaseParameters The database parameters
+ */
+export async function addStreamKeyApi(keyParameters: StreamKeyParameters, databaseParameters: AddNewDatabaseParameters): Promise<void> {
+    const databaseId = await getDatabaseByName(databaseParameters.databaseName);
+    const response = await request(endpoint).post(`/instance/${databaseId}/streams`)
```
ofersteinberg, line 41: You can create one generic function for all the below functions.
See that `/instance/${databaseId}/` is common, and the last part is changed.
WDYT?

ofersteinberg (approved):

tests/e2e/helpers/api/api-keys.ts
vlad-dargel: they are related to different types, so better to leave as it is

PR #964, opened 2022-07-27
#RI-3202_add command type in event

AmirAllayarovSofteq:


redisinsight/api/src/modules/cli/services/cli-business/cli-business.service.ts
```diff
@@ -154,12 +157,16 @@ export class CliBusinessService {
       const reply = await this.cliTool.execCommand(clientOptions, command, args, replyEncoding);
 
       this.logger.log('Succeed to execute redis CLI command.');
+
+      const commandType = await this.checkIsCoreCommand(command) ? 'core' : 'module';
```
zalenskiSofteq, line 161: I think better to do a enum for 'core' and 'module' and use it everywhere

zalenskiSofteq (dismissed):

zalenskiSofteq (approved):

PR #1012, opened 2022-08-12
#RI-2977-add workbench raw mode

AmirAllayarovSofteq:


redisinsight/ui/src/components/query-card/QueryCardHeader/QueryCardHeader.tsx
```diff
@@ -75,11 +79,14 @@ const QueryCardHeader = (props: Props) => {
   }
 
   const sendEvent = (event: TelemetryEvent, query: string, additionalData: object = {}) => {
+    const workbenchMode = localStorageService?.get(BrowserStorageItem.workbenchMode)
```
romansergeenkosofteq, line 82: why do you get mode from storage if you have mode as props variable?

redisinsight/ui/src/components/query/Query/Query.tsx
```diff
@@ -460,6 +488,29 @@ const Query = (props: Props) => {
...
+            data-testid="change-mode-tooltip"
+          >
+            {/* <EuiButtonIcon
+              onClick={() => setMode(mode === 'ASCII' ? 'RAW' : 'ASCII')}
+              disabled={loading}
+              iconType="playFilled"
+              className={cx(styles.submitButton, { [styles.submitButtonLoading]: loading })}
+              aria-label="change mode"
+              data-testid="btn-change-mode"
+            /> */}
```
romansergeenkosofteq, line 503: please, remove commented code

redisinsight/ui/src/telemetry/telemetryUtils.ts
```diff
@@ -56,6 +56,7 @@ const sendEventTelemetry = (payload: ITelemetrySendEvent) => {
   // for analytics is granted or not.
   // If permissions not granted anonymousId includes "UNSET" value without any user identifiers.
   const { event, eventData = {}, nonTracking = false } = payload
+  console.log(event, eventData)
```
romansergeenkosofteq, line 59: remove console.log

redisinsight/ui/src/pages/workbench/components/wb-view/WBViewWrapper.tsx
```diff
@@ -134,10 +136,12 @@ const WBViewWrapper = () => {
     multiCommands: string[] = [],
   ) => {
     const { connectionType, host, port } = state.instance
+    const mode = localStorageService.get(BrowserStorageItem.workbenchMode)
```
romansergeenkosofteq, line 139: you have several places where you get this mode from localStorage. Please, move it to one place.

redisinsight/ui/src/components/query-card/QueryCardHeader/QueryCardHeader.tsx
AmirAllayarovSofteq: Because it is props for query card, but for sending command we use actual editor's "raw mode" state

redisinsight/api/src/modules/workbench/providers/workbench-commands.executor.ts
```diff
@@ -21,16 +21,19 @@ import {
...
 import { WorkbenchAnalyticsService } from '../services/workbench-analytics/workbench-analytics.service';
 
 @Injectable()
 export class WorkbenchCommandsExecutor {
   private logger = new Logger('WorkbenchCommandsExecutor');
 
-  private formatter = new RawFormatterStrategy();
+  private rawFormatter = new RawFormatterStrategy();
+
+  private utf8Formatter = new UTF8FormatterStrategy();
```
arthosofteq, line 36: We should not define each strategy as class property.

redisinsight/api/src/modules/workbench/providers/workbench-commands.executor.ts
```diff
@@ -67,15 +73,20 @@ export class WorkbenchCommandsExecutor {
...
-      const response = this.formatter.format(
-        await this.redisTool.execCommand(clientOptions, command, args, replyEncoding),
-      );
+      const response = mode === WorkbenchMode.ASCII
+        ? this.rawFormatter.format(
+          await this.redisTool.execCommand(clientOptions, command, args, replyEncoding),
+        )
+        : this.utf8Formatter.format(
+          await this.redisTool.execCommand(clientOptions, command, args, replyEncoding),
+        );
```
arthosofteq, line 89: It will be better to have common interface for strategies and work with it like IFormatter.format. We should not check mode here.

redisinsight/api/src/modules/workbench/providers/workbench-commands.executor.ts
```diff
@@ -144,7 +157,9 @@ export class WorkbenchCommandsExecutor {
 
       return {
         ...rest,
-        response: this.formatter.format(rest.response),
+        response: mode === WorkbenchMode.Raw
+          ? this.utf8Formatter.format(rest.response)
+          : this.rawFormatter.format(rest.response),
```
arthosofteq, line 162: again here. we should not check mode in this place. It will be better to determine and set formatter somewhere above. this formatter shold implement some interface. so you can write something like this:
```
const formater = { determine formater based on type and return new FormaterUTF8... } // this formater should implement IFormater interface

then below in the code you can simply do formater.format() without care about anything 
```

redisinsight/api/src/utils/cli-helper.ts
```diff
@@ -264,6 +264,11 @@ export const getASCIISafeStringFromBuffer = (reply: Buffer): string => {
   return result;
 };
 
+export const getUTG8FromBuffer = (reply: Buffer): string => {
+  const decoder = new TextDecoder('utf-8');
+  return decoder.decode(new Uint8Array(reply));
+};
+
```
arthosofteq, line 271: Is there a reason to use TextDecoder and Uint8Array here in node?
Could you please check api/src/common folder. we have transformers there. It will be better to reuse them

arthosofteq (changes requested):

redisinsight/api/src/utils/cli-helper.ts
AmirAllayarovSofteq: TextDecoder still use ascii formatter...

AmirAllayarovSofteq: and in api/src/common/ functions which working with strings, but we get buffer in response

romansergeenkosofteq (approved):

redisinsight/api/src/modules/cli/services/cli-business/output-formatter/output-formatter-manager.ts
```diff
@@ -7,13 +8,13 @@ export class OutputFormatterManager {
   private strategies = {};
 
   addStrategy(
-    name: CliOutputFormatterTypes,
+    name: CliOutputFormatterTypes | RunQueryMode,
```
arthosofteq, line 11: Why we need multiple types here?

redisinsight/api/src/modules/workbench/dto/create-command-executions.dto.ts
```diff
@@ -4,12 +4,7 @@ import {
 } from 'class-validator';
 import { Type } from 'class-transformer';
 import { ClusterSingleNodeOptions } from 'src/modules/cli/dto/cli.dto';
-
-export enum ClusterNodeRole {
```
arthosofteq, line 8: why you deleted this? do we use it somewhere?

redisinsight/api/src/utils/cli-helper.ts
arthosofteq: Ok. let's not use transformer from common for now (I'm not happy with this) BUT let's implement conversion in a right way then. We don't need double conversion here. you can simpy do `reply.toString('utf8')`

arthosofteq (changes requested):
Nice to see formatter strategies here. However some moments still not clear for me

redisinsight/api/src/modules/workbench/providers/workbench-commands.executor.spec.ts
```diff
@@ -17,7 +22,7 @@ import {
 } from 'src/modules/cli/constants/errors';
 import { ICliExecResultFromNode, RedisToolService } from 'src/modules/shared/services/base/redis-tool.service';
 import { CommandsService } from 'src/modules/commands/commands.service';
-import { WorkbenchAnalyticsService } from '../services/workbench-analytics/workbench-analytics.service';
+import { FormatterManager, IFormatterStrategy, FormatterTypes } from 'src/common/transformers';import { WorkbenchAnalyticsService } from '../services/workbench-analytics/workbench-analytics.service';
```
arthosofteq, line 25: 2 imports in a single line here.

arthosofteq (approved):
Looks good

PR #1024, opened 2022-08-16
E2E Format switcher

vlad-dargel:


tests/e2e/pageObjects/browser-page.ts
```diff
@@ -892,3 +912,13 @@ export type ListKeyParameters = {
     keyName: string,
     element: string
 };
+
+/**
+ * Keys Data parameters
+ * @param textType The type of the key
+ * @param keyName The name of the key
+ */
+export type keyData = {
```
ofersteinberg, line 921: Where do you use this type? Also, in type names, we use PascalCase

vlad-dargel: Thanks, fixed!
I use it in addKeysViaCli and deleteKeysViaCli functions as parameter type

ofersteinberg (approved):

PR #1057, opened 2022-08-23
E2e/regression

AlenaNaboka:


tests/e2e/tests/smoke/database/add-db-from-welcome-page.e2e.ts
```diff
@@ -1,27 +1,53 @@
...
         //Add database from Welcome page
         await addNewStandaloneDatabase(ossStandaloneConfig);
         await t.expect(myRedisDatabasePage.dbNameList.withExactText(ossStandaloneConfig.databaseName).exists).ok('The database adding', { timeout: 10000 });
     });
+test
+    .meta({  env: env.web, rte: rte.standalone })('Verify that all the links are valid from Welcome page', async t => {
+        // Verify build from source link
+        await t.expect(addRedisDatabasePage.buildFromSource.visible).ok('Build from source link');
+        await t.click(addRedisDatabasePage.buildFromSource);
+        await t.expect(getPageUrl()).eql('https://developer.redis.com/create/from-source/?utm_source=redis&utm_medium=app&utm_campaign=redisinsight', 'Build from source page');
```
vlad-dargel, line 37: I think that you can use these links as variables for better code readability
e.g. const dockerPage = 'https://developer.redis.com/create/from-source/?utm_source=redis&utm_medium=app&utm_campaign=redisinsight';

tests/e2e/tests/smoke/database/add-db-from-welcome-page.e2e.ts
```diff
@@ -1,27 +1,53 @@
...
-        await myRedisDatabasePage.deleteAllDatabases();
         await t.expect(addRedisDatabasePage.welcomePageTitle.exists).ok('The welcome page title');
         //Add database from Welcome page
         await addNewStandaloneDatabase(ossStandaloneConfig);
         await t.expect(myRedisDatabasePage.dbNameList.withExactText(ossStandaloneConfig.databaseName).exists).ok('The database adding', { timeout: 10000 });
     });
+test
+    .meta({  env: env.web, rte: rte.standalone })('Verify that all the links are valid from Welcome page', async t => {
+        // Verify build from source link
+        await t.expect(addRedisDatabasePage.buildFromSource.visible).ok('Build from source link');
```
vlad-dargel, line 35: please change error messages in expects from 'Build from source link' to something like 'Build from source link not visible' and the same above

vlad-dargel: and seems like you don't need these expects for buttons visibility as testCafe check it in the step below in .click() method
await t.click(addRedisDatabasePage.buildFromSource);

tests/e2e/tests/smoke/database/add-db-from-welcome-page.e2e.ts
AlenaNaboka: fixed

tests/e2e/tests/smoke/database/add-db-from-welcome-page.e2e.ts
AlenaNaboka: fixed both

vlad-dargel (approved):

PR #1074, opened 2022-08-25
[e2e] enhance command helper ui

vlad-dargel:


tests/e2e/pageObjects/cli-page.ts
```diff
@@ -141,4 +142,16 @@ export class CliPage {
             await t.expect(this.cliHelperOutputTitles.nth(i).textContent).eql(listToCompare[i], 'Results in the output contains searched value');
         }
     }
+
+    /**
+    * Check commands list
+    * @param listToCompare The list with commands to compare with opened in Command Helper
+    */
+    async checkCommandsInCommandHelper(listToCompare: string[]): Promise<void> {
```
ofersteinberg, line 150: I'm not sure how related this function is to being on page-object.
Generally, we put these functions on some `common-actions`/`utilities` files

vlad-dargel: created common-actions folder and moved these methods there
thanks!

PR #1079, opened 2022-08-25
E2E big db test fix

vlad-dargel:


tests/e2e/tests/regression/browser/filtering.e2e.ts
```diff
@@ -206,33 +206,41 @@ test
...
+        //Verify that user can use the "Scan More" button to search per another 10000 keys
+        for (let i = 10; i < 100; i += 10) {
+            // scannedValue = scannedValue + 10;
+            await t.expect(browserPage.progressKeyList.exists).notOk('Progress Bar is not displayed', { timeout: 30000 });
+            const scannedValueText = await browserPage.scannedValue.textContent;
+            const regExp = new RegExp(`${i} 00` + '.');
+            await t.expect(scannedValueText).match(regExp, `The database is automatically scanned by ${i} 000 keys`);
+            await t.doubleClick(browserPage.scanMoreButton);
+            await t.expect(browserPage.progressKeyList.exists).ok('Progress Bar is displayed', { timeout: 30000 });
+        }
```
AlenaNaboka, line 229: As I remember, this part of code was in some other e2e. Can you please create a function and reuse it?
And also please add validation for number of found keys that should be increased

vlad-dargel: updated, thanks!

AlenaNaboka (approved):

PR #1089, opened 2022-08-30
Bugfix/ri 3432 workbench commands order

AmirAllayarovSofteq:


redisinsight/api/src/modules/workbench/workbench.service.ts
```diff
@@ -61,8 +61,8 @@ export class WorkbenchService {
     clientOptions: IFindRedisClientInstanceByOptions,
     dto: CreateCommandExecutionsDto,
   ): Promise<CommandExecution[]> {
-    return Promise.all(
-      dto.commands.map(async (command) => await this.createCommandExecution(clientOptions, { ...dto, command })),
+    return this.syncPromiseAll(
```
arthosofteq, line 64: I'm not sure that we must do this. Could you explain why we want to wait previous command response before starting the next one? And what will be the difference between executing 100 commands "1 by 1" and in the "pipeline" mode selected by user?

AmirAllayarovSofteq: To make sure that the commands are executed in the order in which the user sent them. For example if user will sent
set 1 1
set 1 2
set 1 3 
We don't know what value will have string with name "1". Because we execute commands in same time(pipeline more than 2)

arthosofteq: I understand your point but could you investigate further? https://github.com/luin/ioredis/issues/416#issuecomment-275306892
I want to understand how:
1. ioredis sending this commands (I'm assuming they will be sent to redis in the same order)
2. how redis processes bunch of commands (I assume there is sync proccess FIFO)

arthosofteq: Check this please and add proper test if needed

PR #1090, opened 2022-08-30
[E2E] notification categories

AlenaNaboka:


tests/e2e/tests/critical-path/database/clone-databases.e2e.ts
```diff
@@ -0,0 +1,118 @@
...
+    .beforeEach(async t => {
+        await acceptLicenseTerms();
+        await addNewStandaloneDatabaseApi(ossStandaloneConfig);
+        await t.eval(() => location.reload());
+    });
+test
+    .before(async t => {
+        await acceptLicenseTerms();
+        await addNewStandaloneDatabaseApi(ossStandaloneConfig);
+        await t.eval(() => location.reload());
```
ofersteinberg, line 30: One is redundant

AlenaNaboka: removed common fixture, thanks

tests/e2e/tests/critical-path/database/clone-databases.e2e.ts
```diff
@@ -0,0 +1,112 @@
...
+    })
+    .after(async() => {
+        // Delete databases
+        const dbNumber = await myRedisDatabasePage.dbNameList.withExactText(ossStandaloneConfig.databaseName).count;
+        for (let i = 0; i < dbNumber; i++) {
+            await deleteStandaloneDatabaseApi(ossStandaloneConfig);
+        }
+    })
+    .meta({ rte: rte.standalone })('Verify that user can clone Standalone db', async t => {
```
vlad-dargel, line 34: seems like all tests here use  .meta({ rte: rte.standalone })
so it can be moved at the top to the fixture meta

vlad-dargel: or rte above can be changed to rte: rte.ossCluster and rte: rte.sentinel

ofersteinberg (dismissed):

tests/e2e/tests/critical-path/notifications/notification-center.e2e.ts
```diff
@@ -67,6 +75,10 @@ test('Verify that user can open notification center by clicking on icon and see
         await t.expect(notificationPage.notificationTitle.withExactText(jsonNotifications[i].title).exists).ok('Displayed title');
         await t.expect(notificationPage.notificationBody.withExactText(jsonNotifications[i].body).exists).ok('Displayed body');
         await t.expect(notificationPage.notificationDate.withExactText(await notificationPage.convertEpochDateToMessageDate(jsonNotifications[i])).exists).ok('Displayed date');
+        if (!jsonNotifications[i].category !== undefined) {
```
vlad-dargel, line 78: can you please add comment what checked here? for readability

tests/e2e/tests/critical-path/database/clone-databases.e2e.ts
AlenaNaboka: fixed, thanks

tests/e2e/tests/critical-path/notifications/notification-center.e2e.ts
AlenaNaboka: Seems like no explanation needed here. We just have a conditions that we can work with category only in case if it exists in the notification. Don't you think so?

vlad-dargel (approved):

PR #1115, opened 2022-09-05
set very low "component waiting time" to decrease idle time during te…

arthosofteq:
…sts are running

### DIFF
--- a/tests/e2e/helpers/database.ts
+++ b/tests/e2e/helpers/database.ts
@@ -193,7 +193,7 @@ export async function acceptLicenseTermsAndAddRECloudDatabase(databaseParameters
 export async function acceptLicenseTerms(): Promise<void> {
     await t.maximizeWindow();
     await userAgreementPage.acceptLicenseTerms();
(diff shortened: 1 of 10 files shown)


arthosofteq:
@vlad-dargel Here is a suggestion from my side how to speed up e2e tests. 
I found that we are waiting for 5 seconds before each test case to determine that agreements popup is not shown, also we are waiting for 5 seconds each time when we want to check that some component appeared. 
I believe that few tests failed due to this small timeout but there are not much (< 20 as I can see). Can we improve these tests (rewrite somehow or add additional wait time in tests to fix them)? It wight save ~30-50% time for all tests execution

vlad-dargel:
> @vlad-dargel Here is a suggestion from my side how to speed up e2e tests. I found that we are waiting for 5 seconds before each test case to determine that agreements popup is not shown, also we are waiting for 5 seconds each time when we want to check that some component appeared. I believe that few tests failed due to this small timeout but there are not much (< 20 as I can see). Can we improve these tests (rewrite somehow or add additional wait time in tests to fix them)? It wight save ~30-50% time for all tests execution

vlad-dargel:
I will check this and reply here, thanks

PR #1170, opened 2022-09-19
#RI-3543-3544-3546-bugfix

AmirAllayarovSofteq:
RI-3543 - [FE][Group Mode] Tooltip style is not the same as for other tooltips
RI-3544 - [FE][Group Mode] Command result is not on the full screen
RI-3546 - [FE][Group Mode] Group property has False value even if commands were run in group

### DIFF
--- a/redisinsight/ui/src/components/keyboard-shortcut/KeyboardShortcut.tsx
+++ b/redisinsight/ui/src/components/keyboard-shortcut/KeyboardShortcut.tsx
@@ -6,23 +6,27 @@ import { EuiBadge, EuiText } from '@elastic/eui'
 import styles from './styles.module.scss'
 
 export interface Props {
-  items: (string | JSX.Element)[];
-  separator?: string;
-  transparent?: boolean;
+  items: (string | JSX.Element)[]
+  separator?: string
+  transparent?: boolean
+  className?: string
 }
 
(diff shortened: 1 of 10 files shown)


redisinsight/ui/src/components/keyboard-shortcut/KeyboardShortcut.tsx
```diff
@@ -6,23 +6,27 @@ import { EuiBadge, EuiText } from '@elastic/eui'
...
+  return (
+    <div className={styles.container}>
+      {
         items.map((item: string | JSX.Element, index: number) => (
           <div key={isString(item) ? item : item?.props?.children}>
-            { (index !== 0) && <div className={styles.separator}>{separator}</div> }
+            {(index !== 0) && <div className={styles.separator}>{separator}</div>}
             <EuiBadge className={cx(styles.badge, { [styles.transparent]: transparent })}>
-              <EuiText size="s">{item}</EuiText>
+              <EuiText className={className} size="s">{item}</EuiText>
```
romansergeenkosofteq, line 24: I think better to rename this className prop:
maybe `badgeTextClassName`
or as object: `
classNames: {
 badgeText: string
}
`

romansergeenkosofteq (approved):

PR #1204, opened 2022-09-26
fixed tests

vlad-dargel:


tests/e2e/tests/regression/browser/filtering.e2e.ts
```diff
@@ -149,7 +149,6 @@ test
     });
 test('Verify that when user clicks on “clear” control and filter per key name is applied filter is reset and rescan initiated', async t => {
         keyName = `KeyForSearch${chance.word({ length: 50 })}`;
-        keyName2 = `KeyForFearch${chance.word({ length: 10 })}`;
```
AlenaNaboka, line 152: keyName2 is used n the line 
`await browserPage.searchByKeyName(keyName2);`
Are you sure that it's needed to delete with line?

PR #1229, opened 2022-10-03
Allow quering keys without infos + additional endpoint to fetch keys …

arthosofteq:
…infos when needed


AmirAllayarovSofteq (approved):
LGTM

PR #1265, opened 2022-10-07
data interfaces

arthosofteq:


redisinsight/api/provider-interfaces/models/agreements.ts
```diff
@@ -0,0 +1,14 @@
+export class Agreements {
```
dannycohen, line 1: Q: why use named properties as opposed to a hashmap with string keys and boolean values ?

redisinsight/api/provider-interfaces/models/agreements.ts
```diff
@@ -0,0 +1,14 @@
...
+  // as a userId
```
dannycohen, line 2: what does this comment mean ? 
is it "add a user id" ?

redisinsight/api/provider-interfaces/models/agreements.ts
```diff
@@ -0,0 +1,14 @@
...
+  id: string; // todo: Discuss
```
dannycohen, line 3: What is the meaning of an `id` of an agreement (that is distinct from a `userId`) ?

redisinsight/api/provider-interfaces/models/ca-certificate.ts
```diff
@@ -0,0 +1,10 @@
+// todo: discuss
+// probably not needed for cloud
+// if needed - we must introduce userId here
+export class CaCertificate {
```
dannycohen, line 4: The `CaCertificate` applies to all Redis databases and is provided by Redis itself. 
We will need the model, but in the provider we only need the `get` which will always return the same `CaCertificate` regardless of parameter.

redisinsight/api/provider-interfaces/models/ca-certificate.ts
```diff
@@ -0,0 +1,10 @@
...
+  id: string;
```
dannycohen, line 5: I guess `id` is a remnant of the implementation in SQLITE on-premise. right ?

redisinsight/api/provider-interfaces/models/client-certificate.ts
```diff
@@ -0,0 +1,12 @@
+// todo: discuss
+// probably not needed for cloud
```
dannycohen, line 2: The client cert is stored in the Redis Cloud SM database so we can retrieve it from there.
We will need the model, but in the provider we only need the `get` which will return the same Client Certificate for the specific database.

So we need a way to identify the database associated with the client certificate. 
This can be done by piggybacking on the `id` property (i.e. `ClientCertificate.id` will hold the `databaseId`), or we can add an additional `databaseId` property.

WDYT ?

redisinsight/api/provider-interfaces/models/command-execution.ts
```diff
@@ -0,0 +1,20 @@
+// todo: discuss. introduce userId here?
```
dannycohen, line 1: I think we should add a concept of `sessionId` (which identifies user + database + specific session).
This way, when we run a command, we'll know exactly where it came from (from which user, connected to which database, and in the context of which specific session).

WDYT ?

redisinsight/api/provider-interfaces/models/command-execution.ts
```diff
@@ -0,0 +1,20 @@
...
+export class CommandExecution {
+  id: string;
+
+  databaseId: string;
+
+  command: string;
```
dannycohen, line 7: We need to discuss security aspects of accepting the command.

Unlike in the on-premise scenario, in the cloud scenario we need to be very careful about receiving the command and running it as-is...

I am assuming right now we pass the command as-is. Is that correct ? 

(Definitely we need to discuss in depth, but as a starter, I think it needs to be encoded (URL or even Base64 ?))

redisinsight/api/provider-interfaces/models/command-execution.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  mode?: string;
+
+  result: string;
+
+  role?: string;
```
dannycohen, line 13: what is role ? is it an `enum` ?

redisinsight/api/provider-interfaces/models/command-execution.ts
```diff
@@ -0,0 +1,20 @@
...
+export class CommandExecution {
+  id: string;
+
+  databaseId: string;
+
+  command: string;
+
+  mode?: string;
+
+  result: string;
```
dannycohen, line 11: what is result ?

redisinsight/api/provider-interfaces/models/command-execution.ts
```diff
@@ -0,0 +1,20 @@
+// todo: discuss. introduce userId here?
+export class CommandExecution {
+  id: string;
+
+  databaseId: string;
+
+  command: string;
+
+  mode?: string;
```
dannycohen, line 9: what is the `mode` and what are the valid values ? is it an `enum` ?

redisinsight/api/provider-interfaces/models/command-execution.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  result: string;
+
+  role?: string;
+
+  nodeOptions?: string;
```
dannycohen, line 15: explain ?

redisinsight/api/provider-interfaces/models/command-execution.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  encryption: string;
```
dannycohen, line 17: explain ? supported values ? (`enum`?)

redisinsight/api/provider-interfaces/models/command-execution.ts
```diff
@@ -0,0 +1,20 @@
+// todo: discuss. introduce userId here?
+export class CommandExecution {
```
dannycohen, line 2: @arthosofteq - I'm adding lots of pseudo-documentation questions below. I suggest you add comments to the interface describing each property's purpose and supported values.

redisinsight/api/provider-interfaces/models/database.ts
```diff
@@ -0,0 +1,61 @@
...
+export class Database {
+  id: string;
+
+  host: string;
+
+  port: number;
+
+  name: string;
+
+  db: number;
```
dannycohen, line 13: This is virtual database ? so in Redis cloud its always `0` ?

redisinsight/api/provider-interfaces/models/database.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  connectionType: string;
```
dannycohen, line 15: supported values ? is it an `enum` ?

redisinsight/api/provider-interfaces/models/database.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  username: string;
+
+  password: string;
+
+  lastConnection: Date;
+
+  provider: string;
```
dannycohen, line 23: purpose and possible values ? (`enum` ? )

redisinsight/api/provider-interfaces/models/database.ts
```diff
@@ -0,0 +1,61 @@
+import { CaCertificate } from './ca-certificate';
+import { ClientCertificate } from './client-certificate';
+
+export class Database {
```
dannycohen, line 4: In SaaS there will be no listing of databases; just a single database per session, with the values retrieved from the SM using DAS (Data Access Service). 
This means the provider will only implement `get` operations

redisinsight/api/provider-interfaces/models/database.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  sentinel?: {
+    name: string,
+    username: string,
+    password: string,
+  };
+
+  tls?: boolean;
+
+  tlsServername?: string; // sni
```
dannycohen, line 46: `sni` ?

redisinsight/api/provider-interfaces/models/notification.ts
```diff
@@ -0,0 +1,19 @@
+export class Notification {
+  // todo: not implemented. need to add
```
dannycohen, line 2: what does this mean ?

redisinsight/api/provider-interfaces/models/notification.ts
```diff
@@ -0,0 +1,19 @@
...
+  userId: string;
+
+  type: string;
+
+  // todo: Discuss. Used as id currently. Probably need to add userId field and change unique constraint
```
dannycohen, line 7: currently `timestamp` is used as id  ?

redisinsight/api/provider-interfaces/models/plugin-state.ts
```diff
@@ -0,0 +1,14 @@
+// unique constraint: commandExecutionId + visualizationId
+export class PluginState {
```
dannycohen, line 2: To Discuss: Need to align with Command Execution storage.

redisinsight/api/provider-interfaces/models/server.ts
```diff
@@ -0,0 +1,6 @@
+export class Server {
```
dannycohen, line 1: I do not think this is applicable to Redis Cloud. 
(We may want to display information on subscription, but a subscription is different and distinct from server (or cluster))

redisinsight/api/provider-interfaces/models/settings.ts
```diff
@@ -0,0 +1,10 @@
+export class Settings {
+  // as a userId
```
dannycohen, line 2: Yes, I think the settings are associated with a specific `userId`

redisinsight/api/provider-interfaces/models/settings.ts
```diff
@@ -0,0 +1,10 @@
...
+  id: string; // todo: Discuss
+
+  theme: string;
```
dannycohen, line 5: `enum` ?

dannycohen (commented):
@arthosofteq - see comments (only on `models` currently).

PR #1285, opened 2022-10-17
#RI-3612-zset infinity score

AmirAllayarovSofteq:


redisinsight/api/src/common/validators/zset-score.validator.ts
```diff
@@ -0,0 +1,17 @@
...
+} from 'class-validator';
+
+@ValidatorConstraint({ name: 'RedisStringValidator', async: true })
+export class ZSetScoreValidator implements ValidatorConstraintInterface {
+  async validate(value: any) {
+    return value === 'inf' || value === '-inf' || isNumber(value);
+  }
+
+  defaultMessage(args: ValidationArguments) {
+    return `${args.property || 'field'} must be a string or a number`;
```
arthosofteq, line 15: must be a string? "hello world" will be not a valid value here...

redisinsight/api/src/modules/browser/services/z-set-business/z-set-business.service.ts
```diff
@@ -277,8 +277,10 @@ export class ZSetBusinessService {
           BrowserToolZSetCommands.ZScore,
           [keyName, member],
         );
+        const formattedScore = isNaN(parseFloat(score)) ? String(score) : parseFloat(score);
```
arthosofteq, line 280: Can we move this to some function to not duplicate logic everywhere?

redisinsight/api/src/modules/browser/services/z-set-business/z-set-business.service.ts
```diff
@@ -406,11 +407,13 @@ export class ZSetBusinessService {
     const result: ZSetMemberDto[] = [];
     while (reply.length) {
       const member = reply.splice(0, 2);
+      const score = isNaN(parseFloat(member[1])) ? String(member[1]) : parseFloat(member[1]);
       result.push(plainToClass(ZSetMemberDto, {
```
arthosofteq, line 411: we are using "plainToClass" here. Maybe it is better to create own transformer and `isNaN(parseFloat(member[1])) ? String(member[1]) : parseFloat(member[1]);` will be done automatically inside?
P.S. See comment above

redisinsight/api/test/api/z-set/PATCH-instance-id-zSet.test.ts
```diff
@@ -22,9 +22,9 @@ const dataSchema = Joi.object({
   member: Joi.object().keys({
     name: Joi.string().required(),
     // todo: allow(true) - is incorrect but will be transformed to number by BE. Investigate/fix it
-    score: Joi.number().required().allow(true),
+    score: Joi.number().required().allow('inf', '-inf').label('.score'),
```
arthosofteq, line 25: it is better to create own type here the same way as for `redisString` so it will look like `Joi.zSetScore().required()` or something similar

redisinsight/api/test/api/z-set/POST-instance-id-zSet-get_members.test.ts
```diff
@@ -38,7 +39,7 @@ const responseSchema = Joi.object().keys({
   total: Joi.number().integer().required(),
   members: Joi.array().items(Joi.object().keys({
     name: JoiRedisString.required(),
-    score: Joi.number().required(),
+    score: Joi.number().required().allow('inf', '-inf'),
```
arthosofteq, line 42: the same

redisinsight/api/test/api/z-set/POST-instance-id-zSet-search.test.ts
```diff
@@ -41,7 +41,7 @@ const responseSchema = Joi.object().keys({
   nextCursor: Joi.number().integer().required(),
   members: Joi.array().items(Joi.object().keys({
     name: JoiRedisString.required(),
-    score: Joi.number().required(),
+    score: Joi.number().required().allow('inf', '-inf'),
```
arthosofteq, line 44: the same

redisinsight/api/src/modules/browser/dto/z-set.dto.ts
```diff
@@ -77,13 +77,12 @@ export class ZSetMemberDto {
...
-    type: Number,
+    type: Number || String,
     default: 1,
   })
   @IsDefined()
-  @IsNumber({ maxDecimalPlaces: 15 })
-  @Type(() => Number)
-  score: number;
+  @isZSetScore()
+  score: number | 'inf' | '-inf';
```
arthosofteq, line 85: It is better to create a type (like `type ZSetScore = ...`) and use it everywhere

arthosofteq (approved):
Bakend looks good. I've added few comments. It will be nice to address them but not necessary in case if we must move forward asap.

zalenskiSofteq (approved):

PR #1291, opened 2022-10-18
#RI-3650 - [BE] status error on get big key

zalenskiSofteq:
#RI-3650 - [BE] status error on get big key

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,7 @@ lerna-debug.log*
 # Tests
 /coverage
 /.nyc_output
-/redisinsight/api/test/test-runs/coverage
+**/coverage
 
 # IDEs and editors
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -16,6 +16,7 @@ lerna-debug.log*
 # Tests
 /coverage
 /.nyc_output
+/redisinsight/api/test/test-runs/coverage
```
arthosofteq, line 19: is it better to use `coverage/` or `**/coverage` instead of full path's?

arthosofteq (approved):
nice

.gitignore
zalenskiSofteq: thanks, changed to `**/coverage`

PR #1293, opened 2022-10-18
#RI-3656-add plugins to static server

AmirAllayarovSofteq:


arthosofteq (approved):
Fine for now. But seems like you didn't find the root cause.
This folder should be exposed by 
`    ServeStaticModule.forRoot({
      serveRoot: SERVER_CONFIG.staticUri,
      rootPath: join(PATH_CONFIG.staticDir),
      exclude: ['/api/**'],
      serveStaticOptions: {
        fallthrough: false,
      },
    }),
 `
from the `app.module.ts`
previously it worked. seems like after update ServeStaticModule was changed.

arthosofteq (approved):

PR #1347, opened 2022-11-02
#RI-3751-add command execution time

AmirAllayarovSofteq:


redisinsight/api/src/modules/workbench/workbench.service.ts
```diff
@@ -79,9 +85,14 @@ export class WorkbenchService {
...
       }
       const result = await this.commandsExecutor.sendCommand(clientOptions, { ...dto, command });
+      executionTimeInNanoseconds += process.hrtime(startCommandExecutionTime)[1];
       return ({ ...result[0], command });
     }));
 
+    if (executionTimeInNanoseconds !== 0) {
+      commandExecution.executionTime = Math.round(executionTimeInNanoseconds / 1000);
+    }
+
```
arthosofteq, line 95: I'm not sure that using `hrtime()` is a correct way here. 
At first it is legacy function - https://nodejs.org/api/process.html#processhrtimetime. If you want to use  use node:process then you should look into `hrtime.bugint()`
Anothe thing here that you get `nanoseconds` and then you've added additional `check` and `division` to get `milliseconds` and after all you are `rounding` value. Why? Will it be more easy to use `Data.now()` and calculate difference between it? Simple it could be achieved in 2 rows `const start  = Date.now()` and  `commandExecution.executionTime = Date.now() - statrt`. That is it.

arthosofteq (changes requested):

romansergeenkosofteq (approved):

redisinsight/api/src/modules/workbench/workbench.service.spec.ts
```diff
@@ -136,8 +136,10 @@ describe('WorkbenchService', () => {
 
   describe('createCommandExecution', () => {
     it('should successfully execute command and save it', async () => {
-      expect(await service.createCommandExecution(mockClientOptions, mockCreateCommandExecutionDto))
-        .toEqual(mockCommandExecutionToRun);
+      const result = await service.createCommandExecution(mockClientOptions, mockCreateCommandExecutionDto);
+      // can't predict execution time
+      expect(result).toMatchObject(mockCommandExecutionToRun);
```
arthosofteq, line 141: `toMatchObject` nice! didn't know this useful feature :)

arthosofteq (approved):

arthosofteq (approved):

PR #1383, opened 2022-11-09
#RI-3736, #RI-3738

zalenskiSofteq:
* #RI-3736 - Normalize commands in CLI and Workbench
* #RI-3738 - Module name is Redijson for telemetry events


arthosofteq (approved):
Fine for now, but maybe it is better to do `toUppercase` in a single place (e..g. check `command` field inside AnalyticsService and if it is there call `toUppercase`

PR #1426, opened 2022-11-21
#RI-3844 throw gateway timeout error and close the client if no respo…

arthosofteq:
…nse in 45 seconds for ft.search command

+ add error message for global request timeout handler


AmirAllayarovSofteq (approved):
LGTM

PR #1435, opened 2022-11-22
#RI-3742-parse commands

AmirAllayarovSofteq:


redisinsight/api/src/modules/workbench/providers/workbench-commands.executor.ts
```diff
@@ -69,10 +69,12 @@ export class WorkbenchCommandsExecutor {
       nodeOptions,
       mode,
     } = dto;
-
-    const [command, ...commandArgs] = splitCliCommandLine(commandLine);
+    let command = commandLine;
```
arthosofteq, line 72: Why you assigned entire command line to command?

redisinsight/api/src/modules/workbench/providers/workbench-commands.executor.ts
```diff
@@ -69,10 +69,12 @@ export class WorkbenchCommandsExecutor {
...
 
     try {
+      const [parsedCommand, ...commandArgs] = splitCliCommandLine(commandLine);
+      command = parsedCommand;
+
```
arthosofteq, line 77: Did you move parsing command line into try catch block to handle error? Did you add additional handling for it? I'm not sure that I understand...

arthosofteq (changes requested):

redisinsight/api/src/modules/workbench/providers/workbench-commands.executor.ts
AmirAllayarovSofteq: Yep. I moved parsing command to try catch block, reason: do not throw Error when we can't parse command and args from command line.
For example: in pipeline or in group mode user sent
`set 1 1
json.set "key" . "{}`
we parsed and executed `set 1 1`, but cant parse and execute `json.set "key" . "{}`. In current realization we handle it as 500 error, but we executed `set 1 1` command

redisinsight/api/src/modules/workbench/providers/workbench-commands.executor.ts
AmirAllayarovSofteq: Because if we can't parse command in my opinion we still have to send something in analytics.

arthosofteq: We should not send everything. Never. There might be sensitive information.

redisinsight/api/src/modules/workbench/providers/workbench-commands.executor.ts
arthosofteq: With your solution we will return 1 succeed and 1 failed error with 200 OK. If it was a goal - I'm fine with this.

arthosofteq: However in this case it will be better to add error handling to return `{ status: fail, response: "some reasonable error"}` instead of InternalServerException

redisinsight/api/src/modules/workbench/providers/workbench-commands.executor.ts
AmirAllayarovSofteq: That mean we have to create another try catch block without analytics where we will return only errorResults?

arthosofteq (approved):
👍

PR #1466, opened 2022-11-30
E2e/feature/ri 3571 3567 3564 3562 3560 recommendations

vlad-dargel:


tests/e2e/helpers/keys.ts
```diff
@@ -193,6 +194,39 @@ export async function populateSetWithMembers(host: string, port: string, keyArgu
     });
 }
 
+/**
+ * Populate Zset key with members
+ * @param host The host of database
+ * @param port The port of database
+ * @param keyArguments The arguments of key and its members
+ */
+ export async function populateZSetWithMembers(host: string, port: string, keyArguments: AddKeyArguments): Promise<void> {
```
ofersteinberg, line 203: Please set `number` on `port` instead of using the `Number` bellow

vlad-dargel: I use port as a string in this method because we are storing port as string in conf file for all database types:
image

so if I use port as number here then Number should be used in arguments everywhere

vlad-dargel: yes, because we add data from this config as arguments for such methods as ``populateZSetWithMembers``

ofersteinberg (approved):

PR #1497, opened 2022-12-12
#RI-3728 - Base BE implementation. Import certs by plain values

arthosofteq:
#RI-3684 - Base BE implementation. Import certs from files

### DIFF
--- a/redisinsight/api/src/modules/database-import/database-import.service.ts
+++ b/redisinsight/api/src/modules/database-import/database-import.service.ts
@@ -21,6 +21,7 @@ import {
   UnableToParseDatabaseImportFileException,
 } from 'src/modules/database-import/exceptions';
 import { ValidationException } from 'src/common/exceptions';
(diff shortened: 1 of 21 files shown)


redisinsight/api/src/modules/database-import/database-import.service.ts
```diff
@@ -195,6 +253,49 @@ export class DatabaseImportService {
...
   }
 
+  /**
+   * Try to determine connection type based on input data
+   * Should return NOT_CONNECTED when it is not possible
+   * @param data
+   */
+  static determineConnectionType(data: any = {}): ConnectionType {
+    if (data?.connectionType) {
+      switch (data.connectionType) {
```
zalenskiSofteq, line 263: may be better to use this structure?
```return (data.connectionType in ConnectionType) ? ConnectionType[data.connectionType] : ConnectionType.NOT_CONNECTED```

arthosofteq: Make sense. Thanks

PR #1498, opened 2022-12-13
#RI-3570 - add big connected clients recommendation

AmirAllayarovSofteq:


arthosofteq (approved):
Fine but not sure that "big_connected_clients" is a proper name. Maybe big_amount_of_* or a_lot_of*

PR #1503, opened 2022-12-14
#RI-3574 - add set password recommendation

AmirAllayarovSofteq:


redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -280,4 +280,43 @@ export class RecommendationProvider {
...
+    } catch (err) {
+      this.logger.error('Can not determine set password recommendation', err);
+      return null;
+    }
+  }
+
+  private async checkAuth(redisClient: Redis | Cluster): Promise<boolean> {
+    try {
+      await redisClient.sendCommand(
+        new Command('auth', ['pass']),
```
arthosofteq, line 313: what will happen if password is set to 'pass' in fact? (seems like nothing harmful, but...)
Here you just trying to catch if AUTH disabled at all?
Interesting thing here for which user you will send auth command? is here any user required?

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -280,4 +280,43 @@ export class RecommendationProvider {
...
+      );
+    } catch (err) {
+      if (err.message.includes('Client sent AUTH, but no password is set')) {
```
arthosofteq, line 316: Are you sure that such message will be for all redis builds\envs?

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -280,4 +280,43 @@ export class RecommendationProvider {
...
+  async determineSetPasswordRecommendation(
+    redisClient: Redis | Cluster,
+  ): Promise<Recommendation> {
+    if (await this.checkAuth(redisClient)) {
+      return { name: RECOMMENDATION_NAMES.SET_PASSWORD };
+    }
+
+    try {
+      const users = await redisClient.sendCommand(
+        new Command('acl', ['list'], { replyEncoding: 'utf8' }),
```
arthosofteq, line 298: what about databases or users without acl permissions?

arthosofteq (approved):
Code itself looks good. 
If this behavior was agreed and putted into requirements - approve from my side.
However I want to point some risks here (please doublecheck them).
1. This checks should not "mutate" current client - to not affect analysis job (fail for some) reason
2. Most probably `acl` will be switched off for a lot of redis dbs/users.
3. What about users without pass but with tls authentication? Should we create recommendation in this case?
4.

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
AmirAllayarovSofteq: approved by requirements. Without permission or with redis version < 6

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
AmirAllayarovSofteq: i will double check. Thanks

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
AmirAllayarovSofteq: Here i trying to check is any password set. This command will send for current connected user. If in any reason this command will executed it will not change anything. Let me double check

PR #1511, opened 2022-12-16
E2e/feature/ri 3853 show process results for imported databases

vlad-dargel:


tests/e2e/test-data/certs/client.key
```diff
@@ -0,0 +1,3 @@
+-----BEGIN PRIVATE KEY-----
```
ofersteinberg, line 1: I think you could make the whole certs in one file. But this is also a good one.

tests/e2e/test-data/rdm-full.json
```diff
@@ -0,0 +1,219 @@
+[
```
ofersteinberg, line 1: What is that file?
Please add a description on top

vlad-dargel: no, it's json file

vlad-dargel: which is only used to uploading this file in app

tests/e2e/test-data/certs/client.key
vlad-dargel: we decided to not put text to save a memory on CI (no matter what is inside this cert)

tests/e2e/pageObjects/browser-page.ts
```diff
@@ -26,8 +26,6 @@ export class BrowserPage {
     setDeleteButton = Selector('[data-testid=set-delete-btn]');
     streamDeleteButton = Selector('[data-testid=stream-delete-btn]');
     myRedisDbIcon = Selector('[data-testid=my-redis-db-icon]');
-    streamDeleteButton = Selector('[data-testid=stream-delete-btn]');
```
nmammadli, line 29: Why this selectors are deleted? We are using them on other tests

ofersteinberg (approved):

PR #1522, opened 2022-12-20
E2e/feature/import certificates

vlad-dargel:
tests for:
https://redislabs.atlassian.net/browse/RI-3728
https://redislabs.atlassian.net/browse/RI-3684
https://redislabs.atlassian.net/browse/RI-3801


tests/e2e/common-actions/databases-actions.ts
```diff
@@ -45,5 +54,6 @@ export type ImportDatabaseParameters = {
     userName?: string,
     password?: string,
     connectionType?: string,
-    fileName?: string
+    fileName?: string,
+    parsedJson?: any
```
ofersteinberg, line 58: Please add description

tests/e2e/test-data/certs/certsByPath/clientPath.crt
```diff
@@ -0,0 +1,3 @@
+-----BEGIN CERTIFICATE-----
+mockedClientCrtPath
+-----END CERTIFICATE-----
```
ofersteinberg, line 3: Couldn't you use one file for this and the two above?
Same for others which are the same

vlad-dargel: no, I'm testing here import of certificates by path(on machine), there are cases with the same certificates body and different certificates names and with different body and the same names

that's why we need multiple files here

tests/e2e/common-actions/databases-actions.ts
vlad-dargel: added, thanks!

nmammadli (approved):

PR #1561, opened 2023-01-05
#RI-3726 - Prepare for vizualizations for Graph and FT profile and explain commands

zalenskiSofteq:
#RI-3726 - Prepare for vizualizations for Graph and FT profile and explain commands


arthosofteq (approved):
lgtm

PR #1590, opened 2023-01-12
Feature/ri 3971 3572 add recommendations

AmirAllayarovSofteq:


zalenskiSofteq (approved):

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -411,6 +416,122 @@ export class RecommendationProvider {
...
+ * Check dangerous commands recommendation
+ * @param redisClient
+ */
+
+  async determineDangerousCommandsRecommendation(
+    redisClient: Redis | Cluster,
+  ): Promise<Recommendation> {
+    try {
+      const dangerousCommands = await redisClient.sendCommand(
+        new Command('command', ['LIST', 'FILTERBY', 'aclcat', 'dangerous'], { replyEncoding: 'utf8' }),
```
arthosofteq, line 512: This will not work for Redis < 7

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -411,6 +416,122 @@ export class RecommendationProvider {
...
+      ) as string[];
+
+      const filteredDangerousCommands = dangerousCommands.filter((command) => {
+        const commandName = command.split('|')[0];
+        return !redisInsightCommands.includes(commandName);
+      });
+
+      const activeDangerousCommands = await Promise.all(
+        filteredDangerousCommands.map(async (command) => await this.checkCommandInfo(redisClient, command)),
```
arthosofteq, line 521: What is the goal to additionally send `command info <cmd>` here?
Is it possible to use `command` without any args and filter all commands from that list?

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -424,6 +545,20 @@ export class RecommendationProvider {
     return false;
   }
 
+  private async checkCommandInfo(redisClient: Redis | Cluster, command: string): Promise<string> {
```
arthosofteq, line 548: I believe this method is not needed

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -411,6 +416,122 @@ export class RecommendationProvider {
...
+              isJSONOrHash = true;
+            }
+            processedKeysNumber += 1;
+            sortedSetNumber += 1;
+          }
+        }
+
+        return isJSONOrHash ? { name: RECOMMENDATION_NAMES.SEARCH_INDEXES } : null;
+      }
+      const sortedSets = keys
```
arthosofteq, line 479: Check types for standalone should be in separate method to decrease complexity.

arthosofteq: In general I'm fine with how you are processing search recommendations for standalone. nice!

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -411,6 +416,122 @@ export class RecommendationProvider {
...
+   * @param client
+   */
+  // eslint-disable-next-line
+  async determineSearchIndexesRecommendation(
+    redisClient: Redis,
+    keys: Key[],
+    client: any,
+  ): Promise<Recommendation> {
+    try {
+      if (client.isCluster) {
```
arthosofteq, line 432: Please move this to separate method. Let's have 2 separate method determineForCluster*... and determineForStandalone*...

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -411,6 +416,122 @@ export class RecommendationProvider {
...
+        let processedKeysNumber = 0;
+        let isJSONOrHash = false;
+        let sortedSetNumber = 0;
+        while (
```
arthosofteq, line 436: while? why? 
I think we should use here Promise.all to not wait before processing the next key.
In general I think that this implementation should be the same as for Standalone databases but with Promise.all -> sendCommand instead of piepeline

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -411,6 +416,122 @@ export class RecommendationProvider {
...
+            let keyType: string;
+            const sortedSetMember = await redisClient.sendCommand(
+              new Command('zrange', [keys[processedKeysNumber].name, 0, 0], { replyEncoding: 'utf8' }),
+            ) as string[];
+            try {
+              keyType = await redisClient.sendCommand(
+                new Command('type', [sortedSetMember[0]], { replyEncoding: 'utf8' }),
+              ) as string;
+            } catch (err) {
+              if (err && checkRedirectionError(err)) {
```
arthosofteq, line 453: Are you sure that we need this?
As I remember we can simplify this "redirection" functionality a lot.
Let's use not node client but `parent client` (general cluster client) to execute commands. In this case all commands will be routed to needed node automatically, no? did you check this behaviour?

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -411,6 +416,122 @@ export class RecommendationProvider {
...
+                ) as string;
+              }
+            }
+            if (keyType === RedisDataType.JSON || keyType === RedisDataType.Hash) {
+              isJSONOrHash = true;
+            }
+            processedKeysNumber += 1;
+            sortedSetNumber += 1;
+          }
+        }
```
arthosofteq, line 475: In general I would like to see 2 separate methods for cluster and standalone but with almost the same logic except pipeline vs Promise.all sendCommand (on cluster client). Is it possible?

redisinsight/api/src/modules/database-analysis/database-analysis.service.ts
```diff
@@ -55,19 +54,28 @@ export class DatabaseAnalysisService {
...
+        if (idx === 0) {
+          recommendationToExclude = concat(recommendationToExclude, ONE_NODE_RECOMMENDATIONS);
+        }
+        const foundedRecommendations = nodeRecommendations.filter((recommendation) => !isNull(recommendation));
+        const foundedRecommendationNames = foundedRecommendations.map(({ name }) => name);
+        recommendationToExclude = concat(recommendationToExclude, foundedRecommendationNames);
+        recommendationToExclude.push(...foundedRecommendationNames);
+        jobsArray.push(foundedRecommendations);
+        return flatten(jobsArray);
+      }, Promise.resolve([]));
```
arthosofteq, line 77: In general I understand this idea. It is not readable a little bit but should work. I would prefer to have Map with recommendations when each recommendation has own type and based on it we will check this recommendation or not. But let's revise this approach and refactor current solution when we will add batch or scan more functionality

PR #1675, opened 2023-02-01
E2e/feature/ri 4069 client list

nmammadli:


tests/e2e/common-actions/workbench-actions.ts
```diff
@@ -0,0 +1,16 @@
+import {t} from 'testcafe';
+import { WorkbenchPage } from '../pageObjects';
+
+const workbenchPage = new WorkbenchPage();
+
+export class WorkbenchActions {
+    /*
+        Verify after client list command columns are visible
+    */
+    async verifyClientListColumnsAreVisible(): Promise<void> {
```
vlad-dargel, line 10: I think it is better to test it with loop and some dynamic selector.
e.g.
const columns = ['id', 'addr', 'name', 'user'];

for (const column of columns) {
let columnSelector = Selector(`[data-test-subj^=tableHeaderCell_${column}_]`);
await t.expect(columnSelector.visible).ok(`${column} column is not visible`);
}

does it make sense?

vlad-dargel: so you will be able to test any number of columns described in "columns" array

tests/e2e/tests/regression/workbench/command-results.e2e.ts
```diff
@@ -117,3 +118,20 @@ test('Big output in workbench is visible in virtualized table', async t => {
     // Verify that all commands scrolled
     await t.expect(lastExpectedItem.visible).ok('Final execution time message not displayed');
 });
+
+test.before(async t => {
+    await acceptLicenseTermsAndAddDatabaseApi(ossStandaloneRedisearch, ossStandaloneRedisearch.databaseName);
+    // Add index and data
```
vlad-dargel, line 124: wrong comment

tests/e2e/tests/regression/workbench/command-results.e2e.ts
```diff
@@ -117,3 +118,20 @@ test('Big output in workbench is visible in virtualized table', async t => {
...
+    await t.click(myRedisDatabasePage.workbenchButton);
+}).after(async t => {
+    await t.switchToMainWindow();
+    await deleteStandaloneDatabaseApi(ossStandaloneRedisearch);
+})('verify client list plugin shows table', async t => {
```
vlad-dargel, line 129: I don't see this test name in checklist
you need to add it to Scenario column

tests/e2e/tests/regression/workbench/command-results.e2e.ts
```diff
@@ -117,3 +118,20 @@ test('Big output in workbench is visible in virtualized table', async t => {
...
+    const command = 'CLIENT LIST';
+    // Send command in workbench to view client list
+    await workbenchPage.sendCommandInWorkbench(command);
+    await t.expect(workbenchPage.typeSelectedClientsList.visible).ok('client list view button is not visible');
+    await t.switchToIframe(workbenchPage.iframe);
+    // Verify that I can see the client List visualization available for all users.
+    await workBenchActions.verifyClientListColumnsAreVisible();
+});
```
vlad-dargel, line 137: you can automate this scenarios as well:
"Verify that user can see the Client List visualization available for all users."
by checking the number of rows in table and comparing with the number of rows in text view
and also this scenario can be automated
Verify that user have the following error when there is no permission to run the CLIENT LIST: "NOPERM this user has no permissions to run the 'CLIENT LIST' command or its subcommand"

vlad-dargel (approved):

PR #1680, opened 2023-02-03
#RI-3999 - Automatically generate SHA256 for release builds

zalenskiSofteq:


arthosofteq (dismissed):
more better. even without python scripts

arthosofteq (approved):

PR #1723, opened 2023-02-10
Add onbarding steps tests

nmammadli:


vlad-dargel (commented):
Great job!

just please check the names of the tests and make them the same as in the checklist
ideally, test names should be like in the 'Scenario' column, and checklist checks can be as comments inside the tests

vlad-dargel (approved):

PR #1733, opened 2023-02-13
update app version to 2.20.0

vlad-dargel:
Updates for appVersion (should be merged before creating release build  2.20.0)


arthosofteq (approved):
I think that we should get it from package.json but this approach fine with me for now

PR #1738, opened 2023-02-14
add test for https://redislabs.atlassian.net/browse/RI-3995

nmammadli:


tests/e2e/common-actions/browser-actions.ts
```diff
@@ -40,4 +39,17 @@ export class BrowserActions {
             ? await t.expect(browserPage.tooltip.textContent).contains(expectedText, `"${expectedText}" Text is incorrect in tooltip`)
             : await t.expect(browserPage.tooltip.textContent).notContains(expectedText, `Tooltip still contains text "${expectedText}"`);
     }
+    /**
+     * Verify that the new key is displayed at the top of the list of keys and opened and pre-selected in List view
+     * */
+    async verifyKeyDisplayedTopAndOpened(keyName: string): Promise<void> {
+        await t.expect(Selector('[aria-rowindex="1"]').withText(keyName).visible).ok(`element with ${keyName} is not visible in the top of list`);
+        await t.expect(Selector('[data-testid="key-name-text"]').withText(keyName).visible).ok(`element with ${keyName} is not opened`);
```
vlad-dargel, line 47: instead of Selector('[data-testid="key-name-text"]')
use browserPage.keyNameFormDetails

tests/e2e/common-actions/browser-actions.ts
```diff
@@ -40,4 +39,17 @@ export class BrowserActions {
...
+    }
+    /**
+     * Verify that the new key is not displayed at the top of the list of keys and opened and pre-selected in List view
+     * */
+    async verifyKeyIsNotDisplayedTop(keyName: string): Promise<void> {
+        await t.expect(Selector('[aria-rowindex="1"]').withText(keyName).visible).notOk(`element with ${keyName} is not visible in the top of list`);
```
vlad-dargel, line 53: in testCafe tests with .visible + notOK waits for 5 seconds by default - and it increases test run time
use .exists here instead of .visible

tests/e2e/tests/regression/browser/add-keys.e2e.ts
```diff
@@ -44,3 +53,55 @@ test('Verify that user can create different types(string, number, null, array, b
         }
     }
 });
+// https://redislabs.atlassian.net/browse/RI-3995
+test
+    .before(async() => {
+        await acceptLicenseTermsAndAddDatabaseApi(ossStandaloneConfig, ossStandaloneConfig.databaseName);
```
vlad-dargel, line 59: you need to user ossStandaloneBigConfig here
because the main point of this feature is to display the new key at the top in huge databases where previously keys were added somewhere at the middle

vlad-dargel: and seems like with big database you will be able to add test 'Verify that user can see the key removed from the top when refresh List view'

vlad-dargel (commented):
Good job, just need some updates.

And please think about adding some additional tests for Tree view:
1) When some folder with keys by delimiter selected and user adds a new key without delimiter - key should be displayed in "Keys" root folder
2) When some folder with keys by delimiter selected and user adds a new key with delimiter from other folder - key should be displayed in other folder and this folder should be selected
2) When some folder with keys by delimiter selected and user adds a new key with delimiter from the same folder  - key should be displayed in the same folder

these cases must necessarily be covered, because this is a rather vulnerable functionality that often breaks

looks like you can add these tests inside the already existing 'Tree view navigations improvement tests'

vlad-dargel (approved):

PR #1758, opened 2023-02-20
E2e/feature/ri 4070 onboarding wb ft index

nmammadli:


tests/e2e/tests/regression/browser/onboarding.e2e.ts
```diff
@@ -29,9 +39,16 @@ fixture `Onboarding new user tests`
...
     })
     .afterEach(async() => {
+        await cliPage.sendCommandInCli(`DEL ${indexName}`);
         await deleteDatabase(ossStandaloneConfig.databaseName);
     });
-test
+// https://redislabs.atlassian.net/browse/RI-4070, https://redislabs.atlassian.net/browse/RI-4067
+test.before(async() => {
+    await acceptTermsAddDatabaseOrConnectToRedisStack(ossStandaloneConfig, ossStandaloneConfig.databaseName);
+    await cliPage.sendCommandInCli('flushdb'); // to delete all indexes in order not to fail FT.INFO ${indexName}
```
vlad-dargel, line 48: flushdb deletes not only indexes but all keys in db as well
if we add parallelization in future it wil break all tests

to delete all indexes you can use
FT._LIST + FT.DROPINDEX

vlad-dargel: or you can ask Artem to create and add docker container with a new database without keys and indexes (but with plugins) for such tests

tests/e2e/tests/regression/browser/onboarding.e2e.ts
```diff
@@ -88,6 +112,7 @@ test
         // verify onboarding step completed successfully
         await onBoardActions.verifyOnboardingCompleted();
     });
+// https://redislabs.atlassian.net/browse/RI-4070
```
vlad-dargel, line 115: wrong link here
should be https://redislabs.atlassian.net/browse/RI-4067

tests/e2e/helpers/conf.ts
```diff
@@ -19,6 +19,14 @@ export const ossStandaloneConfig = {
     databasePassword: process.env.OSS_STANDALONE_PASSWORD
 };
 
+export const ossStandaloneConfig2 = {
+    host: process.env.OSS_STANDALONE_HOST || 'oss-standalone-2',
+    port: process.env.OSS_STANDALONE_PORT || '6379',
+    databaseName: `${process.env.OSS_STANDALONE_DATABASE_NAME || 'test_standalone'}-${uniqueId}`,
```
vlad-dargel, line 25: the same name as for ossStandaloneConfig
please rename to 'test_standalone_empty'}-${uniqueId}`

tests/e2e/helpers/conf.ts
```diff
@@ -19,6 +19,14 @@ export const ossStandaloneConfig = {
     databasePassword: process.env.OSS_STANDALONE_PASSWORD
 };
 
+export const ossStandaloneConfig2 = {
```
vlad-dargel, line 22: It would be better to name it as ossStandaloneConfigEmpty or something like that

vlad-dargel (approved):

PR #1781, opened 2023-02-26
[RI-3726] Profile Plugin - Add support for leftoff RediSearch expressions

GnaneshKunal:
- Remove edge target marker
- Add support for LEXRANGE, IDS, NOT, OPTION, EXACT, FUZZY, WILDCARD, and PREFIX expressions (All of the expressions' parsing is based on RediSearch's latest source code) - https://github.com/RediSearch/RediSearch/blob/master/src/query.c
- Parse Vector Expressions
- Parse tag identifiers with numbers
- Properly skip backslash if exsits (multiple backslash appears)
- Parse identifiers with pipe character
- Make TAG expressions be parsed via expand expression parser since it might contain children
- Remove support for FT.PROFILE on cluster
- Add max-cap zoom

### DIFF
--- a/redisinsight/ui/src/packages/ri-explain/src/Explain.tsx
+++ b/redisinsight/ui/src/packages/ri-explain/src/Explain.tsx
@@ -82,6 +82,7 @@ export default function Explain(props: IExplain): JSX.Element {
       let [cluster, entityInfo] = ParseProfileCluster(info)
       cluster['Coordinator'].forEach((kv: [string, string]) => profilingTime[kv[0]] = kv[1])
       data = entityInfo
+      return <div className="responseFail">Visualization of FT.PROFILE on cluster is not yet supported.</div>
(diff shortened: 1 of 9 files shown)


redisinsight/ui/src/packages/ri-explain/src/Explain.tsx
```diff
@@ -72,6 +73,18 @@ export default function Explain(props: IExplain): JSX.Element {
 
   const module = ModuleType.Search
 
+  const [parsedRedisReply, setParsedRedisReply] = useState('')
+
+  useEffect(() => {
+    if (command == 'ft.profile') {
```
romansergeenkosofteq, line 79: should be `command?.toLowerCase()`?

GnaneshKunal: the command is lowercased in line 51.

romansergeenkosofteq (approved):

PR #1847, opened 2023-03-21
#RI-4075 - add database recommendation module

AmirAllayarovSofteq:


redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
```diff
@@ -0,0 +1,79 @@
...
+    @InjectRepository(DatabaseRecommendationEntity)
+    private readonly repository: Repository<DatabaseRecommendationEntity>,
+  ) {}
+
+  /**
+   * Save entire entity
+   * @param databaseId
+   * @param recommendationName
+   */
+  async create(databaseId: string, recommendationName: string): Promise<RecommendationDto> {
```
arthosofteq, line 24: what we return here? - Promise<RecommendationDto>. (should be Recommendation)

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
```diff
@@ -0,0 +1,79 @@
...
+    this.logger.log('Creating database recommendation');
+    return await this.repository.save(
```
arthosofteq, line 26: In fact DatabaseRecommendationEntity will be returned. Strange...

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * Return list of database recommendations
+   * @param clientMetadata
+   */
+  async list(clientMetadata: ClientMetadata): Promise<RecommendationsDto> {
+    this.logger.log('Getting database recommendations list');
+    const recommendations = await this.repository
+      .createQueryBuilder('r')
+      .where({ databaseId: clientMetadata.databaseId })
+      .select(['r.id', 'r.name'])
```
arthosofteq, line 40: I strongly believe that you should select r.read here to be able highlight unread (new) recommendations on UI

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
arthosofteq: Also usually there is no need for await when you return Promise in the async function

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * Return list of database recommendations
+   * @param clientMetadata
+   */
+  async list(clientMetadata: ClientMetadata): Promise<RecommendationsDto> {
+    this.logger.log('Getting database recommendations list');
+    const recommendations = await this.repository
+      .createQueryBuilder('r')
+      .where({ databaseId: clientMetadata.databaseId })
```
arthosofteq, line 39: I'm not sure about this filter. I forgot requirements but please double check if we need to return disabled recommendations, for example

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
```diff
@@ -0,0 +1,79 @@
...
+      .where({ databaseId })
+      .set({ read: true })
+      .execute();
+
+    const totalUnread = await this.repository
+      .createQueryBuilder()
+      .where({ databaseId, read: false })
+      .getCount();
+
+    return new RecommendationsDto({
```
arthosofteq, line 74: not sure about response schema. Seems like will not match to **list** response if I'll call it right after **read**. What should happen after read all recommendations?

redisinsight/api/src/modules/database-recommendation/scanner/recommendation.provider.ts
```diff
@@ -0,0 +1,31 @@
...
+@Injectable()
+export class RecommendationProvider {
+  private strategies: Map<string, IRecommendationStrategy> = new Map();
+
+  constructor(
+    databaseService: DatabaseService,
+  ) {
+    this.strategies.set('default', new DefaultRecommendationStrategy());
+    this.strategies.set(RECOMMENDATION_NAMES.SEARCH_STRING, new RediSearchStrategy(databaseService));
+  }
```
arthosofteq, line 20: What is RecommendationProvider and RecommendationStrategy?
if each recommendation type will have own strategy to determine if we must create recommendation - I'm totally fine but I don't understand what is default strategy here? 
We should throw an error (or silently fail) if there is no such recommendation type, no?

arthosofteq: Also why **default** is outside of RECOMMENDATION_NAMES?

redisinsight/api/src/modules/database-recommendation/scanner/strategies/abstract.recommendation.strategy.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { DefaultRecommendationStrategy } from 'src/modules/database-recommendation/scanner/strategies';
+
+describe('AbstractRecommendationStrategy', () => {
+  const strategy = new DefaultRecommendationStrategy();
+
+  describe('isRecommendationReached', () => {
+    it('should get is recommendation reached', async () => {
+      expect(await strategy.isRecommendationReached()).toEqual(false);
```
arthosofteq, line 8: Well, now I see that default strategy should always return false. Will work. Anyway I prefer to have some log or telemetry in the future to understand that we have missed something

redisinsight/api/src/modules/database/middleware/connection.middleware.ts
```diff
@@ -47,7 +46,7 @@ export class ConnectionMiddleware implements NestMiddleware {
     return {
       timeout: req.body?.timeout,
       instanceIdFromReq: req.params?.id,
-     };
+    };
   }
```
arthosofteq, line 51: In general I'm fine with such changes. You found bad code style and fixed it. But I think it should be in a different pr or in a smaller one. Anyway, thanks!

arthosofteq (changes requested):
@AmirAllayarovSofteq well done. 
In general only few comments must be resolved before merge. 
However sometimes it is better to ask before implementation, because I feel like we must rework this to match to other modules.

redisinsight/api/src/modules/database/middleware/connection.middleware.ts
AmirAllayarovSofteq: Got it. I did only after conversation with @romansergeenkosofteq (He told that it is not a problem to refactor other parts in PR)

redisinsight/api/src/modules/database-recommendation/dto/create-recommendation.dto.ts
```diff
@@ -0,0 +1,17 @@
...
+} from 'class-validator';
+
+export class CreateRecommendationDto {
+  @IsNotEmpty()
+  @IsString()
+  name: string;
+
+  @IsOptional()
+  @IsString()
+  disabled?: boolean;
```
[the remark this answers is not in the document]
AmirAllayarovSofteq: It is field for future, we agreed that we will have ability for user to disable recommendation and do not determined it at all.

redisinsight/api/src/modules/database-recommendation/entities/database-recommendation.entity.ts
```diff
@@ -0,0 +1,47 @@
...
+    {
+      nullable: false,
+      onDelete: 'CASCADE',
+    },
+  )
+  @JoinColumn({ name: 'databaseId' })
+  database: DatabaseEntity;
+
+  @Expose()
+  @Column({ nullable: false, default: 'luaScript' })
```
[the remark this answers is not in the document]
AmirAllayarovSofteq: my fault, removed

redisinsight/api/src/modules/database-recommendation/dto/create-recommendation.dto.ts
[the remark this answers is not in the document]
arthosofteq: During creation? I think it should be done in a different endpoint with a different dto, no?

redisinsight/api/src/modules/database/middleware/connection.middleware.ts
AmirAllayarovSofteq: BTW i will create another PR next time

redisinsight/api/src/modules/database-recommendation/dto/create-recommendation.dto.ts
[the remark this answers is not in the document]
AmirAllayarovSofteq: Agree

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
AmirAllayarovSofteq: Right now in mockup we don't have any recommendation highlighting. Just have we new recommendations or not(field totalUnread controlling it).
BTW we have field "read" in each recommendation, so we can highlight them

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+      repository.createQueryBuilder().getCount.mockResolvedValueOnce(1);
+      expect(await service.list(mockClientMetadata)).toEqual({
+        recommendations: [{
+          name: mockDatabaseRecommendation.name,
+        }],
+        totalUnread: 1,
+      });
+    });
+  });
+});
```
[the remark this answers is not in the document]
AmirAllayarovSofteq: add one more test here. Could you remain me what was previous test coverage? (i tried to control test coverage and up it to green values, but i will double check)

redisinsight/api/src/modules/database-recommendation/scanner/strategies/abstract.recommendation.strategy.spec.ts
AmirAllayarovSofteq: Got it. I will think how better cover it by logger. Thanks

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
AmirAllayarovSofteq: Got it

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
arthosofteq: But if we don't need to return notifications and totalUnread will be always 0 do we need to return anything then?

redisinsight/api/src/modules/database-recommendation/models/recommendation.ts
```diff
@@ -0,0 +1,35 @@
+import { Expose } from 'class-transformer';
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+
+export class Recommendation {
```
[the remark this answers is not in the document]
AmirAllayarovSofteq: deprecated. It was updated

redisinsight/api/src/modules/database-recommendation/database-recommendation.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+      {
+        status: 200,
+        type: Recommendation,
+      },
+    ],
+  })
+  @Get('')
+  async list(
+    @BrowserClientMetadata() clientMetadata: ClientMetadata,
+  ): Promise<RecommendationsDto> {
```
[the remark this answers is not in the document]
AmirAllayarovSofteq: Got it. Renamed

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
AmirAllayarovSofteq: I'm not sure that caught you. Here we return Promise<DatabaseRecommendation>

redisinsight/api/src/modules/database-recommendation/database-recommendation.service.ts
```diff
@@ -0,0 +1,70 @@
...
+   * @param clientMetadata
+   * @param recommendationName
+   * @param data
+   */
+  public async check(
+    clientMetadata: ClientMetadata,
+    recommendationName: string,
+    data: any,
+  ): Promise<RecommendationDto> {
+    const result = await this.databaseRecommendationsProvider.list(clientMetadata);
```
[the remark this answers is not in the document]
AmirAllayarovSofteq: Could you help me here? I'm not sure that i got how to do it

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
arthosofteq: Without read field - there is no way to distinguish between old (already seen by user) and new recommendations. 
So if we will think in advance we can predict that this field will be required.
Anyway if there is no such requirement I'm totally fine with current implementation.
P.S. I assume that the same thing with createdAt field

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
arthosofteq: this.repository.save - returns **Promise\<DatabaseRecommendationEntity\>** when you claim that your function should return **Promise\<RecommendationDto\>**

redisinsight/api/src/modules/database-recommendation/entities/database-recommendation.entity.ts
```diff
@@ -0,0 +1,47 @@
...
+  name: string;
+
+  @Column({ nullable: false, default: false })
```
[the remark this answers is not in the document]
arthosofteq: To be honest there is not a big difference. You can do it in any way. Just thought about additional conditions. Nevermind

redisinsight/api/src/modules/database-recommendation/models/redisearch.ts
```diff
@@ -0,0 +1,27 @@
+import { Redis, Cluster } from 'ioredis';
+import { Expose } from 'class-transformer';
+import { ApiProperty } from '@nestjs/swagger';
+import { GetKeyInfoResponse } from 'src/modules/browser/dto';
+
+export class Redisearch {
```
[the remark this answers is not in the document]
arthosofteq: Ok. To be honest I didn't find where you use it in code. Let's think on it together

redisinsight/api/src/modules/database-recommendation/database-recommendation.service.ts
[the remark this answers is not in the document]
AmirAllayarovSofteq: Done. Thank you for help

redisinsight/api/src/modules/database-recommendation/models/redisearch.ts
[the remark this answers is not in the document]
AmirAllayarovSofteq: Got solution, will improve in next PR

arthosofteq (approved):
Looks good.
We have to migrate to dynamic module and repository approach in the future and in the scope of that task let's rework recommendation data models and mechanism to process them
Thanks!

PR #1861, opened 2023-03-23
#RI-4304 - run ft config per database

zalenskiSofteq:
#RI-4304 - run ft config per database


redisinsight/api/src/modules/browser/services/redisearch/redisearch.service.ts
```diff
@@ -23,10 +23,11 @@ import { BrowserHistoryMode } from 'src/common/constants';
 import { BrowserToolService } from '../browser-tool/browser-tool.service';
 import { BrowserHistoryService } from '../browser-history/browser-history.service';
 import { CreateBrowserHistoryDto } from '../../dto/browser-history/create.browser-history.dto';
+import { IMaxSearchResults } from '../../interfaces';
 
 @Injectable()
 export class RedisearchService {
-  private maxSearchResults: number | null = null;
+  private maxSearchResults: IMaxSearchResults = {};
```
arthosofteq, line 30: To be honest I prefer to have Map here but its up to you

arthosofteq (approved):
nice

redisinsight/api/src/modules/browser/services/redisearch/redisearch.service.ts
zalenskiSofteq: good catch, changed to Map

arthosofteq (approved):

PR #1922, opened 2023-04-06
e2e/feature/RI-4084 upvote extended recommendations

vlad-dargel:


tests/e2e/common-actions/recommendations-actions.ts
```diff
@@ -0,0 +1,44 @@
...
+
+        await t.click(voteSelector);
+    }
+
+    /**
+     * Verify that vote is disabled by recommendation name and option
+     * @param recommendationName Name of recommendation
+     * @param option Option can be "very-useful/useful/not-useful"
+    */
+    async verifyVoteDisabled(recommendationName: string, option: string): Promise<void> {
```
mariasergeenko, line 39: will it work if we do verification in tests?

mariasergeenko (approved):

tests/e2e/common-actions/recommendations-actions.ts
vlad-dargel: it is in common-actions - special directory for actions with expects

PR #1924, opened 2023-04-06
add 2 tests related to delete of Key form List view

mariasergeenko:


tests/e2e/tests/critical-path/browser/keylist-actions.e2e.ts
```diff
@@ -0,0 +1,38 @@
...
+} from '../../../helpers/conf';
+import { rte } from '../../../helpers/constants';
+import { deleteStandaloneDatabaseApi } from '../../../helpers/api/api-database';
+import { Common } from '../../../helpers/common';
+
+const browserPage = new BrowserPage();
+const common = new Common();
+
+const keyName = common.generateWord(10);
+fixture `Actions with Key List on Browser page`
```
vlad-dargel, line 15: please add empty line between const and fixture
for better readability

tests/e2e/tests/critical-path/browser/keylist-actions.e2e.ts
```diff
@@ -0,0 +1,38 @@
...
+        await deleteStandaloneDatabaseApi(ossStandaloneConfig);
+    });
+test('Verify that user can delete key in List mode', async t => {
+    // Add new key
+    await browserPage.addStringKey(keyName);
+    await browserPage.deleteKeyByNameFromList(keyName);
+    await t.expect(await browserPage.isKeyIsDisplayedInTheList(keyName)).notOk('The Key wasn\'t deleted');
+});
+
+test.only('Verify that user can delete key in Tree view', async t => {
```
vlad-dargel, line 32: .only

tests/e2e/tests/critical-path/browser/keylist-actions.e2e.ts
```diff
@@ -0,0 +1,38 @@
...
+    .page(commonUrl)
+    .beforeEach(async() => {
+        await acceptLicenseTermsAndAddDatabaseApi(ossStandaloneConfig, ossStandaloneConfig.databaseName);
+    })
+    .afterEach(async() => {
+        // Delete database
+        await deleteStandaloneDatabaseApi(ossStandaloneConfig);
+    });
+test('Verify that user can delete key in List mode', async t => {
+    // Add new key
```
vlad-dargel, line 26: its better to set keyName for each test differently
keyName = common.generateWord(10);

in case if one test will fail on deletion - the second will fail too, because it won't be able to create a key with the same name

vlad-dargel (commented):
Can you please add these test names into checklist?
- Verify that user can delete key in List mode
- Verify that user can delete key in Tree view

and if deletion works also in search capabilities page (redisSearch), we need a test for it
- Verify that user can delete key in Search capabilities

vlad-dargel (approved):

PR #1930, opened 2023-04-09
#RI-4380 BE Upload data in bulk base implementation

arthosofteq:


redisinsight/api/src/modules/bulk-actions/dto/upload-import-file.dto.ts
```diff
@@ -0,0 +1,17 @@
...
+import { IsNotEmpty } from 'class-validator';
+import {
+  IsFile, MaxFileSize, MemoryStoredFile,
+} from 'nestjs-form-data';
+
+export class UploadImportFileDto {
+  @ApiPropertyOptional({
+    type: 'string',
+    format: 'binary',
+    description: 'Import file (with list of commands to execute',
```
AmirAllayarovSofteq, line 11: I guess you lost close round bracket.

redisinsight/api/src/modules/bulk-actions/bulk-import.service.ts
```diff
@@ -0,0 +1,119 @@
...
+      summary: {
+        processed: 0,
+        succeed: 0,
+        failed: 0,
+        errors: [],
+      },
+      progress: null,
+      filter: null,
+      status: BulkActionStatus.Completed,
+      duration: Date.now(),
```
AmirAllayarovSofteq, line 72: Maybe will be better to create another constant startBulkActionTime or something else, it will be more clear i guess. And in the end result.duration = Date.now() - startBulkActionTime. But it is up to you.

AmirAllayarovSofteq (approved):
Looks good.

redisinsight/api/src/modules/bulk-actions/dto/upload-import-file.dto.ts
arthosofteq: thanks

redisinsight/api/src/modules/bulk-actions/bulk-import.service.ts
arthosofteq: sure. if this will be more easy to understand

PR #1933, opened 2023-04-10
#RI-4086 - [FE+BE] Recommend Workbench visualizations in CLI

zalenskiSofteq:
#RI-4086 - [FE+BE] Recommend Workbench visualizations in CLI


arthosofteq (approved):
lgtm

PR #1942, opened 2023-04-11
added missed rules for transformation from ASCII string to Buffer

arthosofteq:


arthosofteq:
@zalenskiSofteq This should fix incorrect behavior for json plugin

zalenskiSofteq (approved):

PR #1967, opened 2023-04-18
#RI-4370 - [BE] Snooze recommendations

zalenskiSofteq:
#RI-4370 - [BE] Snooze recommendations

### DIFF
--- a/redisinsight/api/src/modules/database-recommendation/database-recommendation.controller.ts
+++ b/redisinsight/api/src/modules/database-recommendation/database-recommendation.controller.ts
@@ -9,7 +9,11 @@ import { ClientMetadata } from 'src/common/models';
 import {
   DatabaseRecommendationsResponse,
 } from 'src/modules/database-recommendation/dto/database-recommendations.response';
-import { ModifyDatabaseRecommendationDto } from './dto';
+import {
+  ModifyDatabaseRecommendationDto,
+  DeleteDatabaseRecommendationDto,
+  DeleteDatabaseRecommendationResponse,
+ } from './dto';
 
 @ApiTags('Database Recommendations')
 @Controller('/recommendations')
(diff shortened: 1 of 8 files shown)


AmirAllayarovSofteq (approved):
LGTM

redisinsight/api/src/modules/database-recommendation/database-recommendation.service.ts
```diff
@@ -88,4 +89,35 @@ export class DatabaseRecommendationService {
...
+   * Bulk delete recommendations. Uses "delete" method and skipping error
+   * Returns successfully deleted recommendations number
+   * @param clientMetadata
+   * @param ids
+   */
+  async bulkDelete(clientMetadata: ClientMetadata, ids: string[]): Promise<DeleteDatabaseRecommendationResponse> {
+    this.logger.log(`Deleting many recommendations: ${ids}`);
+
+    return {
+      affected: sum(await Promise.all(ids.map(async (id) => {
```
arthosofteq, line 113: It is not good practice to delete database entries with multiple queries when it is possible to do with a single one.
We should not have an issue here we will not have a lot of recommendations but keep in mind next time

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
```diff
@@ -127,17 +127,38 @@ export class DatabaseRecommendationProvider {
...
+        this.logger.error(`Recommendation with id:${id} was not Found`);
+        throw new NotFoundException(ERROR_MESSAGES.DATABASE_RECOMMENDATION_NOT_FOUND);
       }
 
-      this.logger.log(`Succeed to get recommendation with id: ${id}'`);
-      return model;
+      this.logger.log('Succeed to delete recommendation.');
+    } catch (error) {
+      this.logger.error(`Failed to delete recommendation: ${id}`, error);
+      throw new InternalServerErrorException();
```
arthosofteq, line 161: What about error message?

redisinsight/api/src/modules/database-recommendation/database-recommendation.controller.ts
```diff
@@ -76,4 +80,30 @@ export class DatabaseRecommendationController {
...
+  })
+  @UsePipes(
+    new ValidationPipe({
+      transform: true,
+      whitelist: true,
+      forbidNonWhitelisted: true,
+    }),
+  )
+  async bulkDeleteDatabaseRecommendation(
+    @BrowserClientMetadata() clientMetadata: ClientMetadata,
```
arthosofteq, line 104: `BrowserClientMetadata` related to browser endpoints, we should use own decorator here. Anyway since I missed this even in previous PRs and currently these endpoint database param matches to browser we can go with it

arthosofteq (approved):

redisinsight/api/src/modules/database-recommendation/database-recommendation.service.ts
zalenskiSofteq: I'll keep keep in mind

redisinsight/api/src/modules/database-recommendation/providers/database-recommendation.provider.ts
zalenskiSofteq: added `error.message`

PR #1975, opened 2023-04-19
#RI-4346 - Rename Display on System Tray on mac

zalenskiSofteq:
#RI-4346 - Rename Display on System Tray on mac


romansergeenkosofteq (approved):
Good job!

PR #1984, opened 2023-04-20
E2e/feature/ri 4075 livetime recommendations

mariasergeenko:


tests/e2e/tests/regression/insights/live-recommendations.e2e.ts
```diff
@@ -150,4 +150,16 @@ test('Verify that user can hide recommendations and checkbox value is saved', as
...
+    const recommendation = RecommendationIds.searchVisualization;
+    const commandToGetRecommendation = 'FT.INFO';
+    await browserPage.Cli.sendCommandInCli(commandToGetRecommendation);
+
+    await insightsPage.toggleInsightsPanel(true);
+    await insightsPage.snoozeRecommendation(recommendation);
+
+    await common.reloadPage();
+    await insightsPage.toggleInsightsPanel(true);
+    await t.expect(await insightsPage.isRecommendationExists(recommendation)).notOk('recommendation is displayed when after snoozing');
```
vlad-dargel, line 163: I think that it would be better to add here another check that you can call this recommendation again and it will be displayed

vlad-dargel (approved):

PR #2057, opened 2023-05-03
E2e/feature/ri 4352 bulk upload from tutorials

vlad-dargel:


tests/e2e/helpers/common.ts
```diff
@@ -188,4 +191,34 @@ export class Common {
...
+
+        // Add the contents of the directory to the zip archive
+        archive.directory(sourceDir, false);
+
+        // Finalize the archive and write it to disk
+        archive.finalize();
+        await new Promise((resolve) => {
+            output.on('close', resolve);
+            archive.pipe(output);
+        });
```
mariasergeenko, line 214: could you please clarify why the promise is need here? it looks like it won't work as async code

vlad-dargel: thanks, looks like it can be simplified
fixed

mariasergeenko (approved):

PR #2125, opened 2023-05-23
E2e/feature/ri 4399 hide recommendations for users

vlad-dargel:


tests/e2e/tests/regression/insights/feature-flag.e2e.ts
```diff
@@ -0,0 +1,150 @@
...
+    await updateControlNumber(19.2);
+
+    const featureVersion = await JSON.parse(await getColumnValueFromTableInDB(featuresConfigTable, 'data')).version;
+
+    await t.expect(featureVersion).eql(1, 'Config highest version not applied');
+    await t.expect(browserPage.InsightsPanel.insightsBtn.exists).notOk('Insights panel displayed when disabled in default config');
+});
+test
+    .before(async() => {
+        await acceptLicenseTermsAndAddDatabaseApi(ossStandaloneConfig, ossStandaloneConfig.databaseName);
```
mariasergeenko, line 63: why do we call acceptLicenseTermsAndAddDatabaseApi, if we call it in beforeEach?

vlad-dargel: no, beforeEach applies only if test doesn't have before hook
in this case only before hook applied

mariasergeenko (approved):

PR #2174, opened 2023-06-08
#RI-4265 - Enhance the window opening in Electron

zalenskiSofteq:
#RI-4265 - Enhance the window opening in Electron


redisinsight/electron/index.html
```diff
@@ -19,16 +19,16 @@
 <body class="electron">
   <div id="root" style="height: 100%; width: 100%;"></div>
   <script>
-    if (typeof process === 'object') {
+    if (typeof envVars === 'object') {
```
arthosofteq, line 22: what is envVars? just an alias to process.env?

redisinsight/electron/index.html
```diff
@@ -19,16 +19,16 @@
...
 
-      if (process.env.NODE_ENV === 'development') {
+      if (envVars.NODE_ENV === 'development') {
         // Dynamically insert the DLL script in development env in the
         // renderer process
         scripts.push('../dll/renderer.dev.dll.js');
 
         // Dynamically insert the bundled app script in the renderer process
-        const port = process.env.PORT || 1212;
+        const port = envVars.PORT || 1212;
```
arthosofteq, line 31: again if we decide to use envVars - let's implement it as config on BE.  when you have all default values inside it and here just get the value

redisinsight/electron/main.dev.ts
```diff
@@ -288,6 +292,11 @@ export const createWindow = async (splash: BrowserWindow | null = null) => {
     shell.openExternal(url);
   });
 
+  newWindow.webContents.setWindowOpenHandler((edata: any) => {
```
arthosofteq, line 295: is it to deprecate opening external resources inside the window? if so - nice

redisinsight/electron/preload.ts
```diff
@@ -0,0 +1,26 @@
+// Disable no-unused-vars, broken for spread args
+/* eslint no-unused-vars: off */
+import { contextBridge, ipcRenderer } from 'electron';
+import { IpcEvent } from 'uiSrc/electron/constants';
+
+const electronHandler = {
+  ipcRenderer: {
+    invoke: (channel: IpcEvent, data?: any) => {
+      // whitelist channels
+      if (Object.values(IpcEvent).includes(channel)) {
```
arthosofteq, line 10: I like an idea to have whitelist. Nice!

redisinsight/electron/preload.ts
```diff
@@ -0,0 +1,26 @@
...
+      }
+
+      return new Error('channel is not allowed');
+    },
+  },
+};
+
+contextBridge.exposeInMainWorld('electron', electronHandler);
+
+contextBridge.exposeInMainWorld('ENV_VARS', {
```
arthosofteq, line 21: I don't like to export env variables. Let's have electron.config object with everything needed. What do you think?

redisinsight/electron/util.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint import/prefer-default-export: off */
+import { URL } from 'url';
+import path from 'path';
+
+export function resolveHtmlPath(htmlFileName: string) {
+  if (process.env.NODE_ENV === 'development') {
```
arthosofteq, line 6: I really do not understand why somewhere we are using envVars somewhere process.env and what was the goal

redisinsight/electron/util.ts
```diff
@@ -0,0 +1,13 @@
...
+    const port = process.env.PORT || 1212;
+    const url = new URL(`http://localhost:${port}`);
+    url.pathname = htmlFileName;
+    return url.href;
+  }
+  return `file://${path.resolve(__dirname, '../renderer/', htmlFileName)}`;
```
arthosofteq, line 12: ok. as I understand this function is to load templates (html pages) to our app
Why we have http protocol hardcoded? what will happen if we'll use tls?

redisinsight/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "version": "2.26.0",
   "description": "RedisInsight",
-  "main": "./main.prod.js",
+  "main": "./dist/main/main.js",
```
arthosofteq, line 7: is it typo? `main/main.js`? may be single main.js?

arthosofteq (changes requested):
I dind't look to ui/* part
In general looks nice but have few questions to address before approve

arthosofteq (approved):
Approve with agreement that not resolved comments will be addressed in another PR https://github.com/RedisInsight/RedisInsight/pull/2182

PR #2199, opened 2023-06-15
fixes for unstable tests

vlad-dargel:


tests/e2e/pageObjects/components/browser/bulk-actions.ts
```diff
@@ -41,6 +41,8 @@ export class BulkActions {
      * Open Bulk Actions and confirm deletion
      */
     async startBulkDelete(): Promise<void> {
+        // Wait for Bulk actions animation ends
+        await t.wait(1000);
```
mariasergeenko, line 45: can you do something like this 
await t.expect(this.actionButton.exists).ok('', { timeout: 10000 });

vlad-dargel: no, it won't work for testCafe
click(this.actionButton) already checks this, and it is visible on animation start
and testCafe is trying to click this button when animation not ended and it leads to misclick

mariasergeenko: it is very sad

mariasergeenko (approved):

PR #2226, opened 2023-06-22
#RI-4667 - Telemetry improvements

zalenskiSofteq:
#RI-4667 - Telemetry improvements

### DIFF
--- a/redisinsight/api/src/modules/database-recommendation/database-recommendation.analytics.ts
+++ b/redisinsight/api/src/modules/database-recommendation/database-recommendation.analytics.ts
@@ -12,13 +12,16 @@ export class DatabaseRecommendationAnalytics extends TelemetryBaseService {
   }
 
   sendCreatedRecommendationEvent(recommendation: DatabaseRecommendation, database: Database): void {
-    this.sendEvent(
-      TelemetryEvents.InsightsRecommendationGenerated,
-      {
-        recommendationName: recommendation.name,
-        databaseId: database.id,
-        provider: database.provider,
-      },
-    );
+    try {
+      this.sendEvent(
+        TelemetryEvents.InsightsRecommendationGenerated,
+        {
+          recommendationName: recommendation.name,
+          databaseId: database.id,
+          provider: database.provider,
+        },
+      );
+    } catch (e) {
+    }
   }
 }
--- a/redisinsight/api/src/modules/database-recommendation/database-recommendation.analytics.ts
+++ b/redisinsight/api/src/modules/database-recommendation/database-recommendation.analytics.ts
@@ -1,0 +1,24 @@
+import { Injectable } from '@nestjs/common';
+import { EventEmitter2 } from '@nestjs/event-emitter';
+import { TelemetryBaseService } from 'src/modules/analytics/telemetry.base.service';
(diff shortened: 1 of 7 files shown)


redisinsight/api/src/modules/database-recommendation/repositories/local.database.recommendation.repository.ts
```diff
@@ -33,6 +35,8 @@ export class LocalDatabaseRecommendationRepository extends DatabaseRecommendatio
     @InjectRepository(DatabaseRecommendationEntity)
     private readonly repository: Repository<DatabaseRecommendationEntity>,
     private eventEmitter: EventEmitter2,
+    private readonly analytics: DatabaseRecommendationAnalytics,
+    private readonly databaseService: DatabaseService,
```
arthosofteq, line 39: We should not inject this here. It is better to move it to upper layer so if at some point we will add another repository analytics will still work

redisinsight/api/src/modules/database-recommendation/repositories/local.database.recommendation.repository.ts
```diff
@@ -56,6 +60,10 @@ export class LocalDatabaseRecommendationRepository extends DatabaseRecommendatio
     );
     this.eventEmitter.emit(RecommendationEvents.NewRecommendation, [recommendation]);
 
+    const database = await this.databaseService.get(entity?.databaseId)
+
+    this.analytics.sendCreatedRecommendationEvent(recommendation, database)
+
```
arthosofteq, line 66: I found that you assume that entity might be null (`entity?.databaseId`)
if so - I would say that we have 1 of 2 problems:
1. Are you sure `await this.databaseService.get(entity?.databaseId)` will not throw Not Found error
2. if there is no Not Found error (from 1 option) I could see that inside `sendCreatedRecommendationEvent` there is no check for null and if at some point null will be passed inside it will throw an error

redisinsight/api/src/modules/database-recommendation/database-recommendation.analytics.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Database } from '../database/models/database';
+
+@Injectable()
+export class DatabaseRecommendationAnalytics extends TelemetryBaseService {
+  constructor(protected eventEmitter: EventEmitter2) {
+    super(eventEmitter);
+  }
+
+  sendCreatedRecommendationEvent(recommendation: DatabaseRecommendation, database: Database): void {
+    this.sendEvent(
```
arthosofteq, line 15: please wrap with try\catch

arthosofteq (changes requested):

redisinsight/api/src/modules/database-recommendation/repositories/local.database.recommendation.repository.ts
zalenskiSofteq: moved

redisinsight/api/src/modules/database-recommendation/database-recommendation.analytics.ts
zalenskiSofteq: added

arthosofteq (approved):

PR #2288, opened 2023-07-07
E2E refactoring of API tests

vlad-dargel:


tests/e2e/helpers/api/api-database.ts
```diff
@@ -1,253 +1,341 @@
...
+        databaseParameters: OSSClusterParameters
+    ): Promise<void> {
+        const requestBody = {
+            name: databaseParameters.ossClusterDatabaseName,
+            host: databaseParameters.ossClusterHost,
+            port: Number(databaseParameters.ossClusterPort),
+        };
+        const response = await sendRequest(
+            Methods.post,
+            '/databases',
```
mariasergeenko, line 105: can we store the part of url in the separate file e.g. in enum?

tests/e2e/helpers/api/api-database.ts
```diff
@@ -1,253 +1,341 @@
...
+            201,
+            requestBody
+        );
+        await t
+            .expect(await response.body.name)
```
mariasergeenko, line 110: can we also split SendRequest method to Put, Get and delete as separate method?
and I am bot sure about verifying status code in method. on one hand it is good to understand immediately that bd is not correct, on the other hand it is violate single responsibility
probably we can move the verification on the 1 layer above
what do you think?

PR #2335, opened 2023-07-13
Dependabot fixes

GnaneshKunal:
Fixes
- minimist
- class-validator
- socket.io-parser
- http-cache-semantics
- minimatch
- axios
- word-wrap
- engine.io
- nanoid

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,9 @@
   "homepage": "https://github.com/RedisInsight/RedisInsight#readme",
   "resolutions": {
     "**/node-sass": "^8.0.0",
-    "**/trim": "0.0.3"
+    "**/trim": "0.0.3",
+    "word-wrap": "npm:@aashutoshrathi/word-wrap"
+
   },
   "devDependencies": {
     "@babel/core": "^7.12.9",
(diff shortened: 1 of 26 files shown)


package.json
```diff
@@ -80,7 +80,9 @@
   "homepage": "https://github.com/RedisInsight/RedisInsight#readme",
   "resolutions": {
     "**/node-sass": "^8.0.0",
-    "**/trim": "0.0.3"
+    "**/trim": "0.0.3",
+    "word-wrap": "npm:@aashutoshrathi/word-wrap",
```
arthosofteq, line 84: why can't we use 1.2.4 official version? 
Also I think we must have some agreement about forks. From my point of view we should use only official libs and in some corner cases we might use OWN forks if it is very critical to fix something.

arthosofteq (changes requested):

package.json
arthosofteq: UPD: let's use official lib here and in all other places

GnaneshKunal: Yes. Looks like the official version was just released. Using `1.2.4` from the official library.

arthosofteq (approved):

PR #2340, opened 2023-07-14
E2e/feature/ri 4632 cluster triggered functions

mariasergeenko:


tests/e2e/tests/critical-path/triggers-and-functions/libraries.e2e.ts
```diff
@@ -195,3 +201,30 @@ test.after(async() => {
     await t.expect(await browserPage.isKeyIsDisplayedInTheList(streamKeyName)).ok('The stream key is not opened');
     await t.expect(browserPage.keyDetailsBadge.exists).ok('The key details is not opened');
 });
+
+test.before(async() => {
```
vlad-dargel, line 205: can you please add the link to this story as a comment?
so that we can quickly find the story in case of something

PR #2386, opened 2023-07-25
#RI-4699 - 500 error on connecting to user with -@dangerous rule

zalenskiSofteq:
#RI-4699 - 500 error on connecting to user with -@dangerous rule


AmirAllayarovSofteq (approved):
LGTM

PR #2400, opened 2023-07-26
#RI-4788

zalenskiSofteq:
#RI-4788 - [BE] CLOUD_FREE_DATABASE_FAILED event doesn't have provider and region parameters

### DIFF
--- a/redisinsight/api/src/modules/cloud/database/cloud-database.analytics.ts
+++ b/redisinsight/api/src/modules/cloud/database/cloud-database.analytics.ts
@@ -2,6 +2,9 @@ import { HttpException, Injectable } from '@nestjs/common';
 import { TelemetryBaseService } from 'src/modules/analytics/telemetry.base.service';
 import { EventEmitter2 } from '@nestjs/event-emitter';
 import { TelemetryEvents } from 'src/constants';
+import { CloudSubscriptionCapiService } from '../subscription/cloud-subscription.capi.service';
+import { CloudSubscriptionType } from '../subscription/models';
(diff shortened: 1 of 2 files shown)


redisinsight/api/src/modules/cloud/database/cloud-database.analytics.ts
```diff
@@ -13,7 +16,27 @@ export class CloudDatabaseAnalytics extends TelemetryBaseService {
...
 
-  sendCloudFreeDatabaseFailed(exception: HttpException, eventData: object = {}) {
-    this.sendFailedEvent(TelemetryEvents.CloudFreeDatabaseFailed, exception, eventData);
+  async sendCloudFreeDatabaseFailed(
+    exception: HttpException,
+    eventData: object = {},
+    cloudSubscriptionCapiService: CloudSubscriptionCapiService,
+    data: { planId: number, capiCredentials: CloudCapiAuthDto },
+  ) {
+    const fixedPlans = await cloudSubscriptionCapiService.getSubscriptionsPlans(
```
arthosofteq, line 25: I'm not with DI inside analytics. We avoided such cases till this PR...
From my point of view we should get all required data before sendCloudFreeDatabaseFailed invocation or do not send additional fields. 

If it is urgent I'll approve this with assumption that in the nearest future it will be implemented in according suggestions above BUT you have to wrap with `try {} catch {}` this function since we might crash entire api server with unhandled promise rejection.

arthosofteq (changes requested):

arthosofteq (approved):

PR #2434, opened 2023-08-04
use Api delete keys part #1

mariasergeenko:


mariasergeenko:
> Looks good, but can you please also put these existing API keys functions into class? as it is for api-database.ts?

do you mean wrap the functions in a class?

vlad-dargel:
> 

yes

PR #2453, opened 2023-08-10
add api key adding

mariasergeenko:


tests/e2e/tests/critical-path/database/import-databases.e2e.ts
```diff
@@ -126,7 +126,7 @@ test('Connection import modal window', async t => {
     await t.click(myRedisDatabasePage.removeImportedFileBtn);
     await t.expect(myRedisDatabasePage.importDbDialog.textContent).contains(defaultText, 'File not removed from import input');
 });
-test('Connection import from JSON', async t => {
+test.only('Connection import from JSON', async t => {
```
vlad-dargel, line 129: .only

and it's better to add the test inside of 
"Exporting Standalone, OSS Cluster, and Sentinel connection types" not 'Connection import from JSON'
there both export and import tested
you can select decompressor for database, export it and then check that it was imported too

vlad-dargel (changes requested):

tests/e2e/tests/critical-path/database/export-databases.e2e.ts
```diff
@@ -25,7 +25,7 @@ let foundExportedFiles: string[];
 fixture `Export databases`
     .meta({ type: 'critical_path', rte: rte.none })
     .page(commonUrl);
-test
+test.only
```
vlad-dargel, line 28: .only

the rest looks good to me

vlad-dargel (approved):

PR #2521, opened 2023-08-24
#RI-4894 - add rs preview regions

AmirAllayarovSofteq:


redisinsight/api/config/features-config.json
```diff
@@ -39,6 +39,16 @@
                 "provider": "GCP",
                 "regions": ["asia-northeast1"]
               }
+            ],
+            "redistackPreview": [
```
ViktarStarastsenka, line 43: redisStackPreview

ViktarStarastsenka: what I pointed is that it is not RediStack it is Redis Stack. Let's use the correct wording

AmirAllayarovSofteq: @arthosofteq we have badge, which mark region as redis stack preview, Basically the same logic as for triggers and functions badge icon

arthosofteq: @ViktarStarastsenka Sorry, question was to Amir. Anyway I suggest to not use spaces in literals. RedisStack - is better
@AmirAllayarovSofteq ok. Now I see. Requirements were changed. Before the general plan was to show different set of regions for users navigated from different components. So approve. :) However I don't understand why we didn't reimplement this part. It confuses me...

arthosofteq (approved):

redisinsight/api/config/features-config.json
AmirAllayarovSofteq: Sorry i did not catch you. What you mean when say "reimplement this part"?

arthosofteq (approved):
Approve as CO
@romansergeenkosofteq Please take a look at FE part. (not sure about ":global")

redisinsight/api/config/features-config.json
arthosofteq: never mind. all good :)

romansergeenkosofteq (approved):

PR #2600, opened 2023-09-19
#RI-4813 - [BE] Do not duplicate Cloud databases

zalenskiSofteq:
#RI-4813 - [BE] Do not duplicate Cloud databases

### DIFF
--- a/redisinsight/api/src/modules/database/exeptions/database-already-exists.exception.ts
+++ b/redisinsight/api/src/modules/database/exeptions/database-already-exists.exception.ts
@@ -9,7 +9,7 @@ export class DatabaseAlreadyExistsException extends HttpException {
       statusCode: HttpStatus.CONFLICT,
       error: 'DatabaseAlreadyExists',
       errorCode: CustomErrorCodes.DatabaseAlreadyExists,
-      result: {
+      resource: {
         databaseId,
       },
     };
--- a/redisinsight/api/src/modules/database/exeptions/database-already-exists.exception.ts
+++ b/redisinsight/api/src/modules/database/exeptions/database-already-exists.exception.ts
@@ -1,0 +1,19 @@
+import { HttpException, HttpExceptionOptions, HttpStatus } from '@nestjs/common';
+import { CustomErrorCodes } from 'src/constants';
+import ERROR_MESSAGES from 'src/constants/error-messages';
+
+export class DatabaseAlreadyExistsException extends HttpException {
+  constructor(databaseId: string, message = ERROR_MESSAGES.DATABASE_ALREADY_EXISTS, options?: HttpExceptionOptions) {
+    const response = {
+      message,
+      statusCode: HttpStatus.CONFLICT,
+      error: 'DatabaseAlreadyExists',
+      errorCode: CustomErrorCodes.DatabaseAlreadyExists,
+      result: {
+        databaseId,
+      },
+    };
+
+    super(response, response.statusCode, options);
+  }
+}
--- a/redisinsight/api/src/modules/database/repositories/database.repository.ts
+++ b/redisinsight/api/src/modules/database/repositories/database.repository.ts
@@ -27,8 +27,9 @@ export abstract class DatabaseRepository {
   /**
    * Create database
    * @param database
+   * @param uniqueCheck
    */
-  abstract create(database: Database): Promise<Database>;
+  abstract create(database: Database, uniqueCheck: boolean): Promise<Database>;
 
   /**
    * Update database with new data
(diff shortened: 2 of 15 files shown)


redisinsight/api/src/modules/database/exeptions/database-already-exists.exception.ts
```diff
@@ -0,0 +1,19 @@
...
+import ERROR_MESSAGES from 'src/constants/error-messages';
+
+export class DatabaseAlreadyExistsException extends HttpException {
+  constructor(databaseId: string, message = ERROR_MESSAGES.DATABASE_ALREADY_EXISTS, options?: HttpExceptionOptions) {
+    const response = {
+      message,
+      statusCode: HttpStatus.CONFLICT,
+      error: 'DatabaseAlreadyExists',
+      errorCode: CustomErrorCodes.DatabaseAlreadyExists,
+      result: {
```
arthosofteq, line 12: let's use "resource" instead of "result" if possible, to be compatible with cloud jobs

redisinsight/api/src/modules/database/repositories/local.database.repository.ts
```diff
@@ -90,6 +92,17 @@ export class LocalDatabaseRepository extends DatabaseRepository {
    * @param database
    */
   public async create(database: Database): Promise<Database> {
+    // Do not create a connection if it triggered from cloud and have the same fields
+    if (database.cloudDetails) {
```
arthosofteq, line 96: is it better to check "cloud database id" in the cloudDetails and only then other fields (may be even without host and port)

redisinsight/api/src/modules/database/repositories/local.database.repository.ts
```diff
@@ -90,6 +92,17 @@ export class LocalDatabaseRepository extends DatabaseRepository {
...
+      const existingEntity = await this.findEntity(
+        database,
+        ['host', 'port', 'username', 'password', 'tls'],
```
arthosofteq, line 99: "tls" here just a flag. Not sure that it is enough... certs & key body should be checked too, no?

redisinsight/api/src/modules/database/repositories/local.database.repository.ts
```diff
@@ -205,4 +218,20 @@ export class LocalDatabaseRepository extends DatabaseRepository {
...
     return decryptedEntity;
   }
+
+  /**
+   * Find existing entity by array of fields
+   * @param database
+   * @param fields
+   * @private
+   */
+  private async findEntity(database: Database, fields: string[]): Promise<Database | null> {
```
arthosofteq, line 228: I don't see that you added this method into common databse.repository interface when it should be there too

redisinsight/api/src/modules/database/repositories/local.database.repository.ts
```diff
@@ -90,6 +92,17 @@ export class LocalDatabaseRepository extends DatabaseRepository {
    * @param database
    */
   public async create(database: Database): Promise<Database> {
+    // Do not create a connection if it triggered from cloud and have the same fields
+    if (database.cloudDetails) {
+      const existingEntity = await this.findEntity(
+        database,
+        ['host', 'port', 'username', 'password', 'tls'],
+      );
+      if (existingEntity) {
```
arthosofteq, line 101: based on "findEntity" function I assume that you forgot encrypt fields before comparing values. Here you are receiving  "Database" object with plain values when "Database Entity" stores encrypted values.

arthosofteq (changes requested):

redisinsight/api/src/modules/database/exeptions/database-already-exists.exception.ts
zalenskiSofteq: changed

redisinsight/api/src/modules/database/repositories/local.database.repository.ts
zalenskiSofteq: it is a private method, like near methods `decryptEntity` and `encryptEntity`

redisinsight/api/src/modules/database/repositories/local.database.repository.ts
zalenskiSofteq: added certs and key body

redisinsight/api/src/modules/database/repositories/local.database.repository.ts
zalenskiSofteq: added encryption

redisinsight/api/src/modules/database/repositories/database.repository.ts
```diff
@@ -27,8 +27,9 @@ export abstract class DatabaseRepository {
   /**
    * Create database
    * @param database
+   * @param uniqueCheck
    */
-  abstract create(database: Database): Promise<Database>;
+  abstract create(database: Database, uniqueCheck: boolean): Promise<Database>;
```
arthosofteq, line 32: Should be optional, no?

arthosofteq (approved):

redisinsight/api/src/modules/database/repositories/database.repository.ts
zalenskiSofteq: in the service added default value

PR #2649, opened 2023-10-02
Be/feature/ri 4481 hide security fields

AmirAllayarovSofteq:


redisinsight/api/test/api/database/POST-databases-clone-id.test.ts
arthosofteq, line 1: I assume these tests will fail on main-feature-branch pipieline

redisinsight/api/src/modules/database/database.service.ts
```diff
@@ -273,4 +347,14 @@ export class DatabaseService {
       { groups: ['security'] },
     ));
   }
+
+  private checkSecurityFields(model: Database): Database {
```
arthosofteq, line 351: Let's name it "hide" securty fields or "convert" or whatever, but not "check"

redisinsight/api/src/modules/database/database.service.ts
```diff
@@ -197,6 +232,45 @@ export class DatabaseService {
...
+  /**
+   * Clone database
+   * @param dto
+   */
+  public async clone(
+    id: string,
+    dto: PartialDatabaseDto,
+  ): Promise<Database> {
+    this.logger.log(`Clone database connection: ${id}`);
+    const oldDatabase = omit(await this.get(id, true), 'sshOptions.id');
```
arthosofteq, line 259: you can omit id and sshOptions at the same time

redisinsight/api/src/modules/database/database.service.ts
arthosofteq, line 1: I don't think that create, update and other methods should return data without security fields. We might use these methods internally.

redisinsight/api/src/modules/database/database.service.spec.ts
```diff
@@ -117,6 +137,23 @@ describe('DatabaseService', () => {
     });
   });
 
+  describe('getWithoutSecurityFields', () => {
```
arthosofteq, line 140: lack of other hidden fields

redisinsight/api/src/modules/database/database.service.ts
```diff
@@ -197,6 +232,45 @@ export class DatabaseService {
...
+    const database = merge({}, omit(oldDatabase, 'id'), dto);
+    if (Object.keys(omitBy(dto, isUndefined)).some((field) => this.connectionFields.includes(field))) {
+      return await this.create(database);
+    }
+
+    const createdDatabase = await this.repository.create({
+      ...classToClass(Database, database),
+      new: true,
+    }, false);
+
```
arthosofteq, line 269: It is by requirements? mark cloned databases as new only when connection fields affected?

redisinsight/api/src/modules/database/database.service.ts
```diff
@@ -197,6 +232,45 @@ export class DatabaseService {
...
+  ): Promise<Database> {
+    this.logger.log(`Clone database connection: ${id}`);
+    const oldDatabase = omit(await this.get(id, true), 'sshOptions.id');
+    const database = merge({}, omit(oldDatabase, 'id'), dto);
+    if (Object.keys(omitBy(dto, isUndefined)).some((field) => this.connectionFields.includes(field))) {
+      return await this.create(database);
+    }
+
+    const createdDatabase = await this.repository.create({
+      ...classToClass(Database, database),
```
arthosofteq, line 266: we are converting to class and back to plain?

redisinsight/api/src/modules/database/database.service.ts
```diff
@@ -148,32 +180,35 @@ export class DatabaseService {
     let database = merge({}, oldDatabase, dto);
 
     try {
-      database = await this.databaseFactory.createDatabaseModel(database);
+      if (Object.keys(omitBy(dto, isUndefined)).some((field) => this.connectionFields.includes(field))) {
```
arthosofteq, line 183: I like this check. Well done
Anyway let's move it to own method (even static) since I already found 2 places with this copy paste

arthosofteq (changes requested):
Not bad
Anyway I feel like we are missing major requirements by our architecture. Not sure that we can go as is.
Feel like missed a lot of tests and tests which we already have will failed

redisinsight/api/src/modules/database/database.service.ts
AmirAllayarovSofteq: We marked all cloned databases as new (when connection fields affected we create database and check connection, if connection fields are not affected we will create db instance in repo without check connection) false in 268 line it is a uniqCheck - we don't need this check (only for sso flow)

redisinsight/api/src/modules/database/database.analytics.ts
```diff
@@ -23,7 +23,7 @@ export class DatabaseAnalytics extends TelemetryBaseService {
 
   sendInstanceAddedEvent(
     instance: Database,
-    additionalInfo: RedisDatabaseInfoResponse,
+    additionalInfo?: RedisDatabaseInfoResponse,
```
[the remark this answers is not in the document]
AmirAllayarovSofteq: Thank you

redisinsight/api/src/modules/database/database.service.ts
AmirAllayarovSofteq: rename checkSecurityFields with hideSecurityFields

redisinsight/api/src/modules/database/database.service.spec.ts
AmirAllayarovSofteq: add test for ssh password and ssh passphrase (password was already implemented)

redisinsight/api/src/modules/database/database.service.ts
AmirAllayarovSofteq: Agreed. Thank you

PR #2675, opened 2023-10-11
#RI-5003 show scan more always when next cursor is non-zero

arthosofteq:


zalenskiSofteq:
tests are failed

arthosofteq:
> tests are failed

@zalenskiSofteq Fixed tests. 
Also I had to skip old test for keys details in the footer since nextCursor is not passed into the ScanMore component and it is not shown there anymore. 
I've added "todo" so please take a look later. Probably we must remove obsolete tests and code

romansergeenkosofteq (approved):

PR #2703, opened 2023-10-20
E2e/feature/ri 4815 portion of string

vlad-dargel:


tests/e2e/common-actions/databases-actions.ts
```diff
@@ -69,6 +69,25 @@ export class DatabasesActions {
         }
         return matchedFiles;
     }
+
+    /**
+     * Find file by name starts from directory
+     * @param dir The path directory of file
+     * @param fileStarts The file name should start from
+     */
+    async findFileByFileStarts(dir: string, fileStarts: string): Promise<number> {
```
mariasergeenko, line 78: can we do method more universal:
fileStarts - it is a filter that can be optional and in such case we get common count of file in the folder
and method can be named like: getFileCount

vlad-dargel: thanks, updated

mariasergeenko (approved):

PR #2730, opened 2023-10-29
fix: Optimizing png files using zopfli

pea-sys:
I have optimized a png file using zopfli.
There is no change in appearance.
The total size of png files stored in the resource folder changed as follows.

-----------------

Before:134492
After:93438


zalenskiSofteq (approved):
Looks good!

PR #2750, opened 2023-11-07
E2e/feature/ri 4943 rework tutorials

mariasergeenko:


tests/e2e/pageObjects/components/insights-panel.ts
```diff
@@ -22,6 +25,12 @@ export class InsightsPanel {
         }
     }
 
+    /**
+     * get acrive tab
+     */
+    async getActiveTab(): Promise<string> {
```
vlad-dargel, line 31: looks like better to rename it getActiveTabName()

vlad-dargel (commented):
Great job 👍

tests/e2e/pageObjects/components/insights-panel.ts
mariasergeenko: fixed

PR #2762, opened 2023-11-13
Add autoupdate continius checks

arthosofteq:


redisinsight/desktop/app.ts
```diff
@@ -70,10 +71,12 @@ const init = async () => {
 
     await windowFactory(WindowType.Main, splashWindow, { parsedDeepLink })
 
-    checkForUpdate(process.env.MANUAL_UPGRADES_LINK || process.env.UPGRADES_LINK)
-  } catch (_err) {
-    const error = _err as Error
-    console.log(wrapErrorMessageSensitiveData(error))
+    initAutoUpdateChecks(
+      process.env.MANUAL_UPGRADES_LINK || process.env.UPGRADES_LINK,
+      parseInt(process.env.RIAUTOUPDATEINTERVAL, 10) || 84 * 3600 * 1000,
```
zalenskiSofteq, line 76: i think `RIAUTOUPDATEINTERVAL` should be with underscores `RI_AUTOUPDATE_INTERVAL`

arthosofteq: nope. we have new name convention. all other envs will be renamed

zalenskiSofteq (approved):

redisinsight/desktop/app.ts
zalenskiSofteq: interesting the reason. This new name convention is less readable.

PR #2768, opened 2023-11-13
#RI-5103 - Add wrapper to workbench 

irinaP-local:


arthosofteq (commented):
Looks good
But if you decided to remove fields from command-execution entity, please make sure that:
1. FE will not fail
2. Create migrations (not sure if without migrations we will not fail)
3. After migrations: double check that data succesfully removed on startup

arthosofteq (approved):
lgtm

PR #2804, opened 2023-11-22
#RI-5106 - add wrapper to overview

irinaP-local:


redisinsight/api/src/modules/redis/client/redis.client.ts
```diff
@@ -108,6 +108,10 @@ export abstract class RedisClient {
 
   abstract getCurrentDbIndex(): Promise<number>;
 
+  info(): Promise<any> {
+    return Promise.resolve(null);
+  }
```
arthosofteq, line 113: why do we need this dummy realization?

redisinsight/api/src/modules/redis/utils/totals.util.ts
arthosofteq, line 1: we already have such functionality in this folder. please use it

arthosofteq (approved):

PR #2849, opened 2023-12-05
#RI-5104 - add connection wrapper to profiler

irinaP-local:


redisinsight/api/src/modules/redis/redis.client.factory.ts
```diff
@@ -81,12 +81,16 @@ export class RedisClientFactory implements OnModuleInit {
     options: IRedisConnectionOptions = {},
   ): Promise<RedisClient> {
     const opts = RedisClientFactory.prepareConnectionOptions(options);
+    let connectionStrategy = await this.getConnectionStrategy(opts.clientLib);
+
+    if (clientMetadata.context === ClientContext.Profiler) {
+      connectionStrategy = this.ioredisConnectionStrategy;
+    }
```
arthosofteq, line 88: I think that we shouldn't put such logic to core functionality. I prefer to send `strategy` enum from profiler as an argument to force assign required one.

redisinsight/api/src/modules/redis/redis.client.factory.ts
```diff
@@ -125,21 +129,23 @@ export class RedisClientFactory implements OnModuleInit {
     });
 
     const opts = RedisClientFactory.prepareConnectionOptions(options);
+    let connectionStrategy = await this.getConnectionStrategy(opts.clientLib);
+
+    if (clientMetadata.context === ClientContext.Profiler) {
+      connectionStrategy = this.ioredisConnectionStrategy;
+    }
```
arthosofteq, line 136: no need to duplicate this logic after `strategy` arg introduced

arthosofteq (changes requested):

PR #2927, opened 2024-01-09
RI-5288: Hide export button for RDI instance list

kyle-marcum:


redisinsight/ui/src/components/item-list/ItemList.tsx
```diff
@@ -21,7 +21,8 @@ export interface Props<T> {
   editedInstance: Nullable<T>
   columnVariations: EuiBasicTableColumn<T>[][]
   onDelete: (ids: T[]) => void
-  onExport: (ids: T[], withSecrets: boolean) => void
+  disableExport?: boolean
```
AmirAllayarovSofteq, line 24: Maybe better to cal it hideExport or hideExportBtn. In my opinion, disable - we still have to display it

kyle-marcum: agreed

PR #2929, opened 2024-01-09
#RI-5138 - add dry run endpoint

AmirAllayarovSofteq:


redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
```diff
@@ -31,6 +34,31 @@ export class ApiRdiClient extends RdiClient {
     return null;
   }
 
+  async dryRunJob(data: DryRunJobDto): Promise<DryRunJobResponseDto> {
+    const results: any = {};
```
arthosofteq, line 38: is it known type? why `any`?

redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
```diff
@@ -31,6 +34,31 @@ export class ApiRdiClient extends RdiClient {
...
+        { input: data.input, job: data.job, test_output: true },
+      );
+      results.commands = { status: DyRunJobStatus.Success, data: commands.data };
+    } catch (e) {
+      results.commands = { status: DyRunJobStatus.Fail, error: e.message };
+    }
+
+    return results;
+  }
+
```
arthosofteq, line 61: I prefer to have 2 separate methods to get transformations and commands. You can run them in parallel and compose results in a parent method `dryRunJob`. WDYT?

redisinsight/api/src/modules/rdi/models/rdi-dry-run.ts
```diff
@@ -0,0 +1,13 @@
...
+  Fail = 'fail',
+}
+
+export class RdiDryRunJobResult {
+  status: DyRunJobStatus;
+
+  // TODO validate with RDI team
+  data?: any;
+
+  error?: string;
```
arthosofteq, line 12: not sure if error will be a plain string since we must have some logic on it. But for now it is more then enough

arthosofteq (changes requested):

redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
AmirAllayarovSofteq: Good idea, thank you

redisinsight/api/src/modules/rdi/models/rdi-dry-run.ts
AmirAllayarovSofteq: Yep, will see what kind of error we will have

arthosofteq (approved):
good

PR #2954, opened 2024-01-15
add test for edit rdi instance

mariasergeenko:


tests/e2e/pageObjects/rdi-instances-list-page.ts
```diff
@@ -73,6 +73,23 @@ export class RdiInstancesListPage extends BaseOverviewPage {
         }
     }
 
+    /**
+     * Edit Rdi by name
+     * @param dbName The name of the rdi to be deleted
+     */
+    async editRdiByName(dbName: string): Promise<void> {
```
vlad-dargel, line 80: but it is the same as the method below? clickOnEditRdiByName

mariasergeenko: yes, thank you

vlad-dargel (approved):

PR #2957, opened 2024-01-16
#RI-5148 - add refresh pipeline btn

AmirAllayarovSofteq:


romansergeenkosofteq (approved):

redisinsight/ui/src/pages/rdi/pipeline/components/navigation/Navigation.tsx
```diff
@@ -15,7 +17,14 @@ enum RdiPipelineTabs {
   Jobs = PageNames.rdiPipelineJobs,
 }
 
-const defaultNavList = [
+interface INavItem {
+  id: RdiPipelineTabs
+  title: string
+  fileName: string
+  isShowLoader?: boolean
```
kyle-marcum, line 24: I was going to comment that this variable name reads strange, but after checking the rest of this codebase it would seem that the `is` prefix is used all over the place to denote a boolean. No action needed. Just wanted to comment on it.

redisinsight/ui/src/pages/rdi/pipeline/components/tab/Tab.tsx
```diff
@@ -35,6 +37,7 @@ const Tab = (props: IProps) => {
         >
           <EuiIcon type="document" className="rdi-pipeline-nav__fileIcon" />
           <EuiText className="rdi-pipeline-nav__text">{fileName}</EuiText>
+          {isLoading && <EuiLoadingSpinner data-testid="rdi-nav-config-loader" className={styles.loader} />}
```
kyle-marcum, line 40: do you think it's necessary to have a unit test that checks for the loading indicator when `isLoading = true`?

redisinsight/ui/src/pages/rdi/pipeline/components/refresh-pipeline-popover/RefreshPipelinePopover.tsx
```diff
@@ -0,0 +1,102 @@
...
+    dispatch(fetchRdiPipeline(rdiInstanceId))
+  }
+
+  const handleRefreshWarning = () => {
+    setIsPopoverOpen(true)
+    sendEventTelemetry({
+      event: TelemetryEvent.RDI_PIPELINE_REFRESH_CLICKED,
+      eventData: {
+        id: rdiInstanceId,
+        jobsNumber: data?.jobs?.length || 'none',
```
kyle-marcum, line 39: any specific reason we couldn't just pass `0` instead of `'none'`?

kyle-marcum:
I noticed that the behavior of this popover doesn't close `onBlur`. We might want to consider that this popover should be closable in some way besides the `Refresh` button.

redisinsight/ui/src/pages/rdi/pipeline/components/tab/Tab.tsx
AmirAllayarovSofteq: I don't think this test is necessary. But i can if needed

redisinsight/ui/src/pages/rdi/pipeline/components/refresh-pipeline-popover/RefreshPipelinePopover.tsx
AmirAllayarovSofteq: Not specific, but it is more clear. Because in my opinion there are two cases:
 - we have pipeline, but user removed all jobs. Jobs length = 0
 - we have failed pipeline request, it means there is no pipeline, so none mo clear for me

redisinsight/ui/src/pages/rdi/pipeline/components/navigation/Navigation.tsx
```diff
@@ -102,7 +114,12 @@ const Navigation = () => {
 
   return (
     <div className={styles.wrapper}>
-      <EuiTextColor className={styles.title} component="div">Pipeline Management</EuiTextColor>
+      <div className={styles.title}>
+        <EuiTextColor component="div">Pipeline Management</EuiTextColor>
+        <div className={styles.actions}>
```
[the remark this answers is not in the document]
AmirAllayarovSofteq: It is not a big deal, if there is no css, className will be blank. But i will remove it. Thank you

kyle-marcum:
> I noticed that the behavior of this popover doesn't close `onBlur`. We might want to consider that this popover should be closable in some way besides the `Refresh` button.

To re-create this issue just click the 'refresh' button twice. I should've clarified this with my original comment.

PR #2963, opened 2024-01-18
add test for refresh rdi

mariasergeenko:


vlad-dargel (approved):
Looks good
can you add this test to checklist?

PR #2970, opened 2024-01-19
#RI-5116 - limit links [BE]

irinaAbramova-dev:


redisinsight/api/src/common/validators/github-link.validator.ts
```diff
@@ -0,0 +1,23 @@
...
+import config from 'src/utils/config';
+
+const SERVER_CONFIG = config.get('server');
+
+@ValidatorConstraint({ name: 'GitHubLink', async: false })
+export class GitHubLink implements ValidatorConstraintInterface {
+  validate(value: any) {
+    // TODO: temporary solution for integration tests
+    if (SERVER_CONFIG.env === 'test') return true;
+
```
arthosofteq, line 14: Let's use existing nock lib to mock http response for integration tests

arthosofteq: Or we have config/test.js for test environment

arthosofteq (changes requested):
Not sure that this PR fixes vulnerability. Please double check

arthosofteq (approved):

PR #3045, opened 2024-02-14
E2e/feature/ri 5376 parse redis url

mariasergeenko:


vlad-dargel (approved):

vlad-dargel (commented):
Can you add test name somewhere? as comment

PR #3063, opened 2024-02-20
Build/deep linking/main

mariasergeenko:


.circleci/config.yml
zalenskiSofteq, line 1: in the e2e flow alse we have "choco install nodejs"

vlad-dargel: e2e tests are running without any errors

zalenskiSofteq (approved):

PR #3100, opened 2024-02-27
RI-5150: Added api resources for RDI pipeline statistics

kyle-marcum:


arthosofteq (approved):

arthosofteq:
@kyle-marcum I think we must review entire module and find what we must rework (including this part)

PR #3136, opened 2024-03-05
e2e/feature/RI-5462_dedicated_editor_for_sql_rdi

vlad-dargel:


tests/e2e/pageObjects/components/common/button.ts
```diff
@@ -0,0 +1,6 @@
+import { Selector } from 'testcafe';
+
+export class Button {
```
mariasergeenko, line 3: can we raname the class
since button for me can't contain list of locators ( it can take locator and has some actions like click hover) but if we just store only 2 locators we need to specify what exact locators we have

vlad-dargel: renamed to EditorButton

vlad-dargel: actually I created this class more for future, to move all common locators related to buttons there

mariasergeenko (approved):

mariasergeenko (approved):

PR #3143, opened 2024-03-07
E2e/feature/ri 4616 rdi support

mariasergeenko:
status page test

### DIFF
--- a/tests/e2e/tests/web/critical-path/rdi/status.e2e.ts
+++ b/tests/e2e/tests/web/critical-path/rdi/status.e2e.ts
@@ -62,13 +62,14 @@ test('Verify that user can see data on the status page', async() => {
 
 test('Verify that user can refresh tables', async() => {
 
-    let label =  await rdiStatusPage.processingPerformanceRefreshLabel.textContent;
-    await t.expect(label).contains('Auto refresh', 'Auto refresh is not enabled');
+    let message =  await rdiStatusPage.processingPerformanceRefreshMessage.textContent;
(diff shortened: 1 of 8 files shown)


tests/e2e/tests/web/critical-path/rdi/status.e2e.ts
```diff
@@ -0,0 +1,74 @@
...
+test('Verify that user can refresh tables', async() => {
+
+    let label =  await rdiStatusPage.processingPerformanceRefreshLabel.textContent;
+    await t.expect(label).contains('Auto refresh', 'Auto refresh is not enabled');
+
+    label =  await rdiStatusPage.clientRefreshLabel.textContent;
+    await t.expect(label).contains('Last refresh', 'Last refresh is enabled');
+
+    await t.click(rdiStatusPage.refreshStreamsButton);
+    await rdiStatusPage.dataStreamsRefreshLabel.textContent;
```
vlad-dargel, line 72: what this row do?

tests/e2e/tests/web/critical-path/rdi/status.e2e.ts
```diff
@@ -0,0 +1,74 @@
...
+    await t.expect(label).contains('Last refresh', 'Last refresh is enabled');
```
vlad-dargel, line 73: not sure that this actually checks refreshing
probably better to check that Last refresh value becomes 'now' or something

vlad-dargel (approved):

PR #3166, opened 2024-03-18
RI-5483: Fixed redirect to configuration tab when job name changes

kyle-marcum:


redisinsight/ui/src/pages/rdi/pipeline-management/pages/jobs/Jobs.tsx
```diff
@@ -18,6 +18,7 @@ import TemplatePopover from 'uiSrc/pages/rdi/pipeline-management/components/temp
 const Jobs = () => {
   const { rdiInstanceId, jobName } = useParams<{ rdiInstanceId: string, jobName: string }>()
   const [decodedJobName, setDecodedJobName] = useState<string>(decodeURIComponent(jobName))
+  const [previousJobName, setPreviousJobName] = useState<string>('')
```
romansergeenkosofteq, line 21: Maybe it is better to put previousJobName as ref since we don't need to re-render component

romansergeenkosofteq (approved):

PR #3229, opened 2024-04-06
RI-5597: Fix Dependabot alerts

GnaneshKunal:
#### Bumps
- webpack-dev-middleware
- express (and all NestJS packages)
- follow-redirects
- IP

# Important

I noticed that platform-express used a fixed older version of the express package. While we can force install a specific express version via NPM resolutions, I have explored upgrading the platform-express package to bump the express package. I also took the opportunity to bump the rest of the NestJS packages. Also, I have removed the nest-router package, for which support exists in the existing NestJS library itself.


package.json
```diff
@@ -84,8 +84,7 @@
     "word-wrap": "1.2.4",
     "electron-builder/app-builder-lib/@electron/universal/dir-compare/minimatch": "^3.0.5",
     "**/semver": "^7.5.2",
-    "rawproto/protobufjs": "^7.2.4",
-    "**/follow-redirects": "^1.15.4"
```
arthosofteq, line 88: @GnaneshKunal Are you sure that we must remove this line?
@zalenskiSofteq @romansergeenkosofteq Do you happy to know why we've added it? Is it safe to remove?

ArtemHoruzhenko (dismissed):
I hope that router was the only 1 thing affected by bumping critical version of core core deps
from BE perspective looks good. Thanks!

let see what guys answer about removing follow-redirects

package.json
GnaneshKunal: Yeah. Safe for removal. `follow-redirects` now has a new version released, `1.15.6`, which must be upgraded, but the NPM resolution has blocked the upgrade. So, it's necessary to remove this.

zalenskiSofteq (dismissed):
fix conflict please

PR #3231, opened 2024-04-06
RI-5523: Fallback to non-native big integers for JSON formatting

GnaneshKunal:
Please check - https://redislabs.atlassian.net/browse/RI-5523?focusedCommentId=1793216


GnaneshKunal:
@AmirAllayarovSofteq, Can you approve the CI job and verify it?

AmirAllayarovSofteq (changes requested):
May be I'm doing something wrong. But from example :
Screenshot from 2024-05-20 12-18-46
Screenshot from 2024-05-20 12-18-58
So i guess problem doesn't fixed

GnaneshKunal:
@AmirAllayarovSofteq 

I have pointed this out in the ticket. According to the JSON specification, only big numerics are supported, not big floating-point numbers. The only way to view this is to use scientific notation (which is usually the use case for big numbers). Our vector format mostly has big floating-point numbers that use scientific notation only. 

@ViktarStarastsenka Maybe we can show a chip denoting that the number has been converted to scientific notation?

AmirAllayarovSofteq:
@GnaneshKunal Sorry, my bad.
@ViktarStarastsenka So if it is okay for you, I'm fine with it. But in "edit" mode user will see another value,

AmirAllayarovSofteq (approved):

AmirAllayarovSofteq:
We can display big numbers. 
Screenshot from 2024-05-21 11-11-20
I had error: useNativeBigIntq is not defined.

AmirAllayarovSofteq:
@GnaneshKunal 
1 - We can improve JsonPretty. ` const data = JSONBigInt({ useNativeBigInt }).parse(value)` will return scientific notation fro huge numbers or numbers with huge float part. We can handle scientific notation. This value will have own _isBigNumber property. So we can handle those values in JsonPretty component. 
And update JsonPrimitive. Scientific notation has toString()

2 - Probably in JSONViewer we should add something like this:
`try {
    const jsonValue = RenderJSONValue(value, expanded, space, useNativeBigInt)
    return {
      value: jsonValue,
      isValid: true
    }
  } catch (e) {
      try {
        const jsonValue = RenderJSONValue(value, expanded, space, false)
        return {
          value: jsonValue,
          isValid: true
        }
      } catch (e) {
        return {
          value,
          isValid: false
        }
      }
  }`
  We will cover both cases. When values will have native BigInt or will not have. And can parse both. I'm not sure here. Up to you

GnaneshKunal:
Done. Thanks @AmirAllayarovSofteq. That was an excellent suggestion.

image


I have extended the support to Vector types as well:

image

AmirAllayarovSofteq (approved):
LGTM! Maybe you can add one test for _isBigNumer in JsonPretty

GnaneshKunal:
@AmirAllayarovSofteq,

I have added the tests.

AmirAllayarovSofteq (approved):
Good job!

AmirAllayarovSofteq (approved):

PR #3316, opened 2024-04-25
CR-15: Added feature flags to be used in cloud

kyle-marcum:


arthosofteq:
@romansergeenkosofteq It is not in the scope of this task.

arthosofteq (approved):
So you decided to keep all flags in the feature-config.json?
Ok. let's go with this approach

romansergeenkosofteq:
> @romansergeenkosofteq It is not in the scope of this task.

then we don't need to put navigation items under feature flag now?

arthosofteq:
@romansergeenkosofteq 
No, we need.
The scope of this particular task was defined and according to it - everything is good.
Disabling entire page using dynamic route needed to be discussed and was descoped.
Except this any other comments?

romansergeenkosofteq:
> @romansergeenkosofteq No, we need. The scope of this particular task was defined and according to it - everything is good. Disabling entire page using dynamic route needed to be discussed and was descoped. Except this any other comments?

thats mean that users will be able to open these pages under feature flag by url or from links in our application. We have a lot of places with links leading to theses pages under feature flag

arthosofteq:
@romansergeenkosofteq correct. it is expected behavior for now

romansergeenkosofteq (approved):

PR #3333, opened 2024-05-01
PoC e2e for sso

vlad-dargel:


tests/e2e/helpers/constants.ts
```diff
@@ -80,6 +80,7 @@ export enum ResourcePath {
 export enum ExploreTabs {
     Explore  = 'Explore',
     Tips = 'Tips',
+    RedisCopilot = 'Redis Copilot'
```
mariasergeenko, line 83: we can do not add it since it is separate button right now

tests/e2e/helpers/database.ts
```diff
@@ -345,6 +345,9 @@ export class DatabaseHelper {
     async acceptLicenseTerms(): Promise<void> {
         await t.maximizeWindow();
         await userAgreementDialog.acceptLicenseTerms();
+        if (await myRedisDatabasePage.AddRedisDatabase.aiChatMessage.exists) {
```
mariasergeenko, line 348: can you add TODO here, that we need to remove it with the June release, since as I remember it will be available only during the next sprint

tests/e2e/helpers/google-authorization.ts
```diff
@@ -0,0 +1,37 @@
...
+
+        // Wait for the authorization to complete and the redirect to your specified URI
+        await driver.wait(until.urlContains('#success'), 30000);
+
+        const currentUrl = await driver.getCurrentUrl();
+        const parts = currentUrl.split('?');
+        const modifiedUrl = parts.length > 1 ? parts[1] : currentUrl;
+        const redirectUrl = `${protocol + callbackUrl  }?${  modifiedUrl}`;
+
+        await open(redirectUrl, { app: { name: 'Redis Insight' } });
```
mariasergeenko, line 29: could you please add comment here, that await open works with deeeplink

mariasergeenko (commented):
thanks for the great job

mariasergeenko (approved):

PR #3347, opened 2024-05-07
#RI-5463 - [BE] Add auto-completion for JMESPath expressions

zalenskiSofteq:
#RI-5463 - [BE] Add auto-completion for JMESPath expressions


redisinsight/api/src/modules/rdi/rdi-pipeline.controller.ts
```diff
@@ -100,4 +100,15 @@ export class RdiPipelineController {
   ): Promise<unknown> {
     return this.rdiPipelineService.getTemplate(rdiClientMetadata, options);
   }
+
+  @Get('/job-functions')
+  @ApiEndpoint({
+    description: 'Get job functions',
+    responses: [{ status: 200, type: Rdi }],
```
AmirAllayarovSofteq, line 107: I guess we should not use Rdi type here

AmirAllayarovSofteq (approved):
LGTM. I guess better remove type annotation or create type for functions response

redisinsight/api/src/modules/rdi/rdi-pipeline.controller.ts
zalenskiSofteq: removed

PR #3384, opened 2024-05-15
E2e/feature/ri 5658 update the empty rdi screen

vlad-dargel:


tests/e2e/tests/web/smoke/database/add-db-from-welcome-page.e2e.ts
mariasergeenko, line 1: should we have it?

tests/e2e/tests/web/critical-path/rdi/add-rdi-instance.e2e.ts
```diff
@@ -140,3 +140,18 @@ test('Verify that button is displayed if user does not enter all mandatory infor
         await browserActions.verifyTooltipContainsText(text, true);
     }
 });
+test('Verify that user can see the Redis Data Integration message on the empty RDI list', async t => {
+    const noInstancesMessage = 'Redis Data Integration (RDI) synchronizes data from your existing database into Redis in near-real-time. We\'ve done the heavy lifting so you can turn slow data into fast data without coding.';
+    // const externalPageLink = 'https://docs.redis.com/rdi-preview/rdi/quickstart/'
+
+    await t.expect(rdiInstancesListPage.emptyRdiList.withText(noInstancesMessage).exists).ok('Empty RDI page message not displayed');
```
mariasergeenko, line 147: probably the test doesn't make sence, since we have method addRdi() and if the form doesn't opened all test with adding rdi manually will be failed 
we can just add the first verification to another test

tests/e2e/tests/web/critical-path/database/connecting-to-the-db.e2e.ts
```diff
@@ -181,3 +181,23 @@ test
         await myRedisDatabasePage.clickOnDBByName(sshDbPass.databaseName);
         await Common.checkURLContainsText('browser');
     });
+test
+    .meta({ rte: rte.none })
```
mariasergeenko, line 185: shouldn't we update existing tests where we added db manually, if such tests exist?
and I am not sure about saving the default tab, will it be or not. I mean that if the previous executed tests were RDI tests will it affect tests, that the started tab will be RDI tab

vlad-dargel: thanks, I'll check this

vlad-dargel: updated

mariasergeenko (approved):

PR #3387, opened 2024-05-16
#RI-5739 - change behaviour to unblock chatbot input

romansergeenkosofteq:
#RI-5738 - fix action on login screen in chatbots


AmirAllayarovSofteq (approved):
LGTM

PR #3409, opened 2024-05-24
#RI-5619 - add action result requests

AmirAllayarovSofteq:


redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
```diff
@@ -153,4 +173,30 @@ export class ApiRdiClient extends RdiClient {
...
   }
+
+  private async pollActionStatus(actionId: string): Promise<any> {
+    const startTime = Date.now();
+    while (true) {
+      if (Date.now() - startTime > MAX_POLLING_TIME) {
+        throw new RdiPipelineTimeoutException();
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, POLLING_INTERVAL));
```
ArtemHoruzhenko, line 184: Is it better to sleep after action rather then before? what if everything was processed and ready in ms?

ArtemHoruzhenko (approved):

redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
AmirAllayarovSofteq: If everything will proceed in ms it will still work. BTW i agree that better to sleep after(requests will be proceeded really quickly)
Thanks

PR #3448, opened 2024-05-31
add click on the button

mariasergeenko:


tests/e2e/tests/web/critical-path/rdi/add-job.e2e.ts
```diff
@@ -147,7 +147,9 @@ test('Verify that user can open an additional editor to work with SQL and JMESPa
     await t.expect(await MonacoEditor.getTextFromMonaco()).eql(sqlText, 'Text from SQL editor not applied');
 
     //verify that autocomplete works for JMESPath
-    await t.pressKey('shift+space');
```
vlad-dargel, line 150: lets not delete this, but check opening sql editor by shift+space + by button

mariasergeenko: we have 2 more actions when we open the panel using ('shift+space') in the test in other rows

vlad-dargel (approved):

PR #3464, opened 2024-06-05
Extend slots refresh timeout to 5s 

WeberKuo:
https://github.com/RedisInsight/RedisInsight/issues/3429

The excessively short default slots refresh timeout, connect error occur when responses are delayed. 
Therefore, the timeout has been extended to 5s to ensure that such delays do not result in connection abnormalities.


ArtemHoruzhenko (approved):

ArtemHoruzhenko:
@WeberKuo makes sense. Thank you for contribution!

PR #3479, opened 2024-06-10
add hast field ttl test

mariasergeenko:


tests/e2e/pageObjects/browser-page.ts
```diff
@@ -399,8 +405,9 @@ export class BrowserPage extends InstancePage {
      * @param TTL The Time to live value of the key
      * @param field The field name of the key
      * @param value The value of the key
+     * @param ttl The ttl of the field for Redis databases 7.4 and higher
```
vlad-dargel, line 408: maybe it's better to call it fieldTtl or something? cause ttl is the same as TTL but not capital letters

vlad-dargel (approved):

PR #3510, opened 2024-06-17
enhance executable check

ArtemHoruzhenko:


redisinsight/api/src/utils/hosting-provider-helper.ts
```diff
@@ -97,12 +98,13 @@ export const getHostingProvider = async (client: RedisClient, databaseHost: stri
...
-      }
+      const infoObj = convertRedisInfoReplyToObject(info);
 
-      if (info.includes('executable:/opt/redis-stack/bin/redis-server')) {
-        return HostingProvider.REDIS_STACK;
+      if (infoObj.server.executable.includes('redis-server')) {
+        if (infoObj.server.executable.includes('redis-stack')) {
+          return HostingProvider.REDIS_STACK;
+        }
+        return HostingProvider.REDIS_COMMUNITY_EDITION;
```
vlad-dargel, line 107: but you still have UKNOWN_LOCALHOST?
seems like it should be removed at all, no?

vlad-dargel: and + some itests failed

ArtemHoruzhenko: > but you still have UKNOWN_LOCALHOST? seems like it should be removed at all, no?

No. It should be there

> and + some itests failed

Fixed. let's wait results

vlad-dargel (approved):

PR #3574, opened 2024-06-30
Set X-Frame-Options header to routes

GnaneshKunal:
Fixes https://github.com/RedisInsight/RedisInsight/issues/3087


ArtemHoruzhenko (approved):
Good catch!

PR #3582, opened 2024-07-01
CR-100: Added CSRF token get/set component for RI Cloud

kyle-marcum:


redisinsight/ui/src/components/csrf/Csrf.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const fetchCsrfToken = async () => {
+    let data: CSRFTokenResponse | undefined
+    try {
+      const { data } = await apiService.get<CSRFTokenResponse>(csrfEndpoint())
+
+      setCSRFHeader(data.token)
+    } catch (error) {
+      console.error('Error fetching CSRF token: ', error)
+    } finally {
+      setLoading(false)
```
ArtemHoruzhenko, line 24: In case of an error we will know that we will have errors with all the api calls. Do we need to start rendering children component?

redisinsight/ui/src/components/csrf/Csrf.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  useEffect(() => {
+    if (!csrfEndpoint()) {
+      return
+    }
+
+    fetchCsrfToken()
+  }, [])
+
+  return loading && csrfEndpoint() ? <PagePlaceholder /> : children || null
```
ArtemHoruzhenko, line 38: why do we have `|| null` here? Just asking :)

ArtemHoruzhenko (approved):

redisinsight/ui/src/components/csrf/Csrf.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { ReactElement, useEffect, useState } from 'react'
+
+import apiService, { setCSRFHeader } from 'uiSrc/services/apiService'
+import PagePlaceholder from '../page-placeholder'
+
+const csrfEndpoint = () => process.env.RI_CSRF_ENDPOINT || ''
```
zalenskiSofteq, line 6: if it's a function it should start with "get". Or you can change this to a variable

redisinsight/ui/src/components/csrf/Csrf.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+interface CSRFTokenResponse {
+  token: string;
+}
+
+const Csrf = ({ children }: { children?: ReactElement }) => {
+  const [loading, setLoading] = useState(true)
```
zalenskiSofteq, line 13: I think default value should be `false`, and just call `setLoading(true)` in the `fetchCsrfToken()`

redisinsight/ui/src/services/apiService.ts
```diff
@@ -24,9 +24,13 @@ export const getBaseUrl = () => (!isDevelopment && isWebApp
 
 const mutableAxiosInstance: AxiosInstance = axios.create({
   baseURL: hostedApiBaseUrl || getBaseUrl(),
-  withCredentials: true,
+  withCredentials: !!hostedApiBaseUrl,
 })
 
+export const setCSRFHeader = (token: string) => {
+  mutableAxiosInstance.defaults.headers.common['X-CSRF-Token'] = token
```
zalenskiSofteq, line 31: we have `CustomHeaders` enum

romansergeenkosofteq (commented):
Hello, some questions related to implementation
 - Do we need to work with this token in other components (maybe we need to store it in redux state)?
 - Do we need to get this token on each page reload (store it in LS or cookies)?

redisinsight/ui/src/components/csrf/Csrf.tsx
kyle-marcum: we will capture this work in a future card

redisinsight/ui/src/components/csrf/Csrf.tsx
kyle-marcum: `undefined` is not a valid JSX element however `null` is. I will change the signature of the component so that `children` are not optional.

redisinsight/ui/src/components/csrf/Csrf.tsx
kyle-marcum: updated

redisinsight/ui/src/components/csrf/Csrf.tsx
kyle-marcum: done

redisinsight/ui/src/services/apiService.ts
kyle-marcum: thanks for point that out. Updated

kyle-marcum:
@romansergeenkosofteq

- The token will be stored as a default header and sent with all subsequent requests since it is attached to the base axios instance
- One of the requirements was that this token should exist before any request is made in RI Cloud since the auth service requires it. Putting it as high as possible in the component hierarchy makes this possible.

ArtemHoruzhenko (approved):

PR #3611, opened 2024-07-15
Add provider to all frontend telemetry events #5675

kchepikava:


redisinsight/ui/src/telemetry/telemetryUtils.ts
```diff
@@ -19,6 +20,8 @@ import {
 import { TelemetryEvent } from './events'
 import { checkIsAnalyticsGranted } from './checkAnalytics'
 
+export const getProvider = (): string => get(store.getState(), 'connections.instances.connectedInstance.provider', '')
```
ArtemHoruzhenko, line 23: it is better to have `undefined` instead of `''` (empty string). In this case provider field won't be sent at all

ArtemHoruzhenko (changes requested):

redisinsight/ui/src/telemetry/telemetryUtils.ts
kchepikava: @ArtemHoruzhenko, fixed to be undefined by default

redisinsight/ui/src/telemetry/telemetryUtils.ts
```diff
@@ -19,6 +20,8 @@ import {
 import { TelemetryEvent } from './events'
 import { checkIsAnalyticsGranted } from './checkAnalytics'
 
+export const getProvider = (): string | undefined => get(store.getState(), 'connections.instances.connectedInstance.provider', undefined)
```
zalenskiSofteq, line 23: instead of `T | undefined`, or `T | null` we use `Maybe<T>` and `Nullable<T>`

ArtemHoruzhenko: `get(store.getState(), 'connections.instances.connectedInstance.provider')` (without 3d argument) will work the same, no?

zalenskiSofteq (approved):

ArtemHoruzhenko (dismissed):
lgtm

PR #3637, opened 2024-07-24
RI-5906 removed double fetch call of rdi instance data

kchepikava:
RI-5906


romansergeenkosofteq (approved):

romansergeenkosofteq (changes requested):
please, fix tests

PR #3678, opened 2024-08-06
enhance sso errors

ArtemHoruzhenko:


kchepikava (dismissed):

zalenskiSofteq (approved):
Approve. Only 1 syntax sugar:
Instead of `indexOf(...) > -1` we can use `.includes(...)`

PR #3688, opened 2024-08-08
Feature/CR-120-pubsub

nelsonwellsredis:


nelsonwellsredis:
@ArtemHoruzhenko @zalenskiSofteq @romansergeenkosofteq 

Hi guys, for some context, these changes are to support defaulting to the socket.io polling mechanism for the websocket connections we create. There are some issues with using websockets in our hosted environment and I have been able to prove that it can work with polling instead. My goal was to make the smallest changes without impacting the rest of the app.

redisinsight/api/src/modules/bulk-actions/bulk-actions.gateway.ts
```diff
@@ -27,7 +27,8 @@ const SOCKETS_CONFIG = config.get('sockets');
 @WebSocketGateway({
   path: SOCKETS_CONFIG.path,
   namespace: 'bulk-actions',
-  cors: SOCKETS_CONFIG.cors,
+  cors: SOCKETS_CONFIG.cors
```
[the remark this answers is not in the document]
nelsonwellsredis: Fixed

redisinsight/api/src/modules/profiler/profiler.gateway.ts
```diff
@@ -22,8 +22,8 @@ const SOCKETS_CONFIG = config.get('sockets') as Config['sockets'];
 @WebSocketGateway({
   path: SOCKETS_CONFIG.path,
   namespace: 'monitor',
-  cors: SOCKETS_CONFIG.cors,
-  serveClient: SOCKETS_CONFIG.serveClient,
```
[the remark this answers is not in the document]
nelsonwellsredis: It was an oversight, this is fixed now.

ArtemHoruzhenko (approved):
lgtm

romansergeenkosofteq:
I would suggest in redisinsight/ui/src/components/csrf/Csrf.tsx move logic of fetching and storing token to the store actions, and store token in the store (with loading and token)
it will be like

```
const { loading , token } = useSelector(cstfTokenSelector)
```

```
dispatch(fetchCsrfToken())
```

in this case we dont need to read token in other components from LS, just from store

redisinsight/ui/src/slices/app/csrf.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+})
+
+export const { setLoading, setToken } = appCsrfSlice.actions
+
+export const appCsrfSelector = (state: RootState) => state.app.csrf
+
+export default appCsrfSlice.reducer
+
+export function fetchCsrfToken(
```
romansergeenkosofteq, line 37: I suggest rename here to fetchCstfTokenAction and next have 3 dispatches (fetchCsrtToken, fetchCsrtTokenSuccess, fetchCsrtTokenFail)

redisinsight/ui/src/slices/app/csrf.ts
```diff
@@ -0,0 +1,58 @@
...
+) {
+  return async (dispatch: AppDispatch) => {
+    try {
+      if (getCsrfEndpoint()) {
+        const { data } = await apiService.get<CSRFTokenResponse>(getCsrfEndpoint())
+
+        dispatch(setToken({ token: data.token }))
+        setApiCsrfHeader(data.token)
+        setResourceCsrfHeader(data.token)
+        onSuccessAction && onSuccessAction(data)
```
romansergeenkosofteq, line 49: its enough onSuccessAction?.(data)

redisinsight/ui/src/slices/app/csrf.ts
```diff
@@ -0,0 +1,58 @@
...
+      }
+    } catch (error) {
+      console.error('Error fetching CSRF token: ', error)
+      onFailAction && onFailAction()
```
romansergeenkosofteq, line 53: its enough onFailAction?.()

redisinsight/ui/src/slices/app/csrf.ts
```diff
@@ -0,0 +1,58 @@
...
+const getCsrfEndpoint = () => process.env.RI_CSRF_ENDPOINT || ''
+
+interface CSRFTokenResponse {
+  token: string;
+}
+
+export const initialState: {
+  csrfEndpoint: string;
+  loading: boolean;
+  token: string;
```
romansergeenkosofteq, line 15: we do not use semicolons in the interfaces

redisinsight/ui/src/components/monitor-config/MonitorConfig.tsx
```diff
@@ -31,6 +32,7 @@ interface IProps {
 const MonitorConfig = ({ retryDelay = 15000 } : IProps) => {
   const { id: instanceId = '' } = useSelector(connectedInstanceSelector)
   const { socket, isRunning, isPaused, isSaveToFile, isMinimizedMonitor, isShowMonitor } = useSelector(monitorSelector)
+  const { token } = useSelector(appCsrfSelector);
```
romansergeenkosofteq, line 35: semicolon

redisinsight/ui/src/slices/app/csrf.ts
```diff
@@ -0,0 +1,58 @@
...
+} = {
+  csrfEndpoint: getCsrfEndpoint(),
+  loading: true,
```
romansergeenkosofteq, line 18: I think we should have loading: false by default, and set it to true when we fetching.
and then we can use `(!loading || token) ? children : <PagePlaceholder />` 

but it this case need to think what should be on refetch token (just replace token without deleting?)

nelsonwellsredis: I've updated the initial state so that loading = false, but the logic had to be adjusted somewhat. If the logic was just as you've written, children would be rendered on the first render of the csrf component. The csrf component is made to prevent that from happening because rendering the children will trigger additional web requests, and in a cloud environment, those web requests will fail if the csrf token is not available. So, near the app root level, we effectively block everything from rendering until we have the csrf token, then we can safely render the rest of the app.

redisinsight/ui/src/components/csrf/Csrf.spec.tsx
```diff
@@ -1,83 +1,86 @@
...
+import { render, screen, initialStateDefault } from 'uiSrc/utils/test-utils'
+import Csrf from 'uiSrc/components/csrf/Csrf'
 
-jest.mock('uiSrc/services/apiService', () => ({
-  setApiCsrfHeader: jest.fn(),
-  get: jest.fn(() => ({ data: { token: 'csrf-token' } })),
+jest.mock('react-redux', () => ({
+  useDispatch: jest.fn(),
+  useSelector: jest.fn(),
 }))
```
romansergeenkosofteq, line 10: could you please mock 'uiSrc/slices/app/csrf', and check which actions we call.
You can find examples how we do this in other components

romansergeenkosofteq (approved):
lgtm

PR #3694, opened 2024-08-12
RI 5979 show detailed message instead of common unauthorized

kchepikava:


redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
```diff
@@ -62,7 +62,8 @@ export class ApiRdiClient extends RdiClient {
         jobs,
       };
     } catch (e) {
-      throw wrapRdiPipelineError(e);
+      const message: string = e.response?.data?.detail || e.message || '';
```
zalenskiSofteq, line 65: I think instead of repeat this code several times better move this logic inside `wrapRdiPipelineError`

zalenskiSofteq: and the last value should be `undefined` for all default values messages inside the `RdiPipeline.....Exception(.....)`

zalenskiSofteq (approved):

mariasergeenko (approved):

PR #3712, opened 2024-08-16
CR-114: Added environment specific config file to ui so bundle can be…

kyle-marcum:
… used across different environments


ArtemHoruzhenko (approved):
looks good

PR #3759, opened 2024-08-27
Feature/fe/cr 138 hide onboarding GitHub feedback

pd-redis:


nelsonwellsredis:
@pd-redis - looks like you missed the "Let us know what you think" link in the bottom right corner. Check out the `BottomGroupMinimized.tsx` component

pd-redis:
> @pd-redis - looks like you missed the "Let us know what you think" link in the bottom right corner. Check out the `BottomGroupMinimized.tsx` component

Good catch, handled

nelsonwellsredis (approved):

redisinsight/ui/src/components/navigation-menu/NavigationMenu.tsx
```diff
@@ -315,33 +315,47 @@ const NavigationMenu = () => {
...
-          position="right"
+        <FeatureFlagComponent
+          name={FeatureFlags.disabledByEnv}
+          enabledByDefault
+          otherwise={(
+            <Divider
+              color="transparent"
+              className="eui-hideFor--xs eui-hideFor--s"
+              variant="middle"
+            />
```
romansergeenkosofteq, line 326: Do we need this transparent Divider?

pd-redis: Kind of, if we don't have a divider at all, the sidebar jumps when you hover over the Help button, shows vertical and horizontal scrollbars.
If it is not transparent, then we get a divider that has nothing after it.
Maybe more elegant solution is possible but have to dig into how to disable the hover effect for the help button

romansergeenkosofteq (approved):

PR #3870, opened 2024-09-30
Use proxy path prefix for theme to make it work for multiple subpath instances

GnaneshKunal:
Fix: https://github.com/RedisInsight/RedisInsight/issues/3282

Use proxy path prefix in the theme keyname to respect multiple subpath instances running simultaneously.

### DIFF
--- a/redisinsight/ui/src/services/theme.ts
+++ b/redisinsight/ui/src/services/theme.ts
@@ -7,7 +7,7 @@ interface Themes {
 
 class ThemeService {
   readonly themes: Themes = {}
-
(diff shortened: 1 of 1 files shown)


ArtemHoruzhenko:
In my opinion we must think about general solution. if I'm not mistaken RI might share local storage with another app when proxy path used. Let's design and implement solution 
1. where all our data in local storage will be placed under single property (e.g. `ri`). cons: pretty tricky to make it backward compatible
2. OR save ALL local storage data under unique property based on ri proxy where by default data will be stored in the root. In this scenario we might ignore backward compatibility since very low number of users will be affected
@GnaneshKunal @zalenskiSofteq @romansergeenkosofteq @kchepikava  @ViktarStarastsenka wdyt?

ArtemHoruzhenko:
was done in this PR https://github.com/RedisInsight/RedisInsight/pull/4081

PR #3898, opened 2024-10-04
CR-13: Disable certain routes for cloud

seth-riedel:


redisinsight/ui/src/components/main-router/config.ts
```diff
@@ -0,0 +1,14 @@
+import { isArray } from 'lodash'
+import { IRoute } from 'uiSrc/constants'
+
+export const LAZY_LOAD = window.riConfig?.app?.lazyLoad
```
seth-riedel, line 4: @ArtemHoruzhenko Is this the right way to use the new config?

redisinsight/ui/src/components/main-router/config.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const RI_ROUTES_EXCLUDED_BY_ENV = window.riConfig?.app?.routesExcludedByEnv
+
+export const getRouteIncludedByEnv = (routeDefinition: IRoute | IRoute[]) => {
```
zalenskiSofteq, line 8: let's always have an array
just replace in defaultRoutes
```getRouteIncludedByEnv({....})```
to
```getRouteIncludedByEnv([{....}])```

File: redisinsight/ui/src/components/main-router/config.ts [hunk 3]
```diff
@@ -0,0 +1,14 @@
...
+  if (RI_ROUTES_EXCLUDED_BY_ENV) {
```
zalenskiSofteq, line 9: did you check that the build is smaller with this env variable?

redisinsight/ui/src/components/main-router/config.ts
```diff
@@ -0,0 +1,14 @@
+import { isArray } from 'lodash'
+import { IRoute } from 'uiSrc/constants'
+
+export const LAZY_LOAD = window.riConfig?.app?.lazyLoad
+
+export const RI_ROUTES_EXCLUDED_BY_ENV = window.riConfig?.app?.routesExcludedByEnv
```
zalenskiSofteq, line 6: remove `RI_` please

redisinsight/ui/src/components/main-router/config.ts
zalenskiSofteq: yes

redisinsight/ui/src/components/main-router/config.ts
seth-riedel: The build size doesn't seem to be changing regardless of whether or not this variable is set

zalenskiSofteq (approved):

PR #3908, opened 2024-10-08
E2e/feature/ri 6160 add a list data type with multiple elements

mariasergeenko:


tests/e2e/tests/web/regression/browser/list-key.e2e.ts
```diff
@@ -12,34 +15,90 @@ const databaseHelper = new DatabaseHelper();
...
+        await browserPage.addElementToList([elementForSearch]);
         // Search by element index
         await browserPage.searchByTheValueInKeyDetails('1000001');
         // Check the search result
         const result = await browserPage.listElementsList.nth(0).textContent;
         await t.expect(result).eql(elementForSearch, 'List element not found');
     });
+
+// TODO unskip when the bug is fixed https://redislabs.atlassian.net/browse/RI-6188
+test.skip
```
vlad-dargel, line 50: no need to skip this test, because issue will be fixed in scope of this story in its branch
in order to not forget to unskip it

vlad-dargel (approved):

PR #3977, opened 2024-10-18
CR-173: Updated favicon to match SM UI

seth-riedel:


ArtemHoruzhenko:
@zalenskiSofteq Could have a look please?

PR #3985, opened 2024-10-21
fix: ensure BigInt values show correctly in JSON view

lumyjuwon:
As you know, Node.js doesn't support scalar over 2^53 defaultly but support in Redis. So when receive scalar over 2^53 then loses precision in JSON.

"MailId" is Bigint but BigInt loses precision in JSON view like follow as.  

Group 41

So I added handler about BigInt. BigInt(MailId) doesn't lose precision follow as.  

Group 40


KIvanow:
Thank you for your contribution! I believe we've fixed this in last month's release and will further improve it in the release coming next. If there are any subsequent issues, please, let us know.

PR #3986, opened 2024-10-21
CR-167: Validate browser history mode from request

seth-riedel:


ArtemHoruzhenko:
Let's do it in a common RI way

ArtemHoruzhenko (changes requested):
Let's do it in a common RI way https://github.com/RedisInsight/RedisInsight/pull/3998

PR #3990, opened 2024-10-22
POC/Impl: Connect to Redis behind an MFA enabled SSH

GnaneshKunal:
Addresses https://github.com/RedisInsight/RedisInsight/issues/3214

### DIFF
--- a/redisinsight/api/src/modules/ssh/ssh-tunnel.provider.ts
+++ b/redisinsight/api/src/modules/ssh/ssh-tunnel.provider.ts
@@ -8,24 +8,56 @@ import { SshOptions } from 'src/modules/ssh/models/ssh-options';
 
 @Injectable()
 export class SshTunnelProvider {
+  private activeConnections: Map<string, { client: Client; server: Server | null }> = new Map();
+
+  private getConnectionKey(target: Endpoint, sshOptions: SshOptions): string {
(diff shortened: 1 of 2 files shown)


redisinsight/api/src/modules/ssh/ssh-tunnel.provider.ts
```diff
@@ -1,38 +1,176 @@
...
+      const connectionKey = this.getConnectionKey(target, sshOptions);
+      
+      // Cleanup any existing connection
+      this.cleanupConnection(connectionKey);
+
+      const client = new Client();
+      let server: Server | null = null;
+
+      client.on('keyboard-interactive', (name, instructions, instructionsLang, prompts, finish) => {
+        finish(['032100']);
```
KIvanow, line 50: @GnaneshKunal What is this string representing? Short google search shows it can be used as a sort of a 'hack' for dealing with some authenticators.

GnaneshKunal: @KIvanow, this PR is more of a POC and approach. 
Ideally, we shouldy implement a UI form elementthat takese in the TOTP andpassess it here in the `finish` function.

KIvanow: got it, thanks!

KIvanow:
Outdated

PR #4017, opened 2024-10-26
Ri 5876 fix start electron dev mode

KIvanow:
Steps to start (from the root folder each in a different browser tab):
`yarn dev:electron:ui`
`yarn dev:electron:api`
`yarn dev:electron`

This will start the fe and be with the required environments for electron as well as the electron app.

Live reload works both for the ui and api.

There is a new added script - `dev:api` for the sake of consistency. Right now we had to run (based on the wiki)` yarn dev:ui `and `yarn --cwd redisinsight/api/ start:dev` which seems difficult to remember and inconsistent.

Running just `yarn dev:ui` and yarn `dev:api` in separate terminals (or terminal tabs) will still allow the project to be ran in a browser (no loss of functionality).

All of the changes are aimed for development environment, production builds should be unaffected (tested on local linux builds)

Major changes

1. Added vite configs for the desktop folder to keep it consistent
2. Decoupled the desktop from loading parts of the api to handle window authentication and cloud authentication. On dev now those are done through TCP servers. This is what allows for everything to be ran separately and still have live/hot reload. This is also going to be used only during development

    2.1   All windows have the window id header in the requests and are checked in the api through the TCP communication approach
    
    2.2 The renderer is able to handle click on a social login button, then the main process is able to ping the api, the api is able to generate a URL, pass it back, open it in the default browser, complete the flow and then have the protocol of redisinsight:// be opened as a result. Everything going through the TCP communication instead of the desktop directly invoking methods of the api.
    
4. Added 4 different packages that I've had yarn complain are not installed but required on 2 different OSs

I've tried to keep the changes as minimal as possible. 

Any and all feedback is welcomed (especially if I've missed another place the desktop relied on the api in such a way).


redisinsight/desktop/forge.config.js
```diff
@@ -0,0 +1,61 @@
+const { FusesPlugin } = require('@electron-forge/plugin-fuses');
```
zalenskiSofteq, line 1: Is this file used? I can't find a link to it anywhere.

redisinsight/desktop/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "version": "1.0.0",
+  "main": "dist/index.js",
+  "scripts": {
+    "dev": "cross-env ELECTRON_DEV=true ELECTRON_ENABLE_LOGGING=true ELECTRON_DEBUG_LOGGING=true ELECTRON_ENABLE_STACK_DUMPING=true yarn build:preload && yarn build:renderer && yarn build && electron . --enable-logging --inspect=5858",
+    "build": "vite build --config vite.main.config.ts",
+    "build:preload": "vite build --config vite.preload.config.ts",
+    "build:renderer": "vite build --config vite.renderer.config.ts"
+  },
+  "devDependencies": {
+    "@vitejs/plugin-react": "^4.3.3",
```
zalenskiSofteq, line 12: We should not have duplicates of dependendcies
Please remove all dependencies from this file. They should be taken from root package.json. 
Or move dependencies from root to here.

package.json
```diff
@@ -211,6 +218,9 @@
     "yarn-deduplicate": "^3.1.0"
   },
   "dependencies": {
+    "@antv/hierarchy": "^0.6.14",
```
zalenskiSofteq, line 221: you don't need add dependencies.
Vite optimise warning appears if dependencies from the packages folder (redisinsight/ui/src/packages) have not been installed.
or you can exclude them in the ui vite.config.mjs
```
optimizeDeps: {
    exclude: [
      'react-json-tree',
      'redisinsight-plugin-sdk',
      'plotly.js-dist-min',
      '@antv/x6',
      '@antv/x6-react-shape',
      '@antv/hierarchy'
    ],
```

vite.main.config.ts
```diff
@@ -0,0 +1,53 @@
+import { defineConfig } from 'vite';
```
zalenskiSofteq, line 1: Is this file used?

redisinsight/desktop/vite.main.config.ts
```diff
@@ -0,0 +1,57 @@
...
+        ...builtinModules.map(m => `node:${m}`),
+        /^apiSrc\//,
+        /^src\//,
+      ],
+      output: {
+        format: 'cjs',
+        entryFileNames: '[name].js',
+      }
+    },
+    target: 'node14',
```
zalenskiSofteq, line 46: node14?

zalenskiSofteq (commented):
Please clear this pull request. There are unused files, console.logs, if(.... true)
1. There are changes for BE code, please rename branch to 'be/feature/...' to auto run BE tests
2. check your eslint.  For desktop folder should apply root .eslintrc.js. This code has a lot of eslint errors: import orders, semicolon etc
Backend has own eslint with semicolon.

redisinsight/desktop/forge.config.js
KIvanow: You are right. I've missed to clean it up after when I started removing all different experiments. Removed.

package.json
KIvanow: I am a bit confused, why shouldn't we add them as dependencies, if the warning is caused by them not being installed?

zalenskiSofteq: They aren't used in the RedisInsight application. 
They are a part of plugins (redisinsight/ui/src/packages). Each plugin is built separately from RedisInsight. Plugins dependencies are not used in the Redis Insight. Vite optimization doesn't see the difference and shows useless warning

zalenskiSofteq: there is documentation about Plugins
https://github.com/RedisInsight/RedisInsight/wiki/Plugin-Documentation

redisinsight/desktop/package.json
KIvanow: cleared the entire dependencies for the desktop. Tested to make sure it will still work.

package.json
KIvanow: got it! THank you for the explanation. Cleared those as well.

vite.main.config.ts
KIvanow: Yes, in the corresponding package.json

zalenskiSofteq:
> @zalenskiSofteq your feedback has been addressed, please take another look.
> 
> About the lining, there are multiple lint errors for API and UI from before this PR. When are those typically enforced? Should we set up the husky hooks for that?

To set up husky we need to fix all the errors, it's gonna take time. 
I have auto format after save a file, it fixes almost all the errors

PR #4065, opened 2024-11-06
CR-195 add some ui tests for AppInit

pd-redis:


redisinsight/ui/src/components/init/AppInit.tsx
```diff
@@ -1,16 +1,40 @@
...
+    if (status === STATUS_INITIAL) {
+      initApp()
+    }
   }, [])
 
+  if (status === STATUS_FAIL) {
+    return <ConnectivityError isLoading={false} onRetry={initApp} error={error} />
+  }
+
   return status === STATUS_SUCCESS ? children : <PagePlaceholder />
```
seth-riedel, line 38: Instead of `<PagePlaceholder />`, I think we need to render `<SuspenseLoader />`. Currently, it just renders a black screen while initializing. It'd be a better user experience to show a loader while we're fetching data in case it takes a few seconds.

redisinsight/ui/src/components/init/AppInit.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+        <div>children</div>
+      </AppInit>, {
+        store: mockedStore,
+      }
+    )
+    expect(screen.getByTestId('connectivity-error-message')).toBeInTheDocument()
+    expect(screen.getByText('Something went wrong, please try again later')).toBeInTheDocument()
+  })
+
+  it('should render error page with custom error message, when status is "fail"', () => {
```
seth-riedel, line 83: I think we need some unit tests that don't mock out the functionality of the appinit slice. There is currently a bug where even if the network requests fail, it doesn't dispatch the `initializeAppStateFail` action like it should. This is something a unit test could cover using mock service worker to simulate a failed API request. Right now these tests aren't really testing the store because we've mocked out all the actions/selectors/etc.  What do you think?

seth-riedel (changes requested):

redisinsight/ui/src/components/init/AppInit.tsx
```diff
@@ -1,16 +1,40 @@
...
+  const {
+    status,
+    error = 'Something went wrong, please try again later',
+  } = useSelector(appInitSelector)
+
+  const initApp = useCallback(() => dispatch(initializeAppAction(onSuccess, onFail)), [])
 
   useEffect(() => {
-    dispatch(initializeAppAction())
+    if (status === STATUS_INITIAL) {
```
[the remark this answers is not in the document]
pd-redis: I had to add this because of the tests, they would go into a loop without it. Not sure why

redisinsight/ui/src/components/init/AppInit.tsx
pd-redis: I agree about the loader, current logic was simply following the Crsf component's logic

seth-riedel: Can we go ahead and change? It's an easy change and it's better than showing a black screen to the user.

redisinsight/ui/src/components/init/AppInit.tsx
```diff
@@ -1,16 +1,40 @@
...
+      initApp()
+    }
   }, [])
 
+  if (status === STATUS_FAIL) {
+    return <ConnectivityError isLoading={false} onRetry={initApp} error={error} />
```
seth-riedel, line 35: I don't think we should be showing the actual error message to the user. Right now it will show something like "Failed to fetch feature flags". I don't think this is appropriate for end users. We can probably just show a generic message, something like "An unexpected server error has occurred. Please retry the request."

redisinsight/ui/src/components/init/AppInit.tsx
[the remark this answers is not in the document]
seth-riedel: If I remove this, the error I'm getting is about a Redux action:
image

It's complaining about this dispatch, because something isn't mocked or setup correctly for the test:
`dispatch(initializeAppAction(onSuccess, onFail))`

I think we're still missing some critical tests for this component. We are testing the component in different states with different props, but there are no tests that actually test the interaction of the component with the Redux store. I still think we need to remove this `if` condition, and add more tests that use mock service worker to handle API calls, and use a mocked store so the component can actually dispatch the actions and re-render to the correct states based on those interactions. Otherwise, we don't have any test coverage for how this component behaves in the real world when connected to Redux and actually makes API requests. What are your thoughts?

seth-riedel (approved):

PR #4086, opened 2024-11-13
#RI-6320 - add feature flag for cloud db in the list of dbs

romansergeenkosofteq:


ViktarStarastsenka (approved):

ArtemHoruzhenko (approved):
lgtm
btw @ViktarStarastsenka is it better to use different naming for feature flag? 
e.g. `enhancedCloudUI` or something similar... seems like more generic

ArtemHoruzhenko (approved):

PR #4120, opened 2024-11-20
CR-215 workbench local storage history

pd-redis:
In this PR, I am checking if envDependent flag is off, and if it is, do not perform API request to list/get/delete Workbench commands.
Also, if the flag is off, I am storing results of `sendWBCommandAction` and `sendWBCommandClusterAction` in localStorage, from where they can be loaded after


romansergeenkosofteq (commented):
IMO you should not use localStorage to save history of workbench, better to use IndexedDB API. LocalStorage has limited size of memory, also we limit workbench the size of history results (only 20 results per database)

romansergeenkosofteq (commented):
Please, also be sure that users can work with different databases and see results per database

pd-redis:
> IMO you should not use localStorage to save history of workbench, better to use IndexedDB API. LocalStorage has limited size of memory, also we limit the size of history results (only 20 results per database)

I know that localStorage is limited, but still fas few MB size (about 5). Do you think it is reasonable to expect that 20 commands can exceed that?

romansergeenkosofteq:
> > IMO you should not use localStorage to save history of workbench, better to use IndexedDB API. LocalStorage has limited size of memory, also we limit the size of history results (only 20 results per database)
> 
> I know that localStorage is limited, but still fas few MB size (about 5). Do you think it is reasonable to expect that 20 commands can exceed that?

yes, even 1 command result can exceed localStorage size. Users may have several databases, in this case they will have (number of databases * 20) results. Also 1 command can be in group mode and contain several results of commands.

pd-redis:
> Please, also be sure that users can work with different databases and see results per database

This feature should be used in cloud RI only. I don't think different databases

pd-redis:
> > > IMO you should not use localStorage to save history of workbench, better to use IndexedDB API. LocalStorage has limited size of memory, also we limit the size of history results (only 20 results per database)
> > 
> > 
> > I know that localStorage is limited, but still fas few MB size (about 5). Do you think it is reasonable to expect that 20 commands can exceed that?
> 
> yes, even 1 command result can exceed localStorage size. Users may have several databases, in this case they will have (number of databases * 20) results. Also 1 command can be in group mode and contain several results of commands.

OK, I will look into it

pd-redis:
> > > IMO you should not use localStorage to save history of workbench, better to use IndexedDB API. LocalStorage has limited size of memory, also we limit the size of history results (only 20 results per database)
> > 
> > 
> > I know that localStorage is limited, but still fas few MB size (about 5). Do you think it is reasonable to expect that 20 commands can exceed that?
> 
> yes, even 1 command result can exceed localStorage size. Users may have several databases, in this case they will have (number of databases * 20) results. Also 1 command can be in group mode and contain several results of commands.

Implemented indexedDb storing

pd-redis:
> I think its better to move class IndexedDbStorage to a different file. Also, I would recommend to have a file with instances of different tables (`export const wbHistoryStorage = new IndexedDbStorage('RI_WB_HISTORY')`)
Sure, moving it to separate file makse sense. I don't think we'll have other databases besides wb history one, at least for now. Which is why, I didn't bring external library to manage this.
> 
> IMO it's better to have a separate record for each command execution (like implemented on BE side) than modify each time one.
You mean like, every response from POST /workbench/command-executions to be separate record?
> 
> Will limit of records for 1 database be implemented?

yeah, this is stil Work In Progress

romansergeenkosofteq:
> You mean like, every response from POST /workbench/command-executions to be separate record?
yes, I think you can try to use indexes https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/index

romansergeenkosofteq:
@pd-redis What will be if we need to add one more table/storage?

I think it will be better to use the library (https://dexie.org/) to be more flexible and stable. 
Also, I think we should discuss how to work with local database and sql (maybe we can implement one solution for all build types?) 
@ArtemHoruzhenko @zalenskiSofteq - what do you think?

redisinsight/ui/src/services/workbenchStorage.ts
```diff
@@ -0,0 +1,350 @@
...
+export class WorkbenchStorage {
+  private db?: IDBDatabase
+
+  constructor(private readonly dbName: string, private readonly version = 1, private readonly keySeparator = ':') {
+  }
+
+  private initDb(storeName: string) {
+    return new Promise((resolve, reject) => {
+      if (!window.indexedDB) {
+        reject(new Error('indexedDB is not supported'))
```
seth-riedel, line 19: I believe you need a return after this, otherwise script execution will continue

redisinsight/ui/src/services/workbenchStorage.ts
```diff
@@ -0,0 +1,350 @@
...
+export async function addCommands(dbId: string, commandId: string, data: CommandExecution[]) {
+  // Store command results in local storage!
+  const commandsHistory: CommandHistoryType = await getLocalWbHistory(dbId)
+  const storedData = data.map((item) => {
+    // Do not store command execution result that exceeded limitation
+    if (JSON.stringify(item.result).length > riConfig.workbench.maxResultSize) {
+      item.result = [
+        {
+          status: CommandExecutionStatus.Success,
+          response: `Results have been deleted since they exceed ${formatBytes(riConfig.workbench.maxResultSize)}.
```
seth-riedel, line 295: Looks like there's an error with this when the result is a table:
image

It's still trying to render the table instead of showing this error message

redisinsight/ui/src/services/workbenchStorage.ts
```diff
@@ -0,0 +1,350 @@
...
+      commandId,
+      data: reverse(storedData),
+    } as CommandHistoryItem
+  ])
+
+  return cleanupDatabaseHistory(dbId)
+}
+
+export async function removeCommand(dbId: string, commandId: string) {
+  // Delete command from local storage?!
```
seth-riedel, line 315: After deleting a workbench item, it doesn't get removed from local storage, so it re-appears after navigating away from the page and coming back. It seems like we should delete it

seth-riedel (changes requested):
Please see comments. I think we also need a lot of tests for this, both for the changes to the workbench slice as well as the behavior of `services/workbenchStorage.ts`

pd-redis:
> @pd-redis What will be if we need to add one more table/storage?
> 
> I think it will be better to use the library (https://dexie.org/) to be more flexible and stable. Also, I think we should discuss how to work with local database and sql (maybe we can implement one solution for all build types?) @ArtemHoruzhenko @zalenskiSofteq - what do you think?

Reg dexie - we discussed it, but the consensus was, that bringing 90K dependency just to store WB history is not worth it.
Otherwise, I have used it before and I like it.

redisinsight/ui/src/services/workbenchStorage.ts
pd-redis: that appears to be bug in redisearch plugin

ArtemHoruzhenko (approved):

PR #4136, opened 2024-11-22
CR-264: Session idle timeout

seth-riedel:


ArtemHoruzhenko (commented):
Do we want test lib to be a part of production? Does it affect performance somehow?

seth-riedel:
> Do we want test lib to be a part of production? Does it affect performance somehow?

Oops, no. I changed it

ArtemHoruzhenko (approved):

PR #4141, opened 2024-11-26
RI-6357 fix app start telemetry

kchepikava:


redisinsight/api/src/main.ts
```diff
@@ -77,7 +82,7 @@ export default async function bootstrap(apiPort?: number): Promise<IApp> {
 
   await app.listen(apiPort || port, host);
   logger.log({
-    message: `Server is running on http(s)://${host}:${port}`,
+    message: `Server is running on http(s)://${host}:${apiPort || port}`,
```
ArtemHoruzhenko, line 85: you can use just `serverConfig.port` here. There is no need for additional checks

redisinsight/api/src/main.ts
```diff
@@ -30,6 +30,11 @@ export default async function bootstrap(apiPort?: number): Promise<IApp> {
   }
 
   const { port, host } = serverConfig;
+
+  if (apiPort && apiPort !== port) {
```
ArtemHoruzhenko, line 34: Let's be as simple as possible. Not sure that we must add additional check `apiPort !== port` here to avoid misunderstandings.

ArtemHoruzhenko (changes requested):

ArtemHoruzhenko (approved):
lgtm

PR #4161, opened 2024-12-02
CR-189 add configuration to trigger recommendation fetch on set interval

pd-redis:
default 60 seconds

In case that WS connection does not work in cloud


ArtemHoruzhenko (approved):
lgtm

ArtemHoruzhenko (approved):

PR #4166, opened 2024-12-04
CR-220 do not reference pub sub

pd-redis:
re-posting because previous PR got messed up soemhow

### DIFF
--- a/redisinsight/ui/src/components/query/query-card/QueryCardCommonResult/components/CommonErrorResponse/CommonErrorResponse.tsx
+++ b/redisinsight/ui/src/components/query/query-card/QueryCardCommonResult/components/CommonErrorResponse/CommonErrorResponse.tsx
@@ -18,6 +18,7 @@ import { connectedInstanceSelector } from 'uiSrc/slices/instances/instances'
 
 import { showMonitor } from 'uiSrc/slices/cli/monitor'
 import UseProfilerLink from 'uiSrc/components/monitor/UseProfilerLink'
+import UsePubSubLink from 'uiSrc/components/pub-sub/UsePubSubLink/UsePubSubLink';
 
 const CommonErrorResponse = (id: string, command = '', result?: any) => {
   const { instanceId = '' } = useParams<{ instanceId: string }>()
@@ -33,7 +34,7 @@ const CommonErrorResponse = (id: string, command = '', result?: any) => {
   }
   // Flow if SUBSCRIBE command was executed
   if (checkUnsupportedCommand([CommandSubscribe.toLowerCase()], commandLine)) {
-    return cliTexts.SUBSCRIBE_COMMAND(Pages.pubSub(instanceId))
+    return <UsePubSubLink path={Pages.pubSub(instanceId)} />
   }
   // Flow if PSUBSCRIBE command was executed
   if (checkUnsupportedCommand([CommandPSubscribe.toLowerCase()], commandLine)) {
--- a/redisinsight/ui/src/components/query/query-card/QueryCardCommonResult/components/CommonErrorResponse/CommonErrorResponse.tsx
+++ b/redisinsight/ui/src/components/query/query-card/QueryCardCommonResult/components/CommonErrorResponse/CommonErrorResponse.tsx
@@ -18,7 +18,7 @@ import { connectedInstanceSelector } from 'uiSrc/slices/instances/instances'
 
 import { showMonitor } from 'uiSrc/slices/cli/monitor'
 import UseProfilerLink from 'uiSrc/components/monitor/UseProfilerLink'
-import UsePubSubLink from 'uiSrc/components/pub-sub/UsePubSubLink'
+import UsePubSubLink from 'uiSrc/components/pub-sub/UsePubSubLink/UsePubSubLink';
 
 const CommonErrorResponse = (id: string, command = '', result?: any) => {
   const { instanceId = '' } = useParams<{ instanceId: string }>()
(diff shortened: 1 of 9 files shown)


ArtemHoruzhenko (approved):
lgtm

redisinsight/ui/src/components/query/query-card/QueryCardCommonResult/components/CommonErrorResponse/CommonErrorResponse.tsx
```diff
@@ -33,7 +34,7 @@ const CommonErrorResponse = (id: string, command = '', result?: any) => {
   }
   // Flow if SUBSCRIBE command was executed
   if (checkUnsupportedCommand([CommandSubscribe.toLowerCase()], commandLine)) {
-    return cliTexts.SUBSCRIBE_COMMAND(Pages.pubSub(instanceId))
+    return <UsePubSubLink path={Pages.pubSub(instanceId)} />
```
romansergeenkosofteq, line 37: - by naming convention in React only hooks should start with prefix use
- do we need this component at all? Can we use FeatureFlagComponent with otherwise prop?

pd-redis: In my opinion, all output should be done via components, also we already have `UseProfilerLink` component, which is equivalent to this one. 
Use convention is for function calls, not components, but provided that we decide to change all similar components to different convention, I am OK with this as well

romansergeenkosofteq: you mean output of unsupported commands? There is an object of small functions or texts, why do we need to create components for each?

pd-redis: @zalenskiSofteq why not? This is supposed to be declarative react app. Otherwise, why not just use html strings?
Having the display logic related to specific text encapsulated in component is the best way. If you have 100 components like this, then probably there is something wrong with the architecture

zalenskiSofteq: @pd-redis 
I don’t understand, please explain more clearly. Do you think creating 100 components for each comand is good or bad?

I wrote:

> “We shouldn’t create 100 components.”

You replied:

> “Why not?”

 — this means you think it’s good.

But then you wrote, 

> there is something wrong with the architecture.

pd-redis: If it makes sense for the use cases, yes I think it is good. In this case, there is some logic, that needs to be performed, if we apply it inline for every potential case, that would make CommonErrorResponse.tsx file even harder to parse. Also extra benefit of separate component is ability to test it in isolation, which is not present with CommonErrorResponse approach.

zalenskiSofteq: To test it in isolation and apply the DRY principle, it's better to create a single generic component for such commands.

romansergeenkosofteq (commented):
Why this PR was merged without approve?

pd-redis:
It was previously approved, not clear when approval was withdrawn

KIvanow:
While I can see the argument for both individual components and having a more generic one, what is the difference between the changes introduced here and in this PR [https://github.com/RedisInsight/RedisInsight/pull/3887](https://github.com/RedisInsight/RedisInsight/pull/3887) ?

romansergeenkosofteq:
> While I can see the argument for both individual components and having a more generic one, what is the difference between the changes introduced here and in this PR #3887 ?

The argument is to have small functions or 1 component with these texts which we have now instead of creating some amount of the same components in different places (components/monitor/, components/pub-sub). If we continue to add the same components under feature flag we will have huge folder of components without any folder logic.
The second argument was about naming this component.

I suggested an idea how we can do it
```
SUBSCRIBE_COMMAND_ENV: (
    <div className="cli-output-response-fail" data-testid="user-pub-sub-link-disabled">PubSub not supported
      in this environment.
    </div>
  ),
  SUBSCRIBE_COMMAND_NOT_SUPPORTED: (path: string = '') => (
    <FeatureFlagComponent name={FeatureFlags.envDependent} otherwise={cliTexts.SUBSCRIBE_COMMAND_ENV}>
      {cliTexts.SUBSCRIBE_COMMAND(path)}
    </FeatureFlagComponent>
  ),
  ```
  
 In this case we can still reuse all texts in any place (with feature related or just directly). As I remember, also the problem was that our tests fail with <FeatureFlagComponent> and thats why Artem created component to use it there. I didn't investigate into it, but I think it can be resolved by moving cliOutput into components folder, it would be better

pd-redis:
@romansergeenkosofteq  you still are not addressing the fact, that there is already such somponent - very similar, and you didn't raise any concerns about it?
@ArtemHoruzhenko if you think the changes requested make sense, why did you implement UseProfileLink the way you did? How does this case differ?

pd-redis:
@romansergeenkosofteq tests did fail with this case too, which led to the implementation that is in this PR

KIvanow:
In order to try and bring some clarity around the difference between the old and new components using the same approach - is it correct to summarize that the previous component @pd-redis is referring to was approved and ok because it was one of (an exception of sorts), while this PR is indicating that the problem is deeper and that if we continue down the same path it will lead to problems (abundance of similar components, instead of a single generic one, etc)
?

PR #4180, opened 2024-12-08
RI-6363 include server_name to telemetry

kchepikava:


redisinsight/api/src/modules/database/models/database-overview.ts
```diff
@@ -4,54 +4,60 @@ export class DatabaseOverview {
   @ApiProperty({
     description: 'Redis database version',
     type: String,
-  })
+    })
```
zalenskiSofteq, line 7: I suggest to roll back all these shifts

zalenskiSofteq (approved):

vlad-dargel (approved):

PR #4187, opened 2024-12-09
CR-277: Change log to debug calls, log additional metadata

seth-riedel:


redisinsight/api/src/common/logger/app-logger.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    for (const k in optionalParams) {
+      if (this.isException(optionalParams[k])) {
+        errorMessage = (optionalParams[k] as Error).message;
+        errorStack = (optionalParams[k] as Error).stack;
+      } else if (optionalParams[k] !== contextArg) {
+        meta.push(optionalParams[k]);
+      }
+    }
+    const metaObj = meta.length > 1 ? meta : meta[0];
```
ArtemHoruzhenko, line 67: what if `meta.length === 0`?

redisinsight/api/src/common/logger/app-logger.ts
```diff
@@ -0,0 +1,126 @@
...
+        : {}),
+    };
+  }
+
+  /**
+   * Write a 'log' level log.
+   */
+  log(message: string | LogObject, ...optionalParams: ErrorOrMeta[]) {
+    const parsedArgs = this.parseLoggerArgs(message, optionalParams);
+    if (!AppLogger.startupContexts.includes(parsedArgs.context)) {
```
ArtemHoruzhenko, line 87: DO we want to ignore nestjs core logs?

seth-riedel: I think so. this hides all the starup logs where Nest logs that it is registering all the routers, etc. I don't think this suppresses any other logs that we care about. What do you think?

ArtemHoruzhenko:
@seth-riedel Looks good. Will wait for `clone` + tests

seth-riedel:
@ArtemHoruzhenko please check when you can

ArtemHoruzhenko (approved):

PR #4197, opened 2024-12-11
show original placeholder without applying visualization for big resp…

ArtemHoruzhenko:
…onses message placeholder

### DIFF
--- a/redisinsight/ui/src/components/query/query-card/QueryCard.tsx
+++ b/redisinsight/ui/src/components/query/query-card/QueryCard.tsx
@@ -152,6 +152,12 @@ const QueryCard = (props: Props) => {
 
   const commonError = CommonErrorResponse(id, command, result)
 
+  const isSizeLimitExceededResponse = (result: Maybe<CommandExecutionResult[]>) => {
+    const resultObj = result?.[0]
+    // response.includes - to be backward compatible with responses which don't include sizeLimitExceeded flag
+    return resultObj?.sizeLimitExceeded === true || resultObj?.response?.includes('Results have been deleted')
+  }
+
   return (
     <div
       className={cx(styles.containerWrapper, {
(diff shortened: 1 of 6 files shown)


zalenskiSofteq (dismissed):

zalenskiSofteq (approved):

redisinsight/ui/src/components/query/query-card/QueryCard.tsx
```diff
@@ -196,50 +202,65 @@ const QueryCard = (props: Props) => {
...
-                          query={command}
-                          resultsMode={resultsMode}
-                          result={result}
-                          isNotStored={isNotStored}
-                          isFullScreen={isFullScreen}
-                        />
-                      )}
-                    </>
-                  )}
+                  {isSizeLimitExceededResponse(result)
```
pawelangelow, line 205: A lot of nesting 🥲

KIvanow:
Please start the commits with the ticket id so we can more easily associate the changes with what caused them whenever we need to change anything later on

KIvanow (approved):

redisinsight/ui/src/components/query/query-card/QueryCard.tsx
ArtemHoruzhenko: Yeah. Otherwise we must split/refactor entire component which is time-consuming for such small task.

KIvanow: If any of notices such deep nesting and ternaries being used in the rendering, or anything else that might be a suboptimal practice, but it is done to get it done for now. Please, tag me, so we can discuss it and make a task for it to improve it whenever we have time. Otherwise we'll forget about it and things like these or similar will keep falling through the cracks.
It will also be more productive than emotes ;) 
Tagging @pd-redis and @seth-riedel so they can see it as well

pd-redis: why not use
```typescript
(isGroupResults(resultsMode) || isSizeLimitExceededResponse(result)) ? ...
```

pawelangelow: Love the initiative, @KIvanow! While I’m slightly skeptical about the 'whenever we have time' part, it’s definitely a step forward. And hey, emojis might be simple, but they’ve got some serious discussion-starting power! 😸✨

PR #4222, opened 2024-12-18
RI-6114: prevent long text overlapping in SuperSelect

pawelangelow:
# Notes

Since, the problem itself is in the component library ([check this snippet](https://codesandbox.io/p/sandbox/frosty-tristan-x5nsfw)) we may have it somewhere else as well.


pd-redis (approved):
looks good to me, not fan of !important, but i guess it will be hard to target the element in another wat

pawelangelow:
> looks good to me, not fan of !important, but i guess it will be hard to target the element in another wat

yeah - I hate `important` as well, but the html is pretty simple so I can't select it more specifically (a.k.a. styles with higher priority). An option would be to introduce an ID, but I like this approach even less.

KIvanow (approved):

mariasergeenko (approved):

PR #4223, opened 2024-12-18
CR-292 keytar startup

pd-redis:
CR-292: 
the idea here is not to intialize `keytar` in constructor, but in a getter.
This way, it should be only required when used, there is also check if already the module is required, so it should not cause any extra imports.
Tests pass, but let me know if this is viable approach?


pawelangelow (dismissed):

ArtemHoruzhenko (dismissed):

KIvanow (approved):

ArtemHoruzhenko (commented):
@pd-redis I think we can disable it with already existing env. No need code changes

pd-redis:
resolved via ENV variable

PR #4228, opened 2024-12-19
Fe/feature/ri-6330 Disable the overview refresh

KIvanow:
I've added minimumRefreshRate and defaultRefreshRate as props to the AutoRefresh component to allow it to take different values for different refreshes. In this case these are different env variables specifically for the overview. I've kept the original default values to not change the behavior for any other place the component is being used.

I am unsure where we usually handle env variable configs. Please let me know if redisinsight/ui/vite.config.mjs is the wrong place for it


kchepikava (dismissed):

redisinsight/ui/src/components/database-overview/DatabaseOverview.tsx
```diff
@@ -4,14 +4,21 @@ import { EuiFlexGroup, EuiFlexItem, EuiIcon, EuiToolTip } from '@elastic/eui'
...
 import styles from './styles.module.scss'
 
 interface Props {
   metrics?: Array<IMetric>
+  loadData: () => void
+  lastRefreshTime: number | null
+  handleEnableAutoRefresh: (enableAutoRefresh: boolean, refreshRate: string) => void
 }
 
+const TIMEOUT_TO_GET_INFO = process.env.NODE_ENV !== 'development' ? 5000 : 60_000
```
zalenskiSofteq, line 17: AutoRefresh refreshRate is in seconds

redisinsight/ui/src/components/database-overview/DatabaseOverview.tsx
```diff
@@ -56,46 +63,69 @@ const DatabaseOverview = (props: Props) => {
...
+                </EuiFlexItem>
+              ))
+            }
+            <EuiFlexItem
+              className={cx(styles.overviewItem, styles.autoRefresh)}
+              grow={false}
+              data-testid="overview-auto-refresh"
+            >
+              <EuiFlexItem grow={false} className={styles.overviewItemContent}>
+                <AutoRefresh
```
zalenskiSofteq, line 112: AutoRefresh is turned off by default. Is it right behaviour? I think the auto-activation should stay

zalenskiSofteq (commented):
vite.config.mjs is right place for env variables in the main branch. after merge feature/dynamic-dependencies, the location will be in a separate riConfig file.

redisinsight/ui/src/components/database-overview/DatabaseOverview.tsx
KIvanow: Nice catch! Thank you for the clarification. Updated it to also use the environment variable that was added for that

redisinsight/ui/src/components/database-overview/DatabaseOverview.tsx
KIvanow: Updated

zalenskiSofteq (approved):

seth-riedel (dismissed):

mariasergeenko (approved):

PR #4230, opened 2024-12-23
CR-132: web socket connection

pd-redis:
With this PR, the interface of establishing socket connection is being unified in one place, in order to make it much simpler to use.
`getSocketApiUrl` - required because of how routing on cluster works
`useIoConnection` - extract common logic, that would of been needed in every component that establishes io connection
`wsService` - Connection it-self is established in wsService (WebSocket Service), and because there is some condititional logic for where to put CSRF token, it is also being called via custom hook.

### DIFF
--- a/redisinsight/api/config/default.ts
+++ b/redisinsight/api/config/default.ts
@@ -18,7 +18,6 @@ const defaultsDir = process.env.RI_DEFAULTS_DIR
     : join(__dirname, '..', 'defaults'));
 
 const proxyPath = trim(process.env.RI_PROXY_PATH, '/');
-const socketProxyPath = trim(process.env.RI_SOCKET_PROXY_PATH, '/api');
 
 const customPluginsUri = posix.join('/', proxyPath, 'plugins');
 const staticUri = posix.join('/', proxyPath, 'static');
@@ -28,7 +27,10 @@ const contentUri = posix.join('/', proxyPath, 'static', 'content');
 const defaultPluginsUri = posix.join('/', proxyPath, 'static', 'plugins');
 const pluginsAssetsUri = posix.join('/', proxyPath, 'static', 'resources', 'plugins');
 
+const socketProxyPath = trim(process.env.RI_SOCKET_PROXY_PATH || '/api', '/');
+
 const socketPath = posix.join('/', 'api', 'socket.io');
+
 // const socketPath = posix.join('/', proxyPath, 'socket.io');
 const dataDir = process.env.RI_BUILD_TYPE === 'ELECTRON' && process['resourcesPath']
   ? join(process['resourcesPath'], 'data')
--- a/redisinsight/api/config/default.ts
+++ b/redisinsight/api/config/default.ts
@@ -27,9 +27,9 @@ const contentUri = posix.join('/', proxyPath, 'static', 'content');
 const defaultPluginsUri = posix.join('/', proxyPath, 'static', 'plugins');
 const pluginsAssetsUri = posix.join('/', proxyPath, 'static', 'resources', 'plugins');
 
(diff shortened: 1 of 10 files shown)


redisinsight/api/config/default.ts
```diff
@@ -27,7 +27,11 @@ const contentUri = posix.join('/', proxyPath, 'static', 'content');
 const defaultPluginsUri = posix.join('/', proxyPath, 'static', 'plugins');
 const pluginsAssetsUri = posix.join('/', proxyPath, 'static', 'resources', 'plugins');
 
-const socketPath = posix.join('/', proxyPath, 'socket.io');
+const socketProxyPath = trim(process.env.RI_SOCKET_PROXY_PATH || 'api', '/');
+
+const socketPath = posix.join('/', socketProxyPath, 'socket.io');
+
+// const socketPath = posix.join('/', proxyPath, 'socket.io');
```
KIvanow, line 34: No point in leaving the old implementation in a comment. We can always recover it from git

redisinsight/ui/src/services/wsService.ts
```diff
@@ -0,0 +1,53 @@
...
+  reconnection?: boolean;
+  query?: Record<string, any>;
+  extraHeaders?: Record<string, any>;
+  path?: string;
+}
+
+export function wsService(
+  wsUrl: string,
+  { forceNew = true, token, reconnection, query, extraHeaders, path = getProxyPath() }: WsParams,
+  passTokenViaHeaders: boolean = true,
```
pawelangelow, line 20: I see that we pass `envDependent` here, but what does it do? Looking at the implementation looks like we switch from query params to headers for that token, but was wondering what is the general idea here.

pd-redis: the idea/issue is, that WS connection works different in browser than on the desktop, which is why we have env dependent implementation

pawelangelow: So, `envDependent` is used for other browser/desktop issues?

redisinsight/ui/src/services/wsService.ts
```diff
@@ -0,0 +1,53 @@
...
+  let headers: Record<string, any> = {
+    [CustomHeaders.WindowId]: window.windowId || '',
+  }
+  if (passTokenViaHeaders) {
+    headers = { ...headers, [CustomHeaders.CsrfToken]: token || '' }
+  }
+
+  if (extraHeaders) {
+    headers = { ...headers, ...extraHeaders }
+  }
```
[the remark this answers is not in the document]
pawelangelow: or maybe this
```suggestion
  const csrfHeaderToken = passTokenViaHeaders
  ? { [CustomHeaders.CsrfToken]: token || '' }
  : {};
  
  const csrfQueryStringToken = !passTokenViaHeaders
    ? { [CustomHeaders.CsrfToken.toLowerCase()]: token }
    : {};
  
  const queryParams: Record<string, any> = {
    ...csrfQueryStringToken,
    ...query,
  };
  
  const headers: Record<string, any> = {
    [CustomHeaders.WindowId]: window.windowId || '',
    ...csrfHeaderToken,
    ...extraHeaders,
  };
```

redisinsight/ui/src/services/wsService.ts
pd-redis: Yes, that is what we use for "feature" flag ... I know it is not intuitive namming, but there are reasons why

pawelangelow: Thanks!

KIvanow (dismissed):

pawelangelow (approved):

pawelangelow (approved):

KIvanow (approved):

seth-riedel (approved):

PR #4251, opened 2025-01-03
#RI-6332 - fix bg-color for profiler button

rsergeenko:


redisinsight/ui/src/styles/themes/dark_theme/darkTheme.scss
```diff
@@ -93,6 +93,7 @@
   --browserTreeNodeOpen: #{$browserTreeNodeOpen};
   --defaultGreenColor: #{$defaultGreenColor};
   --insightsTriggerBgColor: #{$insightsTriggerBgColor};
+  --profileBtnBgColor: #{$insightsTriggerBgColor};
```
pawelangelow, line 96: [Question, nothing blocking] Are there mockups or designs that we get these values from? I see that we're using two separate values for the light theme for `profileBtn` and `insightTrigger` and was wondering if we should use two separate for the dark theme as well.

rsergeenko:
already fixed in the different PR

PR #4252, opened 2025-01-03
bugfix web sockets

pd-redis:
fix: get base url from api url origin instead of replacing path.
Whenever there is no path - `http://localhost` - then path becomes `/` and is being replaced in http:// part and breaks the url like `http:/localhost`

### DIFF
--- a/redisinsight/ui/src/utils/common.ts
+++ b/redisinsight/ui/src/utils/common.ts
@@ -12,7 +12,7 @@ export const getSocketApiUrl = (path = '') => {
   let baseUrl = getBaseApiUrl()
   try {
     const url = new URL(baseUrl)
-    baseUrl = baseUrl.replace(url.pathname, '')
+    baseUrl = url.origin
   } catch (e) {
     console.error(e)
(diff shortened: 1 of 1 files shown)


KIvanow (approved):

pawelangelow (approved):
I see that we don't have tests for such functionalities, but I was wondering why don't we add. My comment is not directly for this PR, just to bring up the discussion, so that's why I'm not blocking.

Something in this manner:
```
it('should return the base URL with the path when the base URL is valid', () => {
  (getBaseApiUrl as jest.Mock).mockReturnValue('https://example.com/api');

  const result = getSocketApiUrl('socket/path');

  expect(result).toBe('https://example.com/socket/path');
});

it('should handle http://localhost with no path', () => {
  (getBaseApiUrl as jest.Mock).mockReturnValue('http://localhost:3000/api');

  const result = getSocketApiUrl();

  expect(result).toBe('http://localhost:3000/');
});

...
```

seth-riedel (approved):

PR #4261, opened 2025-01-06
RI-6457: Display proper RDI version

pawelangelow:
# Description 

This task has 3 sub-tasks:

- First is to hide the RDI version on the `Status page` as it's incorrect.
- The second one is to display the proper version of RDI on the connections screen.
- And the third to update few texts related to the task.

- [x] Add tests

# Preview

Screenshot 2025-01-06 at 17 00 45

### DIFF
--- a/redisinsight/api/src/modules/rdi/rdi.service.ts
+++ b/redisinsight/api/src/modules/rdi/rdi.service.ts
@@ -18,6 +18,8 @@ import { deepMerge } from 'src/common/utils';
 import { RdiAnalytics } from './rdi.analytics';
 import { RdiClient } from './client/rdi.client';
 
+const DEFAULT_RDI_VERSION = '-';
+
 @Injectable()
 export class RdiService {
   private logger = new Logger('RdiService');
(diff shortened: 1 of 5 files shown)


redisinsight/api/src/modules/rdi/rdi.service.ts
```diff
@@ -38,6 +38,12 @@ export class RdiService {
     return Object.keys(omitBy(dto, isUndefined)).some((field) => this.connectionFields.includes(field));
   }
 
+  private static async getRdiVersion(client: RdiClient): Promise<string> {
+    const pipelineStatus = await client.getPipelineStatus();
+    const version = pipelineStatus.components.processor.version;
```
pawelangelow, line 43: And I was wondering if there's a scenario where this won't be here. For example legacy version of RDI, etc... Do we have default value for here like `N/A` @ViktarStarastsenka

ViktarStarastsenka: @pawelangelow , let's display "-" if no value is received. Just in case.

pd-redis (approved):

rsergeenko (dismissed):
fe looks good

ViktarStarastsenka (dismissed):

pd-redis (approved):
lgtm

KIvanow:
Why is this pointing to feature/dynamic-dependencies and not to main?

pd-redis:
> Why is this pointing to feature/dynamic-dependencies and not to main?

Have we changed the target for PRs to `main`? So far it has been `feature/dynamic-dependencies` for all PRs

KIvanow (approved):

PR #4273, opened 2025-01-11
DEV - Add eslint integrated prettier

pawelangelow:
# Description

Well, I noticed few files with no formatting, random code style issues, so here's a solution to the problems.

In this PR you're going to see prettier + configuration for it.

⚠️ Nothing is enforced (there's no pre-commit hook, no CI pipeline, etc...). The idea is when you work on a file and want to format it - to use some defined rules and not spend time on manual edits. Current configuration is enhanced from the repo code style - the idea is to change as less as possible.

In addition I'm adding a VSCode configuration to trigger the prettier rules when formatting files. (if you haven't changed your binding it should be `cmd + shift + f`.

# How to test?

`yarn lint`

and

`yarn prettier` or `yarn prettier:update`


seth-riedel:
I'm all for this change, but it's going to cause a lot of problems if we try to do this while maintaining two separate long-lived branches that we constantly merge into one another

pd-redis (approved):

pawelangelow:
> I'm all for this change, but it's going to cause a lot of problems if we try to do this while maintaining two separate long-lived branches that we constantly merge into one another

I agree that this change could introduce some challenges, but once we merge and rebase the branches, I believe things will stabilize moving forward. Could you share more about your specific concerns or scenarios where you foresee issues? That way, we can address them proactively.

PR #4274, opened 2025-01-11
DEV - Apply ESLint and Prettier rules across files

pawelangelow:
This PR includes cosmetic changes across multiple files, such as fixing semicolons, spaces, and other formatting issues. There are no functional or logic changes - only adjustments to resolve linting problems.

⚠️ It's easier to see some of the file changes locally, as GitHub interface... well, 2773 files 😸


seth-riedel:
I don't think we can do this right now. It will create a maintenance nightmare between our long-lived branch and main :(

pd-redis:
I tend to agree with @seth-riedel , I would prefer to have consistent formatting across the files, but not sure if merging the conflicts is worth it.
It could probably be better to enforce git hook on commit or merge to apply formatting to changed files

pawelangelow:
> I tend to agree with @seth-riedel , I would prefer to have consistent formatting across the files, but not sure if merging the conflicts is worth it. It could probably be better to enforce git hook on commit or merge to apply formatting to changed files

Yup, my idea is to have:
- pre-commit hook
- github action to run `lint` and `prettier`

If we go with just the pre-commit hook or something in this manner, won't save us the trouble - next time they are run, this would be the changeset of files again. That's why I decided to leave them as separate PR, as no logical change is done - just the comsetics changes.

About the conflicts, as my [comment](https://github.com/RedisInsight/RedisInsight/pull/4273#issuecomment-2589484568) in #4273 - I believe this would be do-it-once task. Potentially, if someone changed these files the conflict resolving is relatively easy - just run the scripts with the applied changes.

PR #4279, opened 2025-01-14
Feature/ri 6336 allow to hide show columns in browser

KIvanow:


rsergeenko (changes requested):
Could you please change mocks of useSelector and useDispatch to particular selectors (functions), since it can break tests if components use several selectors

redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-size-length/KeyDetailsHeaderSizeLength.tsx
```diff
@@ -41,13 +44,25 @@ const KeyDetailsHeaderSizeLength = ({
               position="left"
               content={(
                 <>
-                  {formatBytes(size, 3)}
+                  {isSizeTooLarge ? 'The key size is too large to run the MEMORY USAGE command, as it may lead to performance issues.' : formatBytes(size, 3)}
```
pd-redis, line 47: I know that this is common in the codebase, but if some day we go for localization, it will be beneficial to have all phrases used in the UI as constants in as few places as possible.

Just something to have in mind

KIvanow:
Moved into https://github.com/RedisInsight/RedisInsight/pull/4290 in order to trigger automated tests against the changes

PR #4289, opened 2025-01-17
#RI-5411 - Optimize builds

egor-zalenski:
### Changes

1. Reworked **ui web build**
	* Decreased size from 19MB to 13 MB
	* Split bundles into smaller chunks to enable faster concurrent downloads.
	* Updated page placeholder behavior: the placeholder remains visible until all necessary JS bundles are downloaded (previously, it was just a white screen).
	
1. Consolidated all plugins into a single plugin, now called the “general” plugin
    * Migrated build system from **Parcel** to **Vite**.
	* Reduced build size from **13MB** to **6.7MB**
	* Improved built time from ~2min to ~1min
	
1. Automated UI vendor (RI styles and fonts) updates for plugins on AWS S3. 
	* The update process is triggered by the dev build of the Linux application.


pawelangelow:
Looks like a great effort ❤️ Will look into it. Why did you closed it?

egor-zalenski:
> Looks like a great effort ❤️ Will look into it. Why did you closed it?

After discussing with Roman, we decided that having separate plugins with separet common packages would be better than a single large plugin.

PR #4294, opened 2025-01-20
RI-5747 filter per RI version in feature config

kchepikava:


redisinsight/api/src/utils/feature-version-filter.helper.spec.ts
```diff
@@ -0,0 +1,65 @@
+// Import the function to test
```
pawelangelow, line 1: probably a leftover

redisinsight/api/src/utils/feature-version-filter.helper.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+import { FeatureConfigFilterCondition } from 'src/modules/feature/model/features-config';
+import { filterVersion } from './feature-version-filter.helper';
+
+describe('filterVersion', () => {
```
pawelangelow, line 5: ❤️ 

Since these tests work with the `path` version (`major`.`minor`.`patch`) - we automatically have tests for the `minor` as well 😸

pawelangelow (approved):

PR #4295, opened 2025-01-21
RI-6547 Apply multiline ellipsis to the table tooltip

dantovska:
RI-6547 Apply multiline ellipsis to the table tooltip text that overflows, instead of the scroll.
The scroll was the auto behavior when the tooltip text was too long and it lead to the tooltip container height to exceed the one on the table result plugin container. So when attempted to scroll and view the whole text, it was impossible because when focus was lost, the tooltip disappears. 

Implemented the multiline ellipsis solution where maximum 7 lines of text are shown in the tooltip and if longer, ends with "...". If the users wants to view the description,  they can copy it using the copy icon or switch to text view.

Before:
image

After:
image

Verified in both browser and desktop.

### DIFF
--- a/redisinsight/ui/src/packages/redisearch/src/components/TableResult/TableResult.tsx
+++ b/redisinsight/ui/src/packages/redisearch/src/components/TableResult/TableResult.tsx
@@ -77,6 +77,7 @@ const TableResult = React.memo((props: Props) => {
             <EuiToolTip
               position="bottom"
               title={title}
+              className="text-multiline-ellipsis"
               anchorClassName={cx('tooltip')}
               content={formatLongName(value.toString())}
             >
--- a/redisinsight/ui/src/packages/redisearch/src/styles/styles.scss
+++ b/redisinsight/ui/src/packages/redisearch/src/styles/styles.scss
@@ -85,6 +85,10 @@ html {
   display: inline-block !important;
 }
 
+.text-multiline-ellipsis>div:last-child{
+  @include eui.multiline-ellipsis($line-count: 7, $line-height: 1.5rem);
+}
+
 .cell {
   position: relative;
 }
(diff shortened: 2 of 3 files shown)


redisinsight/ui/src/packages/redisearch/src/styles/styles.scss
```diff
@@ -85,6 +85,10 @@ html {
   display: inline-block !important;
 }
 
+.text-multiline-ellipsis>div:last-child{
```
pawelangelow, line 88: Maybe it's very nit comment, but it wont hurt to include the structure as comment. Something like:
```
<div class="text-multiline-ellipsis">
  <div>something not relevant</div>
  <div>other non relevant thing</div>
  <div>here's the long text that needs to be multilined</div>
</div>
```

pawelangelow (dismissed):

redisinsight/ui/src/packages/redisearch/src/components/TableResult/TableResult.tsx
```diff
@@ -77,6 +77,7 @@ const TableResult = React.memo((props: Props) => {
             <EuiToolTip
               position="bottom"
               title={title}
+              className="text-multiline-ellipsis"
               anchorClassName={cx('tooltip')}
               content={formatLongName(value.toString())}
```
pawelangelow, line 82: Alternative approach is changing the arguments of `formatLongName` here. However, I like the css implementation more.

redisinsight/ui/src/packages/redisearch/src/styles/styles.scss
dantovska: You are right, I will add some documentation so it is clear to which the style is applied. Thanks!

pawelangelow (approved):

KIvanow (approved):

PR #4296, opened 2025-01-21
RI-6584 migrate to new jsonpath

kchepikava:


redisinsight/api/src/modules/browser/rejson-rl/rejson-rl.service.ts
```diff
@@ -66,7 +92,8 @@ export class RejsonRlService {
         BrowserToolRejsonRlCommands.JsonDebug,
         'MEMORY',
         keyName,
-        path,
+        // path,
```
ArtemHoruzhenko, line 95: do we need this comment?

ArtemHoruzhenko (dismissed):
lgtm

PR #4304, opened 2025-01-23
RI-6527 Resolve an issue on Ubuntu with a space in a folder name && RI-6551 Resolve the app name issue for Debian

KIvanow:
Added the fix suggested in the reported issue - https://github.com/RedisInsight/RedisInsight/issues/3871#issuecomment-2392710593 as an after install script. It resolved the issue on Ubuntu 24

We can make it less verbose if we want.


electron-builder.json
```diff
@@ -131,13 +154,18 @@
       "Comment": "Redis GUI by Redis Ltd"
     }
   },
+  "deb": {
+    "afterInstall": "scripts/linux-after-install.sh"
+  },
```
pawelangelow, line 159: Comment: I personally like the re-formatted file, but it's harder to notice the change.

Question/feedback: `linux-after-install.sh` maybe would be better if its `deb-after-install.sh` as we may need to do other hacks for different distributions.

pawelangelow (dismissed):

electron-builder.json
KIvanow: re the comment - yes, unfortunately I've added it our linter to run on save as suggested by other team members and it leads to this.

question/feedback - that's a nice suggestion. Updated it.

pawelangelow (dismissed):

dantovska (dismissed):

pd-redis (approved):

pawelangelow (approved):

PR #4314, opened 2025-01-24
RI-6613: Rework redisearch integration tests

dantovska:
Given the upcoming release for Redis 8, we want to make sure that all Redis Insight features work fine.
However, RediSearch tests need reworking as RediSearch will be part of Redis 8, and up until now it was part of Redis Stack Server, so reworking will be fixing both testing on Redis 8 and Redis Stack Server.

Changes applied: 

1. RediSearch Info
- Response body types have changed, as well as an error message. 
- Fixed to match both old and new schema types
- Fixed error message test to compare all in lower case, as the casing is the only thing changed

2. RediSearch Search
- Indexing takes place before the search but since it is an async operation, the indexing has not yet finished when the search takes place.
- Added a method to wait some time (not indefinitely) for the indexing to complete, before the searching takes place. 

Verified on redis:8.0-M02-alpine and redis-stack-server:latest


redisinsight/api/test/helpers/data/redis.ts
```diff
@@ -486,8 +486,37 @@ export const initDataHelper = (rte) => {
...
     await sendCommand('ft.create', [constants.TEST_SEARCH_HASH_INDEX_2, 'on', 'hash', 'schema', '*', 'text']);
+
+    await waitIndexingToComplete([constants.TEST_SEARCH_HASH_INDEX_1, constants.TEST_SEARCH_HASH_INDEX_2])
   };
 
+  const waitIndexingToComplete = async (hashIndexes: string[]) => {
+    const retryLimit = 3;
+    let indexesCompleted = new Array(hashIndexes.length).fill(false);
+    for(let retryCounter = 0; retryCounter < retryLimit; retryCounter++) {
+      await new Promise((resolve) => setTimeout(resolve, 300));
```
pawelangelow, line 497: Is there anything special in `300ms` or it can be any number?

redisinsight/api/test/helpers/data/redis.ts
```diff
@@ -486,8 +486,37 @@ export const initDataHelper = (rte) => {
...
+
+      for (let i = 0; i < hashIndexes.length; i++) {
+        // ft.info command returns an array which contains data that shows wether the indexing is in progress
+        // it looks something like this: ["index_name", "the_index_name", ... "indexing", 0, ...]
+        const indexInfo = await sendCommand('ft.info', [hashIndexes[i]]);
+
+        // searching for the "indexing" property index, so we can reach it's value using index + 1 in the info array
+        const indexingPropertyIndex = indexInfo.indexOf("indexing");
+        const indexingValue = indexInfo[indexingPropertyIndex + 1]; // 1 - in progress, 0 - completed
+        indexesCompleted[i] = !indexingValue;
```
pawelangelow, line 507: ❤️

redisinsight/api/test/helpers/data/redis.ts
```diff
@@ -486,8 +486,37 @@ export const initDataHelper = (rte) => {
...
+      }
+
+      if (!indexesCompleted.includes(false)) {
+        break;
+      }
+    }
+
+    if (indexesCompleted.includes(false)) {
+      console.log('Indexing has not yet completed');
```
pawelangelow, line 516:
```suggestion
      console.error('Indexing has not yet completed');
```
probably

pawelangelow (approved):

pawelangelow:
Actually, have you tried testing it with older versions? I guess we want to cover both older version/s and the new one.

KIvanow:
> Actually, have you tried testing it with older versions? I guess we want to cover both older version/s and the new one.

we should keep supporting old versions. There are plenty of people using much older ones

redisinsight/api/test/helpers/data/redis.ts
dantovska: It can be any number, I did some testing with different numbers and saw that some indexing takes no more than 300, some no more than 500 and this made most sense, to try three times 3*300 and not be more than a second.

dantovska:
What do you mean by older versions?

KIvanow:
> What do you mean by older versions?

before CE 8

dantovska:
> > What do you mean by older versions?
> 
> before CE 8

It works fine with redis stack 7 (latest).
Changes do not work with redis stack 6 (two of the tests):

1. The error message has been changed in 7 "Unknown Index name" (capital I in Index) compared to "Unknown index name"
2. In redis stack 6 redisearch info response has the types like before, not the way I changed it

Both errors are actually expected. The question is, how do we want the tests to behave? If we want for the tests to work with both versions, I would actually have to remove these checks I think.
I also think that it is a good thing that we caught these changes in the contract, IMO I should change the api model with the right types (as in redis stack 7 and redis 8).

What do you think? @KIvanow @pawelangelow

pawelangelow:
> What do you think? @KIvanow @pawelangelow

I know for sure we don't want to make a choice which version to support - a.k.a. we need to support both. Maybe the easiest way is to deal with this problem is to have a separate suite of tests for the new version.

In order to do this, we may need to add some logic to `/redisinsight/api/test/helpers/test/conditionalIgnore.ts` similar to the `requirements` helper. Basically skip these tests if version is X. Maybe @ArtemHoruzhenko can throw some light on the matter or have ideas for different approach.

pd-redis (approved):

pd-redis (approved):

redisinsight/api/test/api/redisearch/POST-databases-id-redisearch-info.test.ts
```diff
@@ -66,9 +88,17 @@ const responseSchema = Joi.object({
...
+    Joi.string(),
+    Joi.number()
+  ),
   total_indexing_time: Joi.string(),
-  indexing: Joi.string(),
+  indexing: Joi.alternatives()
+  .try(
+    Joi.string(),
+    Joi.number()
+  ),
```
pawelangelow, line 101: I'm 50:50 on this.

What is the purpose of these tests? - I see it as we want to validate we're supporting the Redis software (including modules, etc..) with the specific API (`indexing` being `string` or `number`). I'm not blocking the PR, just making tests to pass in various of conditions comes with penalty that sometimes you can't rely on them (or at least 100% on them).

pawelangelow (approved):

redisinsight/api/test/api/redisearch/POST-databases-id-redisearch-info.test.ts
pawelangelow: cc: @KIvanow @ArtemHoruzhenko @pd-redis

redisinsight/api/test/api/redisearch/POST-databases-id-redisearch-info.test.ts
```diff
@@ -114,11 +144,17 @@ describe('POST /databases/:id/redisearch/info', () => {
           index: 'Invalid index',
         },
         statusCode: 500,
-        responseBody: {
-          message: 'Unknown Index name',
-          error: 'Internal Server Error',
-          statusCode: 500,
-        },
+        checkFn: async (actual) => {
```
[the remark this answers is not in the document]
dantovska: Not needed. Removing it. Thanks!

redisinsight/api/test/api/redisearch/POST-databases-id-redisearch-info.test.ts
pd-redis: Maybe have separate schemas pre and post v7, check redis version and merge base schema with specific one for the tested version of redis? I am not sure if Joi supports schema merging (Zod does)

pawelangelow: The perfect scenario for me would be to have a test util, similar to `requirements('!rte.bigData', 'rte.modules.search');`  https://github.com/RedisInsight/RedisInsight/blob/89c49f2004bc42a6d291496f960c2df1d1ed977a/redisinsight/api/test/api/redisearch/POST-databases-id-redisearch-info.test.ts#L118

something like `requirements('!rte.bigData', 'rte.modules.search', 'rte.version < 8');` and have separate suites for both versions. This way we have some expectation over the contract and something will break if we don't expect any change.

Of course, this is not part of the current task scope, so I'm OK merging it this way and introduce that util in future PR. Then we can refactor it.

My motivation is that IMO we should treat RTE as third party software, so we should test not only the functionality but the contract (API) as well.

dantovska: Thanks for the suggestions!
I'm working on a solution considering what you mentioned above and will include it as part of this PR changes as there is no reason not to, at least not at this point.

ArtemHoruzhenko: Imho I'm not sure that this does make much sense for this particular case. We are not relying on this property in core logic and it sdoesn't make sense to spend time on it. When we start rely on it most brobably we will have other errors which will require change core + tests logic.

pawelangelow (dismissed):

redisinsight/api/test/api/redisearch/POST-databases-id-redisearch-info.test.ts
dantovska: Thanks! 
It didn't took as much time as I expected since I managed to retrieve the version quite fast and check it as part of the already implemented `requirements()` logic.

redisinsight/api/test/api/redisearch/POST-databases-id-redisearch-info.test.ts
```diff
@@ -97,12 +133,40 @@ describe('POST /databases/:id/redisearch/info', () => {
     );
   });
 
-  describe('Common', () => {
+  describe('Common, redis version <= 6', () => {
+    requirements(() => redisMajorVersion <= 6 );
```
ArtemHoruzhenko, line 137: you can simply write `requirements('rte.env.version<=6')` and it should work the same (at least it should)

Anyway this will be incorrect. If we decided to to have different schemas we must rely on redisearch module version rather than redis version. Not sure if module versions available in `rte.env` (most probably they are) or we need to add this functionality. Also there is a chance that Redis database image might have 99.99 redisearch version but in fact it will be 2.0.13 or something else with old schema. So we need to consider this too

ArtemHoruzhenko (changes requested):

redisinsight/api/test/api/redisearch/POST-databases-id-redisearch-info.test.ts
dantovska: `requirements('rte.modules.search.version>=20800')` this works, so I will use it. Thanks! 
I'm also observing on this page - https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/release-notes/redisearch/, that each redisearch version has a "at least compatible" redis version with which it works, and I notice that wether I use:
`requirements('rte.modules.search.version>=20800')`
or
`requirements('rte.env.version>=7')`
I'm getting the same results. This leads me to think that the redisearch 2.8.X introduced the different contracts because redis 7 required it... 

So I want to see what would happen if I run the tests on redis version 7 with redisearch version earlier than the breaking 2.8.X, for example 2.6.24 since there are chances of such scenario. Can you help me out with that?

pawelangelow (approved):

PR #4335, opened 2025-01-30
RI-6573 Replace displayed RDI deploy errors with 'Download error log' button

pawelangelow:
Well, instead of getting the RDI deploy error on the notification toast, now you can download it on a txt file.

Here's a preview:
Screenshot 2025-01-30 at 16 18 54

And an example log file:
[log (1).txt](https://github.com/user-attachments/files/18605372/log.1.txt)

If you want to test it, you'll need RDI endpoint, so let me know to give you one (I run it in AWS and I stop the machine when not needed)


redisinsight/ui/src/components/notifications/components/rdi-deploy-error-content/helpers/useTextFileGenerator.tsx
```diff
@@ -0,0 +1,27 @@
+import { useCallback } from 'react'
+
+const useTextFileDownload = () => {
```
KrumTy, line 3: does this have to be a hook? No params, no dependencies for useCallback, sounds like it can be a one instance function

pawelangelow: Oh, you're right! I've started it as different approach, but then simplified it, so yeah, doesn't make sense to be a hook

redisinsight/ui/src/components/notifications/components/rdi-deploy-error-content/RdiDeployErrorContent.tsx
```diff
@@ -1,17 +1,61 @@
...
+          <EuiFlexItem grow={false}>
+            <EuiButton
+              fill
+              size="s"
+              color="warning"
+              onClick={downloadFile}
+              className="toast-danger-btn"
+              data-testid="donwload-log-file-btn"
+            >
+              Download Error Log File
```
KrumTy, line 53: I was wondering if this element can be wrapped in a `<a href={fileUrl} download={fileName}>` that does the same thing instead of creating, clicking and removing an html element on the go?

pawelangelow: Good idea. Looks like `EuiButton` is smart 😸 : ` ExclusiveUnion<EuiButtonPropsForAnchor, EuiButtonPropsForButton>;`. Looks way better now, so thanks!

pd-redis (approved):

KrumTy (approved):
LGTM

PR #4368, opened 2025-02-11
RI-6382: Support both sources/targets test connection

pawelangelow:
# Description

We want to be able to test the connection to a source and a target. (currently we test only to the target).

It's a bit tricky as we need to ensure we have backwards compatibility for older versions of RDI, where the sources connection dry-run isn't supported.

My advice is - go commit by commit to review the process. In order to test it - you'll need a RDI endpoint, so let me know for credentials and address.

A little bit domain knowledge:
- even it's `sources` there can be only one
- and a lot of `targets`

# Preview

## All good:

  Screenshot 2025-02-11 at 19 11 10

## Some errors:

  Screenshot 2025-02-11 at 19 12 52

  Screenshot 2025-02-11 at 19 12 03

## Complete disaster 

  Screenshot 2025-02-11 at 19 15 45


redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
```diff
@@ -171,17 +200,37 @@ export class ApiRdiClient extends RdiClient {
...
-      throw wrapRdiPipelineError(e);
+    try {
+      const sourcesOptions = prepareTestSourcesConnectionsOptions(config);
+      const sourcesResponse = await this.client.post(
+        RdiUrl.TestSourcesConnections,
+        { ...sourcesOptions },
+      );
+      sources = sourcesResponse.data;
+    } catch (error) {
+      // do nothing - failing is expected on RDI version below 1.6.0 (1.4.3 for example)
```
KrumTy, line 230: do we want to log this somewhere for analysis?
also you can set sources here to the error value as opposed to init with that value

redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
```diff
@@ -37,6 +39,33 @@ import { RdiResetPipelineFailedException } from '../exceptions/rdi-reset-pipelin
...
+  //   "type": "cdc",
+  //   "connection": {
+  //     "type": "mysql",
+  //     "host": "localhost",
+  //     "port": 3306,
+  //     "database": "demo_db",
+  //     "user": "user",
+  //     "password": "password",
+  //   }
+  // }
```
dantovska, line 61: Do these commented lines refer to the return object `{ ...extractedBody }`, or to the function parameter `config: ConfigWithSources`? Maybe we should add this documentation at the top?

redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
```diff
@@ -37,6 +39,33 @@ import { RdiResetPipelineFailedException } from '../exceptions/rdi-reset-pipelin
...
+
+  const sourcesOptions = (config as any).sources || {};
+  const rootKey = Object.keys(sourcesOptions)[0];
+  const extractedBody = sourcesOptions[rootKey];
```
KrumTy, line 65: Is the example meant to represent `config.sources` ? Or  `config.sources[rootKey]`?
Also rootKey - looks like sources is supposed to be an object that has one key?
Maybe some better typing or more comments would help

redisinsight/ui/src/pages/rdi/pipeline-management/components/test-connections-log/TestConnectionsLog.tsx
```diff
@@ -14,12 +13,12 @@ enum ResultsStatus {
 }
 
 export interface Props {
-  data: Nullable<TransformResult>
+  data: TransformGroupResult
```
dantovska, line 16: I guess it was set as nullable but it cannot actually be null, that why you're changing it?

dantovska (approved):
Tested it locally, works great.
Kudos!

redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
pawelangelow: Okay, both of you commented this, so I'll do something about it.

redisinsight/ui/src/pages/rdi/pipeline-management/components/test-connections-log/TestConnectionsLog.tsx
pawelangelow: Yes, it was this way, because there's a case where you don't have `results` (data) but the `loading` is done - this is when there's some error. I reworked the structure of the parent component `TestConnectionsPanel` because - it's parent job to know, when to call this component (TestConnectionsLog), and this component should be dump if the data is there or not 😸

redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
ViktarStarastsenka: @pawelangelow , I think it is better to log them.

redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
```diff
@@ -37,6 +39,29 @@ import { RdiResetPipelineFailedException } from '../exceptions/rdi-reset-pipelin
...
+//       ...other connection options
+//     }
+//   }
+// }
+interface ConnectionsConfig {
+  sources: Record<string, Record<string, unknown>>;
+}
+
+const prepareTestSourcesConnectionsOptions = (config: ConnectionsConfig) => {
+  const sourcesOptions = config.sources || {};
```
ArtemHoruzhenko, line 59: does it make sense to add `?` here `config?.sources || {};` if we assume that we might have some unexpected config

redisinsight/api/src/modules/rdi/dto/rdi-test-connections.response.dto.ts
```diff
@@ -39,11 +39,28 @@ export class RdiTestConnectionResult {
...
+  @Type(() => RdiSourcesConnectionResult)
+  sources: RdiSourcesConnectionResult;
+
+  @ApiProperty({
+    description: 'Targets connection results',
+  })
+  @Expose()
   @Type(() => RdiTestConnectionResult)
-  sources: Record<string, RdiTestConnectionResult>;
+  targets: Record<string, RdiTestConnectionResult>;
```
ArtemHoruzhenko, line 65: `Record<string, RdiTestConnectionResult>;` doesn't match with what you have in transformer `@Type(() => RdiTestConnectionResult)`. Not sure how it will work.

redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
pawelangelow: I tried few cases, but we always get a `config`. You can send a `config`, without a `sources`, but we got this scenario covered. Anyways this is inside try catch, so we're safe.

Screenshot 2025-02-14 at 16 00 02

redisinsight/api/src/modules/rdi/dto/rdi-test-connections.response.dto.ts
ArtemHoruzhenko: Awesome! It will be nice if you will create new decorator for this so everyone can reuse it in the future. 
But it could be done later in a separate task.

ArtemHoruzhenko (approved):
lgtm

KrumTy (approved):

PR #4369, opened 2025-02-12
API Integration test: fix types + add vscode debugger config

KrumTy:
### Description
Allows easier tests debugging and fixes test ts issues

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,23 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug API Integration Tests",
+      "skipFiles": ["<node_internals>/**"],
+      "program": "${workspaceFolder}/redisinsight/api/node_modules/ts-mocha/bin/ts-mocha",
+      "args": [
+        "--paths",
+        "--config",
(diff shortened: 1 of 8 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
...
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug API Integration Tests",
+      "skipFiles": ["<node_internals>/**"],
+      "program": "${workspaceFolder}/redisinsight/api/node_modules/ts-mocha/bin/ts-mocha",
+      "args": [
+        "--paths",
+        "--config",
```
dantovska, line 12: did you try adding "-p" "<path_to_tsconfig.json>" file here, instead of adding it to .mocharc.yml and removing it from package.json?

dantovska (approved):
I have just have one question/minor suggestion. 
Other than that, it looks great! Thanks x10000!

.vscode/launch.json
KrumTy: yes, it's the same
however I think putting it in the mocha config is better since regardless how we run the mocha tests command, we always want it to point to the same tsconfig (at least for now)

KIvanow (approved):

PR #4371, opened 2025-02-13
RI-6614: Rework Consumer Group integration tests

dantovska:
Given the upcoming release for Redis 8, we want to make sure that all Redis Insight features work fine.
However, consumer group tests need reworking due to some changes in their contracts which occurred in Redis 7, more precisely two new properties were added to the post and patch consumer group responses:
- entries_read
- lag

Changes applied:
- Separate the same tests to run when Redis < 7 and Redis >=7 with the appropriate response body expectations

Same was done for the redisearch tests -> https://github.com/RedisInsight/RedisInsight/pull/4314 

Verified on: redis:8.0-M02-alpine, redis:7.2.4 , redis: 6.2.17


pd-redis (approved):
lgtm

PR #4372, opened 2025-02-13
RI-6605 handle Redis.quit rejection

ArtemHoruzhenko:
`client.quit` is async function which wasn't handled properly


dantovska (approved):
Looks great! 
Reproduced the issue using electron on the main branch and then confirmed that it is fixed on this branch. Approving.

pd-redis (commented):
lgtm

PR #4377, opened 2025-02-17
RI-6658: fallback to HELLO command when INFO is disabled via ACL

KrumTy:
### Description

### Notes
- this PR addresses issues related to db connectivity when access to INFO is disabled
- type-wise, the only mandatory field from the info response was 'version', something that's also present in the hello response
- connecting to db only requires info command's response to check the redis' version (for [recommendation](https://github.com/RedisInsight/RedisInsight/blob/main/redisinsight/api/src/modules/database/database-connection.service.ts#L55-L65) and client info collect)

### Updates
- this PR evolved a bit as per @ArtemHoruzhenko 's suggestion
- the logic for retrieving the info data has been unified inside the base client (see `getInfo` in redis.client.ts)
- it basically checks for info, if no permissions are available - checks for data from hello; 
- in case hello is not available (redis v < 6.2.0) or something fails, defaults to a basic static object 
- the original PR was aimed at solving just RI-6849 but it so happens that it resolves multiple issues

### DIFF
--- a/redisinsight/api/src/common/constants/general.ts
+++ b/redisinsight/api/src/common/constants/general.ts
@@ -4,6 +4,6 @@ export enum TransformGroup {
 
 export const UNKNOWN_REDIS_INFO = {
   server: {
-    version: 'unknown',
+    redis_version: 'unknown',
   },
 };
--- a/redisinsight/api/src/common/constants/general.ts
+++ b/redisinsight/api/src/common/constants/general.ts
@@ -1,3 +1,7 @@
 export enum TransformGroup {
   Secure = 'security',
 }
+
+export const UNKNOWN_REDIS_INFO = {
+  version: 'unknown',
+};
--- a/redisinsight/api/src/common/constants/general.ts
+++ b/redisinsight/api/src/common/constants/general.ts
@@ -5,5 +5,6 @@ export enum TransformGroup {
 export const UNKNOWN_REDIS_INFO = {
   server: {
     redis_version: 'unknown',
+    redis_mode: 'standalone',
   },
 };
--- a/redisinsight/api/src/common/constants/general.ts
+++ b/redisinsight/api/src/common/constants/general.ts
@@ -3,5 +3,7 @@ export enum TransformGroup {
 }
 
 export const UNKNOWN_REDIS_INFO = {
-  version: 'unknown',
+  server: {
+    version: 'unknown',
+  },
 };
--- a/redisinsight/api/src/constants/error-messages.ts
+++ b/redisinsight/api/src/constants/error-messages.ts
@@ -59,6 +59,7 @@ export default {
     'Key with this name does not exist or does not have an associated timeout.',
   SERVER_NOT_AVAILABLE: 'Server is not available. Please try again later.',
   REDIS_CLOUD_FORBIDDEN: 'Error fetching account details.',
+  NO_INFO_COMMAND_PERMISSION: 'has no permissions to run the \'info\' command',
 
   DATABASE_IS_INACTIVE: 'The database is inactive.',
   DATABASE_ALREADY_EXISTS: 'The database already exists.',
--- a/redisinsight/api/src/constants/error-messages.ts
+++ b/redisinsight/api/src/constants/error-messages.ts
@@ -59,7 +59,6 @@ export default {
     'Key with this name does not exist or does not have an associated timeout.',
   SERVER_NOT_AVAILABLE: 'Server is not available. Please try again later.',
   REDIS_CLOUD_FORBIDDEN: 'Error fetching account details.',
-  NO_INFO_COMMAND_PERMISSION: 'has no permissions to run the \'info\' command',
 
   DATABASE_IS_INACTIVE: 'The database is inactive.',
   DATABASE_ALREADY_EXISTS: 'The database already exists.',
--- a/redisinsight/api/src/modules/cluster-monitor/cluster-monitor.service.ts
+++ b/redisinsight/api/src/modules/cluster-monitor/cluster-monitor.service.ts
@@ -2,7 +2,7 @@ import { get } from 'lodash';
 import {
   BadRequestException, HttpException, Injectable, Logger,
 } from '@nestjs/common';
-import { catchAclError, convertRedisInfoReplyToObject } from 'src/utils';
+import { catchAclError } from 'src/utils';
 import { IClusterInfo } from 'src/modules/cluster-monitor/strategies/cluster.info.interface';
 import { ClusterNodesInfoStrategy } from 'src/modules/cluster-monitor/strategies/cluster-nodes.info.strategy';
 import { ClusterShardsInfoStrategy } from 'src/modules/cluster-monitor/strategies/cluster-shards.info.strategy';
@@ -41,10 +41,7 @@ export class ClusterMonitorService {
         return Promise.reject(new BadRequestException('Current database is not in a cluster mode'));
       }
 
-      const info = convertRedisInfoReplyToObject(await client.sendCommand(
-        ['info', 'server'],
-        { replyEncoding: 'utf8' },
-      ) as string);
+      const info = await client.getInfo(true, 'server');
 
       const strategy = this.getClusterInfoStrategy(get(info, 'server.redis_version'));
 
--- a/redisinsight/api/src/modules/cluster-monitor/cluster-monitor.service.ts
+++ b/redisinsight/api/src/modules/cluster-monitor/cluster-monitor.service.ts
@@ -41,7 +41,7 @@ export class ClusterMonitorService {
         return Promise.reject(new BadRequestException('Current database is not in a cluster mode'));
       }
 
-      const info = await client.getInfo(true, 'server');
+      const info = await client.getInfo('server');
 
       const strategy = this.getClusterInfoStrategy(get(info, 'server.redis_version'));
 
--- a/redisinsight/api/src/modules/database/providers/database-info.provider.ts
+++ b/redisinsight/api/src/modules/database/providers/database-info.provider.ts
@@ -79,10 +79,7 @@ export class DatabaseInfoProvider {
    */
   public async determineDatabaseServer(client: RedisClient): Promise<string> {
     try {
-      const reply = convertRedisInfoReplyToObject(await client.call(
-        ['info', 'server'],
-        { replyEncoding: 'utf8' },
-      ) as string);
+      const reply = await this.getRedisInfo(client);
       return reply['server']?.redis_version;
     } catch (e) {
       // continue regardless of error
@@ -138,10 +135,7 @@ export class DatabaseInfoProvider {
     client: RedisClient,
   ): Promise<RedisDatabaseInfoResponse> {
     try {
-      const info = convertRedisInfoReplyToObject(await client.sendCommand(
-        ['info'],
-        { replyEncoding: 'utf8' },
-      ) as string);
+      const info = await this.getRedisInfo(client);
       const serverInfo = info['server'];
       const memoryInfo = info['memory'];
       const keyspaceInfo = info['keyspace'];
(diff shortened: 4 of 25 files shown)


redisinsight/api/src/utils/converter.ts
```diff
@@ -36,3 +36,21 @@ export const convertStringToNumber = (value: any, defaultValue?: number): number
...
+        : {
+            ...memo,
+            [current]: Array.isArray(array[index + 1]?.[0])
+              ? array[index + 1].map((element) =>
+                  convertArrayOfKeyValuePairsToObject(element),
+                )
+              : array[index + 1],
+          },
+    {},
+  );
```
pawelangelow, line 56: I'm sure there's a crying kitten somewhere that wanted tests for this fn 😞

pawelangelow: And since we use lodash, i'm wondering something looking less like elf spell 😄 

```
export const convertArrayOfKeyValuePairsToObject = (array: any[]): Record<string, any> => {
  const obj = _.fromPairs(_.chunk(array, 2));

  return _.mapValues(obj, (value) =>
    Array.isArray(value?.[0]) ? value.map((element) => convertArrayOfKeyValuePairsToObject(element)) : value
  );
};
```
🤔

redisinsight/api/src/modules/database/providers/database-info.provider.ts
```diff
@@ -261,4 +274,17 @@ export class DatabaseInfoProvider {
       throw catchAclError(e);
     }
   }
+
+  private async getRedisHelloInfo(client: RedisClient): Promise<RedisDatabaseHelloResponse> {
+    try {
+      const helloResponse = convertArrayOfKeyValuePairsToObject(await client.sendCommand(
+        ['hello'],
+        { replyEncoding: 'utf8' },
+      ) as any[]);
```
pawelangelow, line 283: Isn't it `string[]`?

redisinsight/api/src/utils/converter.ts
KIvanow: Please add a few test cases for this util, regardless of the crying kittens.

Also be careful with the nested ternaries. After a certain depth, it makes it really difficult to read

KrumTy: agreed, will refactor + add tests

redisinsight/api/src/modules/database/providers/database-info.provider.ts
KrumTy: apparently not (redis v7+ at least)
is it possible for it to not be an array for any reason?

redisinsight/api/src/utils/converter.ts
```diff
@@ -36,3 +36,15 @@ export const convertStringToNumber = (value: any, defaultValue?: number): number
...
   return num;
 };
+
+export const convertArrayOfKeyValuePairsToObject = (array: any[]) => {
+  const result: Record<string, any> = {};
+
+  for (let i = 0; i + 1 < array.length; i += 2) {
+    const key = array[i];
+    const value = array[i + 1];
+    result[key] = value;
```
dantovska, line 46: I noticed that in the previous form of the method, we were recursively calling this method if the value is an array. Is there a reason why it is not happening now?

KrumTy: I've simplified the function to ease understanding and testing.
The array case was just one field (modules) so I simply call the function again: see `database-info.provider.ts:303`

redisinsight/api/src/modules/database/providers/database-info.provider.ts
```diff
@@ -261,4 +258,54 @@ export class DatabaseInfoProvider {
...
+          replyEncoding: 'utf8',
+        })) as string,
+      );
+    } catch (error) {
+      if (error.message.includes(ERROR_MESSAGES.NO_INFO_COMMAND_PERMISSION)) {
+        // Fallback to getting basic information from `hello` command
+        return this.getRedisHelloInfo(client);
+      }
+
+      throw error;
```
ArtemHoruzhenko, line 275: what will happen when `hello` command is also disabled?
Also you are handling an error in different way for `info` and `hello` commands. Is it expected behavior or we need to add `catchAclError` here too?

ArtemHoruzhenko (commented):
I'm not sure that using `DatabaseInfoProvider` will be easy in all app modules. I assume we might have a lot of circular dependencies then. Let's move this logic to RedisClient so it will be available everywhere. wdyt?

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -41,13 +39,7 @@ export class RecommendationProvider {
...
   ): Promise<Recommendation> {
     try {
-      const info = convertRedisInfoReplyToObject(
-        await redisClient.sendCommand(
-          ['info', 'memory'],
-          { replyEncoding: 'utf8' },
-        ) as string,
-      );
-
+      const info = await redisClient.getInfo(true, 'memory');
```
ArtemHoruzhenko, line 42: let's remove `memory` arg

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -98,12 +90,7 @@ export class RecommendationProvider {
       return null;
     }
     try {
-      const info = convertRedisInfoReplyToObject(
-        await redisClient.sendCommand(
-          ['info', 'keyspace'],
-          { replyEncoding: 'utf8' },
-        ) as string,
-      );
+      const info = await redisClient.getInfo(true, 'keyspace');
```
ArtemHoruzhenko, line 93: and keyspace here

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -296,12 +283,7 @@ export class RecommendationProvider {
     redisClient: RedisClient,
   ): Promise<Recommendation> {
     try {
-      const info = convertRedisInfoReplyToObject(
-        await redisClient.sendCommand(
-          ['info', 'clients'],
-          { replyEncoding: 'utf8' },
-        ) as string,
-      );
+      const info = await redisClient.getInfo(true, 'clients');
```
ArtemHoruzhenko, line 286: clients

redisinsight/api/src/modules/recommendation/providers/recommendation.provider.ts
```diff
@@ -343,12 +325,7 @@ export class RecommendationProvider {
     redisClient: RedisClient,
   ): Promise<Recommendation> {
     try {
-      const info = convertRedisInfoReplyToObject(
-        await redisClient.sendCommand(
-          ['info', 'server'],
-          { replyEncoding: 'utf8' },
-        ) as string,
-      );
+      const info = await redisClient.getInfo(true, 'server');
```
ArtemHoruzhenko, line 328: server

redisinsight/api/src/modules/redis/client/redis.client.ts
```diff
@@ -157,18 +161,67 @@ export abstract class RedisClient extends EventEmitter2 {
    * Get redis database info
    * Uses cache by default
    * @param force
+   * @param infoSection - e.g. server, clients, memory, etc.
    */
-  public async getInfo(force = false): Promise<object> {
+  public async getInfo(force = true, infoSection?: string) {
```
ArtemHoruzhenko, line 166: let's remove infoSection

redisinsight/api/src/modules/redis/client/redis.client.ts
```diff
@@ -157,18 +161,67 @@ export abstract class RedisClient extends EventEmitter2 {
...
+      try {
+        const infoData = convertRedisInfoReplyToObject(await this.call(
+          infoSection ? ['info', infoSection] : ['info'],
+          { replyEncoding: 'utf8' },
+        ) as string);
+
+        this.info = {
+          ...this.info,
+          ...infoData,
+        }
```
ArtemHoruzhenko, line 177: let's replace this with entirely new object we've got

redisinsight/api/src/modules/redis/utils/keys.util.ts
```diff
@@ -1,16 +1,11 @@
...
 
 export const getTotalKeysFromInfo = async (client: RedisClient) => {
   try {
     const currentDbIndex = await client.getCurrentDbIndex();
-    const info = convertRedisInfoReplyToObject(
-      await client.sendCommand(['info', 'keyspace'], {
-        replyEncoding: 'utf8',
-      }) as string,
-    );
+    const info = await client.getInfo(true, 'keyspace');
```
ArtemHoruzhenko, line 8: keyspace here

redisinsight/api/src/common/constants/general.ts
```diff
@@ -1,3 +1,10 @@
 export enum TransformGroup {
   Secure = 'security',
 }
+
+export const UNKNOWN_REDIS_INFO = {
+  server: {
+    redis_version: 'unknown',
+    redis_mode: 'standalone',
```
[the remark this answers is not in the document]
KrumTy: so this should only ever happen when INFO is disabled and HELLO is not available, which is pre v6, I will test and verify

redisinsight/api/src/constants/error-messages.ts
```diff
@@ -59,6 +59,7 @@ export default {
     'Key with this name does not exist or does not have an associated timeout.',
   SERVER_NOT_AVAILABLE: 'Server is not available. Please try again later.',
   REDIS_CLOUD_FORBIDDEN: 'Error fetching account details.',
+  NO_INFO_COMMAND_PERMISSION: 'has no permissions to run the \'info\' command',
```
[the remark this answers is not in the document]
KrumTy: previous word was the username iirc

redisinsight/api/src/modules/redis/client/redis.client.ts
```diff
@@ -47,7 +51,7 @@ export enum RedisFeature {
 export abstract class RedisClient extends EventEmitter2 {
   public readonly id: string;
 
-  protected info: object;
+  protected info: any;
```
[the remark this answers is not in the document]
KrumTy: I'd say typing info should be done separately (out of scope) since there might be contract changes across the different versions
`any` because ts complains when trying to access 2+ level deep properties of the object

redisinsight/api/src/modules/cluster-monitor/cluster-monitor.service.ts
```diff
@@ -41,10 +41,7 @@ export class ClusterMonitorService {
         return Promise.reject(new BadRequestException('Current database is not in a cluster mode'));
       }
 
-      const info = convertRedisInfoReplyToObject(await client.sendCommand(
-        ['info', 'server'],
-        { replyEncoding: 'utf8' },
-      ) as string);
+      const info = await client.getInfo(true, 'server');
```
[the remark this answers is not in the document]
KrumTy: Why though? Preserving the behavior is one thing, but I figured querying just specific sections should be faster. E.g. if we have an interval, querying the data and only carrying about one section it would make sense to only query that section. I might be wrong though.

redisinsight/api/src/common/constants/general.ts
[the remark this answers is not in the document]
ArtemHoruzhenko: oh. forgot about it. don't waste your time. seems like edge case

redisinsight/api/src/modules/cluster-monitor/cluster-monitor.service.ts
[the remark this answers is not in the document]
ArtemHoruzhenko: I don't think this slightly impact performance. Just don't like idea of merging objects in `getInfo` funciton. 
If we want to stay with this logic we should enhance `getInfo` command to reply only particular section as it works now (most probably without storing it)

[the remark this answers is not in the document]
KrumTy: refactored; I've removed the stored info object

pawelangelow (approved):
Tested and it looks OK. 🚀 

Side note: I know that we merged few branches to this one, so this makes the whole change list bigger and it's hard to follow all the changes (big PR well known problem). I don't have perfect solution, but I'd rather prefer to merge them to main, than to a single branch that is merged back to main. (a.k.a. dont merge 3 > 2 > 1, but 1 to main, 2 to main and 3 to main). This way it's easier for me to see what happened.

Of course, the above is debatable, just saw it here and wanted to bring it up.

ArtemHoruzhenko (approved):
lgtm

PR #4378, opened 2025-02-17
Unifying local and CI runs for e2e test

katanaBoy:
- skip unstable tests
- add skipComments
- add no-sandbox execution for app image for local Ubuntu execution

### DIFF
--- a/tests/e2e/helpers/common.ts
+++ b/tests/e2e/helpers/common.ts
@@ -268,7 +268,6 @@ export class Common {
      * @param filePath Path to file
      */
     static async deleteFolderIfExists(filePath: string): Promise<void> {
-      if (fs.existsSync(filePath)) {
             fs.rm(filePath, {
                 recursive: true,
                 force: true
@@ -281,7 +280,6 @@ export class Common {
 
                 }
             });
-        }
     }
     /**
       * Read file from folder
--- a/tests/e2e/helpers/common.ts
+++ b/tests/e2e/helpers/common.ts
@@ -264,6 +264,26 @@ export class Common {
     }
 
     /**
+     * Delete folder
+     * @param filePath Path to file
+     */
+    static async deleteFolderIfExists(filePath: string): Promise<void> {
+      if (fs.existsSync(filePath)) {
+            fs.rm(filePath, {
+                recursive: true,
+                force: true
+            }, (error) => {
+                if (error) {
+                    console.log(error);
+                }
+                else {
+                    console.log(`Directory Deleted: ${filePath}`);
+
+                }
+            });
+        }
+    }
+    /**
       * Read file from folder
       * @param filePath Path to file
      */
--- a/tests/e2e/helpers/common.ts
+++ b/tests/e2e/helpers/common.ts
@@ -1,5 +1,7 @@
 import * as path from 'path';
 import * as fs from 'fs';
+import * as fsp from 'fs/promises';
+
 import { ClientFunction, RequestMock, t } from 'testcafe';
 import { Chance } from 'chance';
 import { apiUrl } from './conf';
@@ -268,19 +270,14 @@ export class Common {
      * @param filePath Path to file
      */
     static async deleteFolderIfExists(filePath: string): Promise<void> {
-            fs.rm(filePath, {
-                recursive: true,
-                force: true
-            }, (error) => {
-                if (error) {
-                    console.log(error);
-                }
-                else {
-                    console.log(`Directory Deleted: ${filePath}`);
-
-                }
-            });
+        try {
+            await fsp.rm(filePath, { recursive: true, force: true });
+            console.log(`Directory Deleted: ${filePath}`);
+        } catch (error) {
+            console.error(`Failed to delete directory: ${filePath}`, error);
+        }
     }
+
     /**
       * Read file from folder
       * @param filePath Path to file
--- a/tests/e2e/tests/electron/critical-path/a-first-start-form/autodiscovery.e2e.ts
+++ b/tests/e2e/tests/electron/critical-path/a-first-start-form/autodiscovery.e2e.ts
@@ -14,7 +14,7 @@ const databaseAPIRequests = new DatabaseAPIRequests();
 const standalonePorts = [8100, 8101, 8102, 8103, 12000];
 const otherPorts = [28100, 8200];
 
-//TODO Other tests are messing up the state of this test for now seams like deleting the folder  will do as workaround
+// TODO: Other tests are messing up the state of this test for now seams like deleting the folder  will do as workaround
 const cleanBeforeClientStarts = async () => {await Common.deleteFolderIfExists(workingDirectory)}
 cleanBeforeClientStarts().then(async () => {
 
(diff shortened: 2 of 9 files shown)


tests/e2e/helpers/common.ts
```diff
@@ -263,6 +263,26 @@ export class Common {
...
+                if (error) {
+                    console.log(error);
+                }
+                else {
+                    console.log(`Directory Deleted: ${filePath}`);
+
+                }
+            });
+        }
+    }
```
pawelangelow, line 285:
```suggestion
    static async deleteFolderIfExists(filePath: string): Promise<void> {
        try {
            await fs.rm(filePath, { recursive: true, force: true });
            console.log(`Directory Deleted: ${filePath}`);
        } catch (error) {
            console.error(`Failed to delete directory: ${filePath}`, error);
        }
    }
```

as per [documentation](https://nodejs.org/api/fs.html#fspromisesrmpath-options) `force: true` would prevent missing file exceptions (a.k.a. if file is there it will be deleted; if not - no error) so we can safely remove `fs.existsSync(filePath)`

tests/e2e/tests/electron/critical-path/a-first-start-form/autodiscovery.e2e.ts
```diff
@@ -11,12 +12,17 @@ const databaseAPIRequests = new DatabaseAPIRequests();
 const standalonePorts = [8100, 8101, 8102, 8103, 12000];
 const otherPorts = [28100, 8200];
 
+//TODO Other tests are messing up the state of this test for now seams like deleting the folder  will do as workaround
+const cleanBeforeClientStarts = async () => {await Common.deleteFolderIfExists(workingDirectory)}
+cleanBeforeClientStarts().then(() => {})
+
```
pawelangelow, line 18: isn't there a hook like `beforeAll` where we can plug this logic? Something like:
```
fixture `Autodiscovery`
    .meta({ type: 'critical_path', rte: rte.none })
    .page(commonUrl)
    .beforeAll(async () => Common.deleteFolderIfExists(workingDirectory))
    ....
```

tests/e2e/tests/electron/critical-path/a-first-start-form/autodiscovery.e2e.ts
```diff
@@ -11,12 +12,17 @@ const databaseAPIRequests = new DatabaseAPIRequests();
...
 fixture `Autodiscovery`
     .meta({ type: 'critical_path', rte: rte.none })
     .page(commonUrl)
-    .beforeEach(async() => {
-        await databaseHelper.acceptLicenseTerms();
+    .beforeEach(async(t) => {
+         await databaseHelper.acceptLicenseTerms();
```
pawelangelow, line 23:
```suggestion
        await databaseHelper.acceptLicenseTerms();
```
not sure tho, I just saw an empty space and I'm wondering if should be there; maybe it wasn't formatted OK in the first place

tests/e2e/tests/electron/critical-path/files-auto-update/enablement-area-autoupdate.e2e.ts
```diff
@@ -32,15 +32,15 @@ if (fs.existsSync(workingDirectory)) {
     tutorialsTimestampFile.save();
 
     fixture `Auto-update in Enablement Area`
-        .meta({ type: 'critical_path', rte: rte.standalone })
+        .meta({ type: 'critical_path', rte: rte.standalone, skipComment: "Skipped because it is not running in CI" })
```
pawelangelow, line 35: I'm wondering if we can add "automatic" skip when tests are running in CI. I was thinking for something like:
```
if (process.env.CI === "true") {
    console.log("Running in CI environment");
}

test.skipIfCI = process.env.CI ? test.skip : test;

test.skipIfCI("test something cool here", () => {
    expect(true).toBe(true);
});
```

maybe it's not for this PR, so treat it as just an idea. The problem that bugs me is that someone manually has to go and remove `.skip` when running these tests locally = 100% chance something would be forgotten.

tests/e2e/tests/electron/critical-path/a-first-start-form/autodiscovery.e2e.ts
```diff
@@ -11,12 +12,17 @@ const databaseAPIRequests = new DatabaseAPIRequests();
...
+//TODO Other tests are messing up the state of this test for now seams like deleting the folder  will do as workaround
+const cleanBeforeClientStarts = async () => {await Common.deleteFolderIfExists(workingDirectory)}
+cleanBeforeClientStarts().then(() => {})
+
 fixture `Autodiscovery`
     .meta({ type: 'critical_path', rte: rte.none })
     .page(commonUrl)
-    .beforeEach(async() => {
-        await databaseHelper.acceptLicenseTerms();
+    .beforeEach(async(t) => {
```
dantovska, line 22: where are we using the t parameter we are adding here? I see another one on the async on line 30 and its uses in the following function block.

katanaBoy: that's left from the  debugging, removed. Thanks

tests/e2e/tests/electron/critical-path/a-first-start-form/autodiscovery.e2e.ts
pawelangelow: Okay, so we should anyway rework it as implemented this way - it's not guaranteed that the fixture would start **after** the folder is deleted - maybe it works now, but in future this can lead to unexpected behavior and many WTFs per minute (see the attachment).

So what I'm trying to say is that:
```
async () => {
  await Common.deleteFolderIfExists(...);
  
  // this will start after delete is complete
  fixture``
  ....
}
```
ezgif-233939bcd62a57

katanaBoy: I already tried something similar 

const cleanBeforeClientStarts = async () => {await Common.deleteFolderIfExists(workingDirectory)}
cleanBeforeClientStarts().then(  () => {
    fixture `Autodiscovery`
. 
'
'
})
But the whole execution failed with ReferenceError: fixture is not defined

katanaBoy: Made it work with the  first suggestion.

I also tried adding await at the top level (and change the Tpescript config file) however this still fails with 
```
(node:378958) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
You cannot override the "target" compiler option in the TypeScript configuration file.

You cannot override the "module" compiler option in the TypeScript configuration file.

Cannot prepare tests due to the following error:

Error: TypeScript compilation failed.
.../a-first-start-form/autodiscovery.e2e.ts (20, 1): Top-level 'await' expressions are only allowed when the 'module' option is set to 'es2022', 'esnext', 'system', 'node16', or 'nodenext', and the 'target' option is set to 'es2017' or higher.

    at async ..../runner/bootstrapper.ts:311:25 {
  code: 'E1035',
  data: [
    'Error: TypeScript compilation failed.\n' +
      .../RedisInsight/tests/e2e/tests/electron/critical-path/a-first-start-form/autodiscovery.e2e.ts (20, 1): Top-level 'await' expressions are only allowed when the 'module' option is set to 'es2022', 'esnext', 'system', 'node16', or 'nodenext', and the 'target' option is set to 'es2017' or higher.\n"
  ]
}

Process finished with exit code 1
```

katanaBoy: I did try again with adding  ```import { fixture, test } from "testcafe";  const wrapper = async()=> { add all code here })

wrapper(); ```

but this  didn't because the test was failing again with a bizarre exception that can't find the window

pawelangelow (approved):

PR #4381, opened 2025-02-19
RI-6823 fix claim button

ArtemHoruzhenko:
It is required to be at least 1 consumer (not 2) after filtering consumers list


pd-redis (approved):
lgtm

PR #4382, opened 2025-02-19
[RI-6848] [RI-6842]: handle disabled INFO command for telemetry

KrumTy:
### Description

### Notes
- this PR addresses issues related to telemetry analytics when access to INFO is disabled
- requires https://github.com/RedisInsight/RedisInsight/pull/4377
- two failing integration tests in GET-databases-id-overview.test.ts that pass locally; needs investigating, could be branch target/github action related

### DIFF
--- a/redisinsight/api/src/modules/database/entities/database.entity.ts
+++ b/redisinsight/api/src/modules/database/entities/database.entity.ts
@@ -29,7 +29,6 @@ export enum HostingProvider {
   UPSTASH = 'UPSTASH',
   UNKNOWN_LOCALHOST = 'UNKNOWN_LOCALHOST',
   UNKNOWN = 'UNKNOWN',
-  INFO_COMMAND_IS_DISABLED = 'INFO_COMMAND_IS_DISABLED',
 }
 
 export enum ConnectionType {
--- a/redisinsight/api/src/modules/database/entities/database.entity.ts
+++ b/redisinsight/api/src/modules/database/entities/database.entity.ts
@@ -29,6 +29,7 @@ export enum HostingProvider {
   UPSTASH = 'UPSTASH',
   UNKNOWN_LOCALHOST = 'UNKNOWN_LOCALHOST',
   UNKNOWN = 'UNKNOWN',
+  INFO_COMMAND_IS_DISABLED = 'INFO_COMMAND_IS_DISABLED',
 }
 
 export enum ConnectionType {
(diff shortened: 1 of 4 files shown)


redisinsight/api/src/modules/database/entities/database.entity.ts
```diff
@@ -29,6 +29,7 @@ export enum HostingProvider {
   UPSTASH = 'UPSTASH',
   UNKNOWN_LOCALHOST = 'UNKNOWN_LOCALHOST',
   UNKNOWN = 'UNKNOWN',
+  INFO_COMMAND_IS_DISABLED = 'INFO_COMMAND_IS_DISABLED',
```
ArtemHoruzhenko, line 32: I do not think we can use this enum for such case

redisinsight/api/src/utils/hosting-provider-helper.ts
```diff
@@ -107,7 +108,9 @@ export const getHostingProvider = async (client: RedisClient, databaseHost: stri
         return HostingProvider.REDIS_COMMUNITY_EDITION;
       }
     } catch (e) {
-      // ignore error
+      if (e.message.includes(ERROR_MESSAGES.NO_INFO_COMMAND_PERMISSION)) {
+        return HostingProvider.INFO_COMMAND_IS_DISABLED;
+      }
```
ArtemHoruzhenko, line 113: I don't think that any error from try block is because of INFO command is disabled. Also we shouldn't force return result since there is other logic below.

ArtemHoruzhenko (commented):
I think that covering hosting provider is not needed at all. We should not fail when there are no `info` or/and `hello` commands available and continue processing.

redisinsight/api/src/modules/database/providers/database-overview.provider.ts
```diff
@@ -74,13 +76,14 @@ export class DatabaseOverviewProvider {
    */
   private async getNodeInfo(client: RedisClient) {
     const { host, port } = client.options;
+    let infoData;
+
+    try {
+      infoData = await this.databaseInfoProvider.getRedisInfo(client);
+    } catch (e) {}
```
dantovska, line 83: why not log the error here?

redisinsight/api/src/utils/hosting-provider-helper.ts
KrumTy: agreed, I noticed this as well
this is more of a draft PR currently, sorry for time wasting

PR #4383, opened 2025-02-19
RI-6844: hide db overview metrics when data is not available

KrumTy:
### Description

### Notes
- this PR addresses issues related to db overview metrics when access to INFO is disabled
- requires https://github.com/RedisInsight/RedisInsight/pull/4377


pd-redis (approved):
lgtm

PR #4384, opened 2025-02-20
RI-6707 move back patch-package to RI main part

ArtemHoruzhenko:
Apply latest changes to the patch itself
Fix docker build errors due to postinstall script

### DIFF
--- a/redisinsight/api/package.json
+++ b/redisinsight/api/package.json
@@ -9,6 +9,7 @@
     "url": "https://redis.com/redis-enterprise/redis-insight"
   },
   "scripts": {
+    "postinstall": "npx patch-package",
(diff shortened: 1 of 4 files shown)


pd-redis (approved):
lgtm

redisinsight/api/package.json
```diff
@@ -87,6 +89,7 @@
     "node-version-compare": "^1.0.3",
     "quicktype-core": "^23.0.116",
     "redis": "^4.6.10",
+    "redis-parser": "3.0.0",
```
pawelangelow, line 92: why is this changed?

ArtemHoruzhenko: Since we have patch for it we want to make sure that version is fixed

KrumTy (approved):

PR #4385, opened 2025-02-20
RI-6775: Change key not found error message on delete

dantovska:
When delete is called on a key that has already been deleted, the error message should be "Key with that name does not exist".

e.g. key had expired after its details have been shown, and user can still click on the delete button.


pd-redis (approved):
looks good. The description of the PR is kind of confusing :)

dantovska:
> looks good. The description of the PR is kind of confusing :)

Thanks @pd-redis, I tried to simplify it a little bit.

PR #4387, opened 2025-02-21
RI-6806: Disable refresh when index not selected

dantovska:
When in Redisearch / index search mode and index is not selected, an error is shown to the user which is not a good UX experience. Instead, the following was applied:
* Disabled the refresh button when index is not selected in redisearch mode, which stops the refresh / data init from happening
* Add a tooltip on the disabled refresh button - "Select an index to refresh keys."
* Add some styling to the arrow next to the refresh button so it is clear that is disabled when hovered.


KrumTy (approved):

pd-redis (approved):
lgtm

PR #4388, opened 2025-02-21
RI-6589: remove circleci

pd-redis:
Removed all references that I found of circleci, ran most of the build scripts in packacge.json, no issues noticed


pawelangelow (approved):
There is one certain way that we'll know if we needed something from here 😄 🚀

PR #4389, opened 2025-02-21
RI-6382: Test multiple source connections

pawelangelow:
As additional work to #4368, here you can test multiple sources.

Maybe going commit by commit would be easier to review.

# Preview

Screenshot 2025-02-21 at 15 23 40
Screenshot 2025-02-21 at 15 23 25
Screenshot 2025-02-21 at 15 23 14

### DIFF
--- a/redisinsight/api/src/common/decorators/transform-to-map.decorator.ts
+++ b/redisinsight/api/src/common/decorators/transform-to-map.decorator.ts
@@ -1,0 +1,10 @@
+import { Transform, plainToClass } from 'class-transformer';
+
+export function TransformToMap<T>(cls: new (...args: any[]) => T) {
+  return Transform(({ value }) => Object.fromEntries(
+    Object.entries(value || {}).map(([key, val]) => [
+      key,
+      plainToClass(cls, val),
+    ]),
+  ));
+}
--- a/redisinsight/api/src/common/decorators/transform-to-map.decorator.ts
+++ b/redisinsight/api/src/common/decorators/transform-to-map.decorator.ts
@@ -4,30 +4,36 @@ import { ClassType } from 'class-transformer/ClassTransformer';
 
 export function TransformToMap<T>(targetClass: ClassType<T>) {
   return applyDecorators(
-    Transform((value) => {
-      if (!value) {
-        return value;
-      }
+    Transform(
+      (value) => {
+        if (!value) {
+          return value;
+        }
 
-      return Object.fromEntries(
-        Object.entries(value).map(([key, val]) => [
-          key,
-          plainToClass(targetClass, val),
-        ]),
-      );
-    }),
+        return Object.fromEntries(
+          Object.entries(value).map(([key, val]) => [
+            key,
+            plainToClass(targetClass, val),
+          ]),
+        );
+      },
+      { toClassOnly: true },
+    ),
 
-    Transform((value) => {
-      if (!value) {
-        return value;
-      }
+    Transform(
+      (value) => {
+        if (!value) {
+          return value;
+        }
 
-      return Object.fromEntries(
-        Object.entries(value).map(([key, instance]) => [
-          key,
-          classToPlain(instance),
-        ]),
-      );
-    }),
+        return Object.fromEntries(
+          Object.entries(value).map(([key, instance]) => [
+            key,
+            classToPlain(instance),
+          ]),
+        );
+      },
+      { toPlainOnly: true },
+    ),
   );
 }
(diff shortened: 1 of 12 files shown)


redisinsight/api/src/common/decorators/transform-to-map.decorator.ts
```diff
@@ -0,0 +1,10 @@
+import { Transform, plainToClass } from 'class-transformer';
+
+export function TransformToMap<T>(cls: new (...args: any[]) => T) {
+  return Transform(({ value }) => Object.fromEntries(
+    Object.entries(value || {}).map(([key, val]) => [
+      key,
+      plainToClass(cls, val),
+    ]),
+  ));
+}
```
pawelangelow, line 10: @ArtemHoruzhenko you suggested this in https://github.com/RedisInsight/RedisInsight/pull/4368#discussion_r1957737684, so I'd appreciate if you take a look

pawelangelow: I've tried to add tests, but it looks like `class-transfomer` has to be updated.

This is the setup:
```
class DummyClass {
  @Expose()
  value: string;
}

class TestDto {
  @TransformToMap(DummyClass)
  @Expose()
  data: Record<string, DummyClass>;
}

it('should transform each map value into an instance of DummyClass', () => {
  const input = {
    data: {
      key1: { value: 'test1' },
      key2: { value: 'test2' },
    },
  };

  const instance = plainToInstance(TestDto, input);
  console.log('instance :>> ', instance);
});
```
With current version (`0.3.1`), where `plainToInstance` is `plainToClass` I'm getting:
```
instance :>>  TestDto { data: undefined }
```

With latest version (`0.5.1`), where `plainToInstance` exists I'm getting:
```
instance :>>  TestDto {
  data: {
    key1: DummyClass { value: 'test1' },
    key2: DummyClass { value: 'test2' }
  }
}
```

pawelangelow: You can see the implementation here: https://github.com/RedisInsight/RedisInsight/pull/4389/commits/b6d31361065c6a6abb77b0d2bf83c5dcd45395e9

redisinsight/api/src/common/decorators/transform-to-map.decorator.ts
```diff
@@ -0,0 +1,15 @@
+import { Transform, plainToClass } from 'class-transformer';
+
+export function TransformToMap<T>(cls: new (...args: any[]) => T) {
+  return Transform(({ value }) => {
```
KrumTy, line 4: @pawelangelow your tests should pass with:
```suggestion
  return Transform((value) => {
```

ArtemHoruzhenko: I mean we must consider bidirectional transformation (to class and to plain). You can find example here: https://github.com/RedisInsight/RedisInsight/blob/b2d53dc903cb78cc4093b6456b64049e06305143/redisinsight/api/src/common/decorators/object-as-map.decorator.ts

pawelangelow: Thanks for the feedback, both of you! So, here's the [commit](https://github.com/RedisInsight/RedisInsight/pull/4389/commits/fa4b3017779ed4e75ea9dfc6a04d98166fc5b035).

redisinsight/api/src/modules/rdi/client/api.rdi.client.ts
```diff
@@ -219,12 +197,19 @@ export class ApiRdiClient extends RdiClient {
...
+
+      if (sourceConfigs.length) {
+        await Promise.all(
+          sourceConfigs.map(async (source) => {
+            const response = await this.client.post(
+              RdiUrl.TestSourcesConnections,
+              { ...config.sources[source] },
+            );
+            sources[source] = response.data;
+          }),
```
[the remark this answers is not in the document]
pawelangelow: Yes, my idea is to run all the calls to the RDI simultaneously. This is a strange implementation in RDI, but I hope they'll fix it in future - the endpoint to test the connection accepts just one `source`, so that's why we do N calls (N = number of sources). Honestly, I don't have observation how many sources potentially user could have, but the `Promise.all` options looks like the robust way in this situation. (e.g. if user has 1-2 sources, probably it's not going to feel bad if we don't run all the requests simultaneously, but if they have 10, it will)

redisinsight/api/src/common/decorators/transform-to-map.decorator.ts
KrumTy: @pawelangelow don't you need to include [these](https://github.com/RedisInsight/RedisInsight/blob/b2d53dc903cb78cc4093b6456b64049e06305143/redisinsight/api/src/common/decorators/object-as-map.decorator.ts#L22) transform options as well?

pawelangelow: Yes! I don't know but kind of expected the transformation relies on order (forward is the first transformation, backwards is the second one). 😄

KrumTy (approved):
tested and verified, LGTM

PR #4390, opened 2025-02-21
RI-6848: add info_command_is_disabled to analytics when INFO is disabled via ACL

KrumTy:


redisinsight/api/src/modules/database/database-connection.service.ts
```diff
@@ -75,20 +76,23 @@ export class DatabaseConnectionService {
...
     }
 
-    this.collectClientInfo(clientMetadata, client, generalInfo?.version);
+    this.collectClientInfo(clientMetadata, client, generalInfo);
 
     this.logger.debug(`Succeed to connect to database ${clientMetadata.databaseId}`, clientMetadata);
   }
 
-  private async collectClientInfo(clientMetadata: ClientMetadata, client: RedisClient, version?: string) {
+  private async collectClientInfo(clientMetadata: ClientMetadata, client: RedisClient, generalInfo: RedisDatabaseInfoResponse) {
```
ArtemHoruzhenko, line 84: Let's keep function signature why we need to change it?

redisinsight/api/src/modules/database/database-connection.service.ts
```diff
@@ -75,20 +76,23 @@ export class DatabaseConnectionService {
...
+      const infoCommandIsDisabled = generalInfo?.server?.info_command_is_disabled;
       const intVersion = parseInt(version, 10) || 0;
       const clients = await this.databaseInfoProvider.getClientListInfo(client) || [];
 
       this.analytics.sendDatabaseConnectedClientListEvent(
         clientMetadata.sessionMetadata,
-        clientMetadata.databaseId,
         {
+          databaseId: clientMetadata.databaseId,
+          ...(infoCommandIsDisabled ? { info_command_is_disabled: true } : {}),
```
ArtemHoruzhenko, line 95: if we will add `isInfoCommandDisabled` flag into `RedisClient` we can then avoid adding custom logic here and slightly simplify this. It become ```{ isInfoComaandDisabled: client.isInfoCommandDisabled }``` or something similar

redisinsight/api/src/modules/redis/client/redis.client.ts
```diff
@@ -203,6 +203,7 @@ export abstract class RedisClient extends EventEmitter2 {
         server_name: helloResponse.server,
         redis_version: helloResponse.version,
         redis_mode: helloResponse.mode,
+        info_command_is_disabled: true,
```
ArtemHoruzhenko, line 206: Let's make it a RedisClient property? And update each time right here (set to false or set to true when `info` command respond without error)

redisinsight/api/src/modules/database/database-connection.service.ts
KrumTy: updated, however I pass no key/value in case of false

redisinsight/api/src/modules/redis/client/redis.client.ts
```diff
@@ -182,6 +183,7 @@ export abstract class RedisClient extends EventEmitter2 {
       if (error.message.includes(ERROR_MESSAGES.NO_INFO_COMMAND_PERMISSION)) {
         try {
           // Fallback to getting basic information from `hello` command
+          this.isInfoCommandDisabled = true;
```
ArtemHoruzhenko, line 186: Seems like you forgot to add `this.isInfoCommandDisabled = true;` into try black when info command wasn't fail.

ArtemHoruzhenko: `this.isInfoCommandDisabled = false` I mean

KrumTy: updated default value and made it boolean; currently there's no need of a 3rd value

PR #4396, opened 2025-02-25
RI-6425: generate big data

pd-redis:
For some tests we need very big databases, since import in the cloud does not work well, this PR is adding generating and sending commands to appropriately sized DB

### DIFF
--- a/tests/e2e/helpers/keys.ts
+++ b/tests/e2e/helpers/keys.ts
@@ -6,6 +6,7 @@ import { BrowserPage } from '../pageObjects';
 import { KeyData, AddKeyArguments } from '../pageObjects/browser-page';
 import { COMMANDS_TO_CREATE_KEY, KeyTypesTexts } from './constants';
 import { Common } from './common';
+import { populateBigKeys, populateDb } from './scripts/generate-big-data';
 
 const browserPage = new BrowserPage();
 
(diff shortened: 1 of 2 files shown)


tests/e2e/helpers/scripts/generate-big-data.ts
```diff
@@ -0,0 +1,304 @@
...
+): Promise<void> => {
+
+  try {
+    console.log('Starting...');
+    client.on('error', err => console.log('Redis Client Error', err));
+    let executions = 0;
+    await client.connect();
+    for (let bk of baseKeys) {
+      const generator = generateBigData(bk, separatorPrimary, mainKeysLimit);
+      for (const commands of generator) {
```
pawelangelow, line 263: why `generateBigData` has to be generator?

tests/e2e/helpers/keys.ts
```diff
@@ -298,6 +299,32 @@ export async function deleteAllKeysFromDB(host: string, port: string): Promise<v
     }
 }
 
+export async function populateBigData(host: string, port: string): Promise<void> {
```
KrumTy, line 302: jsdoc for this function would be appreciated

tests/e2e/helpers/scripts/generate-big-data.ts
pd-redis: because it could potentially generate huge amout of data and cause out of memory issue

pawelangelow: Okay, so that's what I thought. What I was thinking when I saw the implementation of the generator was why do we batch them in `commands.push(command);` and not yield the command there and batch them here in `populateDb`.

I haven't tested it tho - but I was thinking it will be more memory efficient if every command is yielded and processed. Kind of looks the same if we batch them here in `populateDb`, but IMO it's better to do it here as the generator will stop yielding if we're waiting on redis to acknowledge the data. If we keep the batch inside the generator - the generator will hold the data.

Probably it's same, but i wanted to discuss it.

pawelangelow (approved):
My comment isn't blocking, so I'm approving it. Nice!

tests/e2e/helpers/scripts/generate-big-data.ts
pd-redis: could make some difference in rare cases, but will require adding batching logic here as well, and that will add some duplication of code

PR #4401, opened 2025-02-27
RI-6841: Add close button for unsupported key types

dantovska:
For some keys, their details page does not contain the keys header so the close button is not there.
in this PR such close button is added so the details panel could be closed. 
Types affected:
- Unsupported key types - unexpected key types
- Module type details - excpected types, but no details representation (example - TimeSeries)
- Keys with long key names - contain prefix [Truncated due to length]

Changes applied:
- Created a wrapper component to wrap all three detail components which contains the close button
- Removed all repetitive styling and other logic from the three detail components since it is moved to the wrapper

Side fixes:
- Remove the key from onClose button, as it is not used
- Make disabledRefreshMessage optional


redisinsight/ui/src/pages/browser/modules/key-details/components/modules-type-details/ModulesTypeDetails.tsx
```diff
@@ -1,14 +1,22 @@
...
 
-const ModulesTypeDetails = ({ moduleName = 'unsupported' }: { moduleName: string }) => {
+const ModulesTypeDetails = (
+  { moduleName = 'unsupported', 
+    onClose 
+  }: 
+    { moduleName: string, 
+      onClose: () => void,
+    }
+  ) => {
```
KrumTy, line 19: for component props I usually prefer extracting their type in `{$componentName}Props` type such as:
```suggestion
type ModulesTypeDetailsProps = {
  moduleName: string, 
  onClose: () => void,
}
const ModulesTypeDetails = ({ moduleName = 'unsupported', onClose }: ModulesTypeDetailsProps) => {
```
that way I think it's easier to read, extend, export the type, etc

redisinsight/ui/src/pages/browser/modules/key-details-header/KeyDetailsHeader.tsx
```diff
@@ -36,7 +36,7 @@ import { KeyDetailsHeaderSizeLength } from './components/key-details-header-size
 import styles from './styles.module.scss'
 
 export interface KeyDetailsHeaderProps {
-  onCloseKey: (key: RedisResponseBuffer) => void
+  onCloseKey: () => void
```
KrumTy, line 39: good catch, if this really is not used

KrumTy (approved):
overall LGTM

pd-redis (approved):

PR #4410, opened 2025-03-06
Ri 6866 - autodiscovery flow and establishing a connection to an azure based redis instance

KIvanow:
This is a follow up of [RI-6621 - Implement new auth flow with node-redis](https://github.com/RedisInsight/RedisInsight/pull/4373)

Due to some changes in the design it actually takes care of the 2nd and 3rd step of the process.

The main changes are:

1. The auth flow is separated from the regular social logins, since it triggers from a different location and allows for the user to still be separately signed into our Redis Cloud.
2. Noticed multiple repeated components - [AnalyticsPageRouter.tsx](https://github.com/RedisInsight/RedisInsight/blob/main/redisinsight/ui/src/pages/analytics/AnalyticsPageRouter.tsx), [SentinelPage.tsx](https://github.com/RedisInsight/RedisInsight/blob/main/redisinsight/ui/src/pages/autodiscover-sentinel/sentinel/SentinelPage.tsx), [RedisCloudPage.tsx](https://github.com/RedisInsight/RedisInsight/blob/main/redisinsight/ui/src/pages/autodiscover-cloud/redis-cloud/RedisCloudPage.tsx), [InstancePageRouter.tsx](https://github.com/RedisInsight/RedisInsight/blob/main/redisinsight/ui/src/pages/instance/InstancePageRouter.tsx)
With a slight difference only for the last one. These are replaced by a new common component to avoid continuing to copy paste them
3. The way the resources and databases need to be visualized is almost identical to the way the cloud ones currently are. There are 2 new common components for listing resources and databases that are used by both the cloud autodiscovery flow, and the azure auto discovery flow

There are a few hardcoded places with a TODO statement indicating this isn't the final flow, it is simplified for testing purposes

- [x] Initial authentication flow implementation with the new node-redis library. This includes the file structure changes and the basic token flow as recommended and described [here](https://github.com/redis/node-redis/blob/master/packages/entraid/README.md)
Please focus and provide feedback on just these
Details on how to trigger the changes and see them in action will be provided separately in a different channel
- [x] Implement the newly updated designs to trigger the basic auth flow introduced in this PR and add a connection to an AMR (azure managed redis) with the data received from this implementation
- [x] Add the auto discovery flow. Both the UI and BE part.
- [ ] Add logic for storing the connections per database (current product requirement)
- [ ] Add edit for the stored database's connection
- [ ] Add the user flow to be triggered whenever a token expires
- [ ] Add tests for all of the changes
- [ ] Add telemetry support - _TODO: Discuss with Viktar regarding requests_
- [ ] Final review and polishing
  - [ ] Handling of no subcriptions / no database screens

_I swear the next few PRs will be smaller, now that the main functionality is there and working_

### DIFF
--- a/redisinsight/api/src/modules/azure/client/api.azure.client.ts
+++ b/redisinsight/api/src/modules/azure/client/api.azure.client.ts
@@ -79,7 +79,7 @@ export class ApiAzureClient extends AzureClient {
 
   private wrapAzureError(error: any): Error {
     if (error.response) {
-      const { status, data } = error.response as { status: number data: AzureErrorResponse }
+      const { status, data } = error.response as { status: number, data: AzureErrorResponse }
       return new Error(`Azure API error (${status}): ${JSON.stringify(data)}`)
     }
 


redisinsight/api/src/modules/auth/microsoft-auth/microsoft-azure-auth.service.ts
```diff
@@ -2,12 +2,14 @@ import { Injectable, Logger } from '@nestjs/common';
...
 import config from 'src/utils/config';
-import { MicrosoftAuthSession } from './models/microsoft-auth-session.model';
 import { EntraidCredentialsProvider } from '@redis/entraid/dist/lib/entraid-credentials-provider';
+import { CloudAuthStatus } from 'src/modules/cloud/auth/models/cloud-auth-response';
 
 const { idp: { microsoft: idpConfig } } = config.get('cloud');
 
 interface MicrosoftAuthOptions {
+    action: any;
+    callback: any;
```
KrumTy, line 12: can these be typed?

redisinsight/api/src/modules/auth/microsoft-auth/microsoft-azure-auth.service.ts
```diff
@@ -107,24 +119,37 @@ export class MicrosoftAuthService {
...
+                password: '',
+            };
+        }
+
+        try {
+            if (authRequest?.callback) {
+                authRequest.callback(result)?.catch?.((e: Error) =>
+                    this.logger.error('Async callback failed', e)
+                );
+            }
```
KrumTy, line 146: just doublechecking what this callback is about and if it's something that needs to be awaited

redisinsight/api/src/modules/auth/microsoft-auth/microsoft-azure-auth.service.ts
```diff
@@ -135,13 +160,34 @@ export class MicrosoftAuthService {
         this.inProgressRequests.delete(query?.state);
     }
 
-    async getSession(id: string): Promise<MicrosoftAuthSession> {
+    async getSession(): Promise<any> {
```
KrumTy, line 163: why back to any? Would `MicrosoftAuthSession | null | undefined` work?

redisinsight/api/src/modules/azure/autodiscovery/azure-autodiscovery.service.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  async getSubscriptions() {
+    try {
+      const accessToken = await this.microsoftAuthService.getAccessToken()
+      if (!accessToken) {
+        throw new UnauthorizedException('No Azure access token available')
+      }
+
+      const response = await axios.get(
+        'https://management.azure.com/subscriptions?api-version=2024-08-01',
```
KrumTy, line 21: perhaps move this to a constant, bellow I see api-version 2020

redisinsight/ui/src/components/databases-list/DatabasesBase.tsx
```diff
@@ -0,0 +1,324 @@
...
+  const BackButton = () => (
+    <EuiButton
+      onClick={onBack}
+      color="secondary"
+      className="btn-back"
+      data-testid="btn-back"
+    >
+      {backButtonText}
+    </EuiButton>
+  )
```
KrumTy, line 215: declaring components inside another components is usually inefficient (harder to understand and test too). I'd suggest extracting these and pass the related props. Also that way they can be easily memoized if their parent rerenders often but their props don't change.
PS maybe some can be reused in [SubscriptionsBase.tsx](https://github.com/RedisInsight/RedisInsight/pull/4410/files#diff-b9f1ca6d6310f36d9b77e2eeb991a4a71a344eaf8e7906727d07d25a84e8f3dd)

redisinsight/api/src/modules/azure/autodiscovery/azure-autodiscovery.controller.ts
```diff
@@ -0,0 +1,52 @@
...
+      );
+
+      const databasesArrays = await Promise.all(databasePromises);
+
+      return databasesArrays.flat();
+    } catch (error) {
+      this.azureAutodiscoveryService.logger.error('Failed to fetch databases from multiple subscriptions', error);
+      throw error;
+    }
+  }
```
ArtemHoruzhenko, line 51: we should use DTOs and validate input data.
business logic  should be part of service. we are trying to keep controllers very tiny.

redisinsight/api/src/modules/auth/microsoft-auth/microsoft-azure-auth.service.ts
KIvanow: Updated

redisinsight/api/src/modules/auth/microsoft-auth/microsoft-azure-auth.service.ts
KIvanow: Updated

redisinsight/api/src/modules/auth/microsoft-auth/microsoft-azure-auth.service.ts
KIvanow: Updated

redisinsight/api/src/modules/azure/autodiscovery/azure-autodiscovery.service.ts
KIvanow: That was great suggestion, thank you! Updated it

redisinsight/ui/src/components/databases-list/DatabasesBase.tsx
KIvanow: Updated. It doesn't make sense to keep such small buttons as separate components. Even in the refactored base components it doesn't work well and they have some small differences. Reworked it to be part of the return/render part directly

redisinsight/api/src/modules/azure/autodiscovery/azure-autodiscovery.service.ts
```diff
@@ -0,0 +1,94 @@
...
+        tenantId: sub.tenantId,
+        isActive: sub.state === 'Enabled',
+      }))
+    } catch (error) {
+      this.logger.error('Failed to fetch Azure subscriptions', error)
+      throw error
+    }
+  }
+
+  async getDatabases(subscriptionId: string) {
```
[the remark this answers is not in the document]
KIvanow: Updated it to be typed. Still thinking, playing around with the models and how that data will need to be stored/retrieve

redisinsight/api/src/modules/azure/autodiscovery/azure-autodiscovery.service.ts
```diff
@@ -0,0 +1,94 @@
...
+      if (!accessToken) {
+        throw new UnauthorizedException('No Azure access token available')
+      }
+
+      const response = await axios.get(
+        'https://management.azure.com/subscriptions?api-version=2024-08-01',
+        {
+          headers: {
+            Authorization: `Bearer ${accessToken}`,
+          },
```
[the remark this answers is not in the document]
KIvanow: reworked as part of the client refactoring (suggested earlier in the comments)

redisinsight/ui/src/slices/instances/instances.ts
```diff
@@ -393,8 +394,11 @@ export function createInstanceStandaloneAction(
 
       dispatch(defaultInstanceChangingFailure(errorMessage))
 
-      if (error?.response?.data?.error === ApiErrors.SentinelParamsRequired) {
-        checkoutToSentinelFlow(payload, dispatch, onRedirectToSentinel)
+      if (
+        error?.response?.data?.error === ApiErrors.SentinelParamsRequired
+        && !options?.isAzureInstance
```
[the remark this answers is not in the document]
KIvanow: Why? The goal was to reuse the connection function. The added boolean property check is to block the sentinel flow, as it isn't applicable

redisinsight/api/src/modules/azure/autodiscovery/azure-autodiscovery.controller.ts
KIvanow: You are absolutely correct. Updated it.

PR #4413, opened 2025-03-07
Ri 6645 web e2e tests runs

katanaBoy:
- Desktop tests running  back in CI
- Web/Docker  tests running  in CI
- Skip some unstable tests 
- Modify some parts that were causing test runs to stall


.github/e2e/test.app-image.sh
```diff
@@ -1,26 +1,33 @@
...
 # run rte
-#docker compose -f tests/e2e/rte.docker-compose.yml build
-#docker compose -f tests/e2e/rte.docker-compose.yml up --force-recreate -d -V
-#./tests/e2e/wait-for-redis.sh localhost 12000 && \
+docker compose -f tests/e2e/rte.docker-compose.yml build
+docker compose -f tests/e2e/rte.docker-compose.yml up --force-recreate -d -V
+./tests/e2e/wait-for-redis.sh localhost 12000 && \
 
-# run tests
+# run tests add DEBUG=testcafe:* to debug framework execution
```
pawelangelow, line 29: I'd leave it where it was, so something like:
```
# DEBUG=testcafe:* # <- to debug framework execution
```
the idea is to make it very useful to run. Because, you know where to put it, but if someone runs it for first time, maybe won't get very fast to where this line should be.

Even more robust would be (maybe not for this PR, just an idea)
```
DEBUG=0 # <- set this to 1 if you want to run the debug mode

[ "$DEBUG" -eq 1 ] && DEBUG=testcafe:*
```
or something like this.

tests/e2e/web.runner.ts
```diff
@@ -11,7 +11,7 @@ import testcafe from 'testcafe';
                         experimentalDecorators: true
                     } })
                 .src((process.env.TEST_FILES || 'tests/web/**/*.e2e.ts').split('\n'))
-                .browsers(['chrome --cache --allow-insecure-localhost --disable-search-engine-choice-screen --ignore-certificate-errors'])
+                .browsers(['chromium --disable-search-engine-choice-screen --ignore-certificate-errors --disable-dev-shm-usage --no-sandbox'])
```
pawelangelow, line 14: Why we switched from `chrome` to `chromium`?

tests/e2e/tests/web/regression/browser/hash-field.e2e.ts
```diff
@@ -41,4 +41,4 @@ test('Verify that user can search per exact field name in Hash in DB with 1 mill
     // Check the search result
     const result = await browserPage.hashFieldsList.nth(0).textContent;
     await t.expect(result).eql(fieldForSearch, 'Hash field not found');
-});
+});//.skip.meta({skipComment: " "});
```
pawelangelow, line 44: is this a leftover?

tests/e2e/helpers/database-scripts.ts
```diff
@@ -10,77 +11,119 @@ export class DatabaseScripts {
...
+
+        const query = `SELECT ${dbTableParameters.columnName}
+                       FROM ${dbTableParameters.tableName}
+                       WHERE ${dbTableParameters.conditionWhereColumnName} = ?`;
+        try {
+            const getAsync = (query: string, p: (string | number | undefined)[]) => promisify(db.get.bind(db));
+            const row = await Promise.race([
+                getAsync(query, [dbTableParameters.conditionWhereColumnValue]),
+                new Promise<any>((_, reject) =>
+                    setTimeout(() => reject(new Error('Query timed out after 10 seconds')), 10000)
```
pawelangelow, line 60: This repeats through the tests. Maybe we need to extract it as a helper:
```
const createTimeout = (ms: number) => new Promise ....

Promise.race([dbCall, createTimeout(10_000)]
```

tests/e2e/web.runner.ts
katanaBoy: My reasoning is that  the running them locally  should be close to how they run in the CI. Atm they are running at the CI at least till we see consistent runs

tests/e2e/tests/web/regression/browser/hash-field.e2e.ts
katanaBoy: yeah this one was failing at some point and I was planning to skip it too thanks

.github/e2e/test.app-image.sh
katanaBoy: Though this information is also in Conflunce it makes sense to have it in the way that you described. 
I'v made the change. Initially I went what I thought is more sensible approach 
```
if [ "$DEBUG" = "1" ]; then
  DEBUG_FLAG="DEBUG=testcafe:*"
else
  DEBUG_FLAG=""
fi

COMMON_URL=$(tail -n 1 apppath)/resources/app.asar/dist/renderer/index.html \
ELECTRON_PATH=$(tail -n 1 apppath)/redisinsight \
RI_SOCKETS_CORS=true \
$DEBUG_FLAG \
yarn --cwd tests/e2e dotenv -e .desktop.env yarn --cwd tests/e2e test:desktop:ci
```

however this produced ./.github/e2e/test.app-image.sh: line 37: DEBUG=testcafe:*: command not found

so I got it working in the way that is at the momment

.github/e2e/test.app-image.sh
```diff
@@ -1,26 +1,42 @@
...
+  ELECTRON_PATH=$(tail -n 1 apppath)/redisinsight \
+  RI_SOCKETS_CORS=true \
+  DEBUG=testcafe:* \
+  yarn --cwd tests/e2e dotenv -e .desktop.env yarn --cwd tests/e2e test:desktop:ci
+else
+  COMMON_URL=$(tail -n 1 apppath)/resources/app.asar/dist/renderer/index.html \
+  ELECTRON_PATH=$(tail -n 1 apppath)/redisinsight \
+  RI_SOCKETS_CORS=true \
+  yarn --cwd tests/e2e dotenv -e .desktop.env yarn --cwd tests/e2e test:desktop:ci
+fi
```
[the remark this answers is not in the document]
katanaBoy: It didn't work in this form, but exporting the variables did it.

tests/e2e/helpers/database-scripts.ts
katanaBoy: Extracted in utils file for now

.github/e2e/test.app-image.sh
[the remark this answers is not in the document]
pawelangelow: Yeah, it was kind of pseudo code, as I haven't run it, so take it with pinch of salt :D Super!

tests/e2e/web.runner.ts
pawelangelow: Sure! It makes sense. I don't know if we have some documentation for these tests but it won't hurt if we have `You need chromium installed in order to run the tests` or something in this manner. It would be nice if we use the same version as well.

pawelangelow (approved):

PR #4418, opened 2025-03-10
RI-3619: Add database key name formatting

pawelangelow:
## Description

Sometimes, Unicode keys may contain symbols that **do not render well**, such as **empty spaces, null characters, line separators, or carriage returns**. To enhance usability, we allow users to **render key names in their HEX representation**. This ensures that all characters are visible, giving users more control over how keys are displayed.

### Limitations

One limitation of this approach is that the **Tree View** becomes **obsolete**.

#### Why?

The **Tree View** relies on delimiters to structure keys. For example, if we have a key structured as:

```
root:folder:key1
```

and split it by `:`, we get:

```
- root
  - folder
    - key1
```

However, in **HEX format**, the same key would look like:

```
726F6F743A666F6C6465723A6B657931
```

Since there is no **clear delimiter** in the HEX representation, it is **not possible to split** the key into a hierarchical structure.


redisinsight/api/migration/index.ts
```diff
@@ -48,6 +48,7 @@ import { DatabaseCreatedAt1733740794737 } from './1733740794737-database-created
 import { DbSettings1737362130798 } from './1737362130798-db-settings';
 import { DatabaseForceStandalone1738829743482 } from './1738829743482-database-forceStandalone';
 import { RdiOptionalAuth1740579711635 } from './1740579711635-rdi-optional-auth';
+import {KeyNameFormatAdded1741621786907} from './1741621786907-key-name-format'
```
ArtemHoruzhenko, line 51: seems like this file is not formatted.

ArtemHoruzhenko (dismissed):
overall looks good

ArtemHoruzhenko (approved):
lgtm

ArtemHoruzhenko (approved):

PR #4423, opened 2025-03-14
FIX: typescript type fail

pd-redis:


katanaBoy (approved):
Looks good to me 
Tests running https://github.com/RedisInsight/RedisInsight/actions/runs/13852613383/job/38762844722

PR #4433, opened 2025-03-19
save and load theme using settings api

yashschandra:


yashschandra:
resolves https://github.com/RedisInsight/RedisInsight/issues/4432

redisinsight/ui/src/services/theme.ts
```diff
@@ -24,7 +24,7 @@ class ThemeService {
 
     document.adoptedStyleSheets = [sheet]
 
-    localStorageService.set(BrowserStorageItem.theme, actualTheme)
```
pawelangelow, line 27: I've tested both desktop + docker image, and IMO this is the only change we need. Can you:
- verify it suits your needs
- leave just this change

yashschandra: The problem if we keep just this change is that my settings will not be saved anywhere so if I open new Redisinsight window (eg. chrome incognito) it will be opened with default theme ie. `System Theme`. Added videos for before and after including the (complete) changes.

before - 

https://github.com/user-attachments/assets/108e48f6-fac7-463f-937f-8b56a0b60c07

at 7s I changed the theme to `Dark`, at 15s the theme is again light (`System`)

after - 

https://github.com/user-attachments/assets/8ecec791-f46f-4156-88c8-e6988f4045d2

at 5s I changed the theme to `Dark`, at 15s the theme is `Dark` (fetched from server)

Thats why other changes are necessary for complete solution.

yashschandra: @pawelangelow shall I mark this conversation as resolved?

pawelangelow: I'll double-check it. Just to confirm, we want to cover the following scenarios:

- When the system theme changes and the application follows it, the app should respond accordingly (e.g., from dark to light mode).

- When the user changes the theme manually, the choice should be persistently saved.

yashschandra: correct

redisinsight/ui/src/pages/settings/SettingsPage.tsx
```diff
@@ -62,12 +61,20 @@ const SettingsPage = () => {
     })
   }, [])
 
+  useEffect(() => {
+    if (config) {
+      setThemeValue(config.theme);
+      theme = config.theme;
+    }
+  }, [config])
+
```
pawelangelow, line 70: Let's extract the entire theme setup into a separate component, similar to `<ConsentsPrivacy />` or `<WorkbenchSettings />`. This will help keep the SettingsPage cleaner and may also reduce unnecessary re-renders.

redisinsight/ui/src/pages/settings/SettingsPage.tsx
```diff
@@ -36,19 +37,17 @@ import styles from './styles.module.scss'
 
 const SettingsPage = () => {
   const [loading, setLoading] = useState(false)
+  const [themeValue, setThemeValue] = useState('')
```
pawelangelow, line 40:
```suggestion
  const [selectedTheme, setSelectedTheme] = useState('')
```

pawelangelow (changes requested):
**First of all, thanks for your work!**
It's always great to see people engage in the process with fresh ideas and perspectives - I really appreciate it. 🙏 🍺 

**Second, apologies for the delay.**

**Third, now that the requirements are clear, the changes make sense and look good overall.** 🚀 
I just have a few comments related to code style. Here's what I suggest:

- Rebase onto the latest main
- Go through the comments
- Add some unit tests

I'm happy to help with any (or all) of these steps - just let me know!

yashschandra:
> **First of all, thanks for your work!** It's always great to see people engage in the process with fresh ideas and perspectives - I really appreciate it. 🙏 🍺
> 
> **Second, apologies for the delay.**
> 
> **Third, now that the requirements are clear, the changes make sense and look good overall.** 🚀 I just have a few comments related to code style. Here's what I suggest:
> 
> * Rebase onto the latest main
> * Go through the comments
> * Add some unit tests
> 
> I'm happy to help with any (or all) of these steps - just let me know!

Restructured code as suggested

redisinsight/ui/src/pages/settings/SettingsPage.tsx
yashschandra: added a new component `<ThemeSettings />`

pawelangelow:
@yashschandra I opened a PR with a few additional tweaks and polish to complement your changes. https://github.com/yashschandra/RedisInsight/pull/1

pawelangelow (approved):
Thanks for your time and cooperation on this PR! 🙏

PR #4457, opened 2025-03-31
RI-6904: show/hide databases list columns

dantovska:
Added a "Columns" button, which will be used to configure which columns to show:
image

When clicked:
image


redisinsight/ui/src/pages/home/components/database-list-component/DatabasesListWrapper.tsx
```diff
@@ -514,6 +516,15 @@ const DatabasesListWrapper = (props: Props) => {
...
   ]
 
+  const [columns, setColumns] = useState<
+    EuiTableFieldDataColumnType<Instance>[]
+  >(initialColumns.filter((c) => shownColumns.includes(c.field)))
+
+  useEffect(() => {
+    const filteredColumns = initialColumns.filter((column) => shownColumns.includes(column.field))
+    setColumns([...filteredColumns])
+  }, [shownColumns])
```
KrumTy, line 526: this adds quite a lot of overhead to the already rather complex flow, let me explain.
We have DatabaseListWrapper that renders the ItemList (virtualized list) and on every react render we have:
DatabaseListWrapper
- initialColumns array being generated 
- initial state being generated for columns state (line 521), set once but generated every time
- whenever shownColumns changes, it updates the columns state
ItemList
- columns state from DatabaseListWrapper is passed as unmemoized prop
- it is then immediately stored in another state
- this second state now updates whenever the columns prop changes via useeffect, sort of making it obsolete
- the second state is also updated on width change via adjustColumns 

if such major modifications to the db list is required, it might be worth extracting this whole logic in a custom hook and optimizing it (no double states or plain objects etc)

redisinsight/ui/src/pages/home/components/database-list-header/DatabaseListHeader.tsx
```diff
@@ -79,6 +90,34 @@ const DatabaseListHeader = ({ onAddInstance }: Props) => {
...
+    } else {
+      hidden.push(column)
+    }
+
+    sendEventTelemetry({
+      event: TelemetryEvent.DATABASE_LIST_COLUMNS_CLICKED,
+      eventData: {
+        shown,
+        hidden
+      }
```
KrumTy, line 117: wouldn't is make sense to send data about all columns here and not just the one that's modified? The shown + hidden arrays suggest otherwise

dantovska: Makes sense, I just followed the already existing telemetry logic on the Keys Header columns for the TTL and Size... @ViktarStarastsenka  what do you think?

KrumTy (approved):

PR #4458, opened 2025-03-31
RI-6998: flex components

pd-redis:
Goal of this PR is to implement alternatives to EuiFlexItem and EuiFlexGroup components and replace them throughout the codebase
If you are reviewing this PR, please checkout the branch and click around, it is not unlikely to have small (or not so small) differences in the margins, paddings etc, mainly because of the overrides applied in custom scss files.


redisinsight/ui/src/components/base/layout/Flex.tsx
```diff
@@ -0,0 +1,205 @@
...
+  4,
+  5,
+  6,
+  7,
+  8,
+  9,
+  10,
+] as const
+export type FlexItemProps = React.HTMLAttributes<HTMLDivElement> & {
+  grow?: boolean | 0 | 5 | 2 | 8 | 3 | 7 | 1 | 4 | 6 | 9 | 10 | null
```
KIvanow, line 163: How/why isn't this in order?

KIvanow:
There are some UI issues with the changes
image
image
For instance there is space missing here between the buttons for adding the database

image
image
Also, the breaking on different size for the buttons and type of databases behaves differently (might be out of scope for this PR though)

pd-redis:
> There are some UI issues with the changes image image For instance there is space missing here between the buttons for adding the database
> 
> image image
> 
> Also, the breaking on different size for the buttons and type of databases behaves differently (might be out of scope for this PR though)

Do we have designs for small screens? In my opinion it looks worse on `main` branch, but I can probably reach the same look with custom component as well

pd-redis:
> I couldn't find any specific designs for the smaller screens. Maybe @Maayanbt or @ViktarStarastsenka could help us here. It doesn't need to be exactly the same as the current implementation, but with these changes we are clipping buttons and elements outside of the screen, which makes it worse than the current implementation

this should be addressed now

dantovska (approved):

redisinsight/ui/src/components/base/layout/flex/flex.tsx
```diff
@@ -0,0 +1,138 @@
...
+      {...rest}
+      align={contentCentered ? 'center' : align}
+      justify={contentCentered ? 'center' : justify}
+      className={classes}
+      direction={reverse ? 'columnReverse' : 'column'}
+    />
+  )
+}
+
+export const Row = ({
```
dantovska, line 92: Is the documentation for Row missed, as it is added for FlexItem and Col?

dantovska (approved):
looks great!

pawelangelow:
I merged the big PR (https://github.com/RedisInsight/RedisInsight/pull/4500), and prepared a merge commit for this PR: https://github.com/RedisInsight/RedisInsight/commit/ab61c12df7442ee51fa035f577cdfb364e12a879 if you want to cherry pick it.

dantovska (approved):

dantovska (approved):

KIvanow (approved):

PR #4460, opened 2025-03-31
RI-6968: Send the number of keys in the Viewed Browser Page event

seth-riedel:


redisinsight/ui/src/pages/browser/BrowserPage.tsx
```diff
@@ -134,10 +135,10 @@ const BrowserPage = () => {
   }, [openedSidePanel])
 
   useEffect(() => {
-    if (connectedInstanceName && !isPageViewSent) {
-      sendPageView(instanceId)
+    if (connectedInstanceName && overview?.totalKeys !== undefined && !isPageViewSent) {
+      sendPageView(instanceId, overview?.totalKeys)
     }
-  }, [connectedInstanceName, isPageViewSent])
+  }, [connectedInstanceName, overview, isPageViewSent])
```
ArtemHoruzhenko, line 141: Not sure if it works this way but are we going to trigger sendPageView overview each time when overview was fetched?

seth-riedel: No, we have the other state `isPageViewSent` that gets set to `true` after we send a page view, so it'll only trigger once

ArtemHoruzhenko (approved):

PR #4462, opened 2025-04-01
RI-6999 focus trap

pd-redis:
replace EuiFocusTrap with custom implementation

### DIFF
--- a/package.json
+++ b/package.json
@@ -269,6 +269,7 @@
     "react": "^18.2.0",
     "react-contenteditable": "^3.3.5",
     "react-dom": "^18.2.0",
+    "react-focus-on": "^3.9.4",
     "react-hotkeys-hook": "^3.3.1",
     "react-jsx-parser": "^1.28.4",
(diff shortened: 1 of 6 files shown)


KIvanow (changes requested):
looks good and it seems to work, but the unit tests need fixing.

pd-redis:
> looks good and it seems to work, but the unit tests need fixing.

From what I see, the errors are caused by other branches, eg.:
image
image

I could still "fix"  them to pass here, but this will add conflicts with main

pd-redis:
> > > looks good and it seems to work, but the unit tests need fixing.
> > 
> > 
> > From what I see, the errors are caused by other branches, eg.: image image
> > I could still "fix" them to pass here, but this will add conflicts with main
> 
> It's something that was missed from somebody during a PR review at some point. I am flagging it the moment I see it. We shouldn't just ignore broken test cases in any PR

It seems it was fixed in main, I rebased on main and tests pass

PR #4464, opened 2025-04-01
fix failing cloudAds unit tests

KrumTy:


redisinsight/ui/src/pages/home/components/database-list-header/DatabaseListHeader.spec.tsx
```diff
@@ -59,12 +62,6 @@ describe('DatabaseListHeader', () => {
   })
 
   it('should show promo cloud button with enabled feature flag', () => {
-    (appFeatureFlagsFeaturesSelector as jest.Mock).mockReturnValueOnce({
-      enhancedCloudUI: {
-        flag: false
-      }
-    })
-
```
dantovska, line 67: Is this not needed at is already set to false, or removing this lines means we are leaving the enhancedCloudUI to true?

redisinsight/ui/src/components/messages/module-not-loaded/ModuleNotLoaded.tsx
```diff
@@ -85,7 +85,7 @@ const ModuleNotLoaded = ({ moduleName, id, type = 'workbench', onClose }: IProps
     if (!cloudAdsFeature?.flag) {
       return (
         <EuiText className={cx(styles.text, styles.marginBottom)}>
-          Open a database with {moduleName?.toLowerCase()}.
+          Open a database with {moduleName}.
```
dantovska, line 88: This makes more sense to me, but are we sure we want to change the actual text and no the test that is failing?

redisinsight/ui/src/pages/home/components/database-list-header/DatabaseListHeader.spec.tsx
KrumTy: it is set to false on top of the file, yes

dantovska (approved):

PR #4465, opened 2025-04-01
RI-7000: add ShowFor, HideFor components

pd-redis:
EUI ShowFor and HideFor components are used to conditionally render their children for specified screen sizes

### DIFF
--- a/redisinsight/ui/src/components/base/utils/ShowHide.spec.tsx
+++ b/redisinsight/ui/src/components/base/utils/ShowHide.spec.tsx
@@ -1,0 +1,153 @@
+import React from 'react'
+import { render, screen } from 'uiSrc/utils/test-utils'
+
+import {
+  Breakpoints,
+  HideFor,
+  ShowFor,
+} from 'uiSrc/components/base/utils/ShowHide'
+
+describe('ShowHide', () => {
+  beforeAll(() => {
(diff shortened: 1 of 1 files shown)


redisinsight/ui/src/components/base/utils/ShowHide.tsx
```diff
@@ -0,0 +1,86 @@
...
+export interface ShowHideForProps {
+  /**
+   * Required otherwise nothing ever gets returned
+   */
+  children: ReactNode
+  /**
+   * List of all the responsive sizes to hide the children for.
+   * Array of #BreakpointKey
+   */
+  sizes: BreakpointKey[] | 'all' | 'none'
```
KIvanow, line 29: What's the point of 'all' or 'none' in this case? Won't one of them just always render (identical to just a div), and the other render never?

pd-redis: It is replicating the API of the EUI component, my guess is that this would be to dynamically change the behavior of the component. For example we could use this instead of FeatureFlagComponent with conditional `sizes={flag === true ? 'all':'none'}` 
Obviously, we don't use these components for any of the cases, so I am ok with leaving just the bare minimum of matching API

KIvanow (approved):

PR #4467, opened 2025-04-01
RI-6945: Introduce Monaco editor for editing ReJSONs

pawelangelow:
# Interesting parts

Apart from the JSON validation, which is handled by Monaco editor - not. You can check the `useMonacoValidation` hook for more details. Another alternative would be `JSON.parse()` but using Monaco is more memory efficient. However it comes with some complexity - it's async, so we need to listen on the event emitters the editor provides.

# Preview

https://github.com/user-attachments/assets/c115658b-21cc-4233-8bb9-9bbeac51ba9b


redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/rejson-details/RejsonDetails.tsx
```diff
@@ -87,13 +89,14 @@ const RejsonDetails = (props: BaseProps) => {
     <div className={styles.jsonData} id="jsonData" data-testid="json-data">
       <>
         {(isObject || isArray) && (
-          <div className={styles.row}>
+          <div className={cx(styles.row, styles.topRow)}>
```
KIvanow, line 92: Is there a benefit for using cx instead of a template literal? Just wandering about it, it's not a required change :)

redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/styles.module.scss
```diff
@@ -255,3 +262,26 @@
...
+      margin-right: 1em;
+    }
+  }
+}
+
+.editor, .editorWrapper {
+  // Using 99% instead of 100% to avoid a weird Monaco animation that kicks in at full height,
+  // which causes the editor to overflow its container.
+  // TODO: Find a cleaner fix for this.
+  height: 99% !important;
```
KIvanow, line 286: weird question, but wouldn't calc(100% - 1px) be cleaner as the 99% might be more off in some screens/resolutions?

redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/RejsonDetailsWrapper.tsx
```diff
@@ -96,9 +97,16 @@ const RejsonDetailsWrapper = (props: Props) => {
     })
   }
 
+  const shouldShowDefaultEditor =
```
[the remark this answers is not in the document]
KIvanow: I can also see the benefit of doing it this way to ensure that if we add a third option, the `else` won't be triggered for it by default.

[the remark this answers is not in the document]
pawelangelow: I was wondering for this, and it's definitely a personal preference, I just don't want to use nested ternaries (the loading and the chosen editor). I can change it if it's creating additional complexity.

redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/styles.module.scss
pawelangelow: You're right, got some extra pixels. Looks like this is a problem related to the borders, but can't find the core issue. So `calc(100% - 2px)` seems to fix the problem.

KrumTy (dismissed):
LGTM, well done 👍

KrumTy (approved):

PR #4473, opened 2025-04-03
RI-7026: Add JSON key -> Create free db dialog

dantovska:
Changing the action of this button:
Screenshot 2025-04-03 at 9 08 12

Currently, it opens a new browser tab - [https://redis.io/try-free/](https://redis.io/try-free/). This will still be the case when cloudSso is false. But when it is true we want the sso dialog to open:
image

Changes applied:
- Replace the "free trial redis cloud database" JSON link with the CloudLink component which still does what the link did when ssoCloud=false (open a browser tab with /try-free), and opens the SSO dialog when ssoCloud=true.
- Add UTM properties for the link
- Center the "Free trial Cloud database" title

Prerequisites:
- The database does not contain JSON module (ex. not a redis stack db, just simple redis 7 server)
- cloudSso is set to true
- cloudAds is set to true - this is true by default

### DIFF
--- a/redisinsight/ui/src/components/markdown/CloudLink/CloudLink.tsx
+++ b/redisinsight/ui/src/components/markdown/CloudLink/CloudLink.tsx
@@ -6,17 +6,18 @@ import { OAuthSsoHandlerDialog } from 'uiSrc/components'
 export interface Props {
   url: string
   text: string
+  source: OAuthSocialSource
 }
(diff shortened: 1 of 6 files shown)


ArtemHoruzhenko (approved):
lgtm

pawelangelow (approved):

PR #4486, opened 2025-04-08
RI-7008: replace eui spacer

pd-redis:


KIvanow (approved):
looks good. Did some spot checking on all of the major screens. Didn't see anything obviously wrong

pd-redis:
> Lots of changes 😄
> 
> What might help for other PRs is to run the linter before actual changes. I imagine process like:
> 
> 1. Discover the files that needs to be changed
> 2. Format and commit them
> 3. Change what you need to change and commit
> 
> this way, the significant changes can be isolated and reviewed more easily (yes, on the "Files changed" tab everything will be shown, but if you review commit by commit this will definitely help).
> 
> If it helps, you can use this dummy script (or you can alter it the way you like):
> 
> ```
> #!/usr/bin/env bash
> 
> set -e
> 
> echo "🔍 Enter search string (e.g., <EuiSpacer):"
> read search
> 
> echo ""
> echo "📂 Searching files for: $search"
> 
> # Find matching files (excluding node_modules, dist, etc.)
> files=$(grep -rl --include=\*.{js,jsx,ts,tsx} "$search" . \
>   --exclude-dir={node_modules,dist,build,out,coverage} || true)
> 
> if [[ -z "$files" ]]; then
>   echo "❌ No matching files found."
>   exit 0
> fi
> 
> # Count files
> file_count=$(echo "$files" | grep -c '^' || echo 0)
> 
> # Convert newlines to space-separated list
> file_list=$(echo "$files" | tr '\n' ' ')
> 
> echo ""
> echo "📄 Found $(echo "$files" | wc -l) files:"
> echo "$files"
> 
> echo ""
> echo "🧼 Running Prettier..."
> ./node_modules/prettier/bin/prettier.cjs --write $file_list || echo "⚠️ Prettier failed."
> 
> echo ""
> echo "✅ Done."
> ```

I appreciate the suggestion, but unfotunately identifying the files is not always easy :)

PR #4488, opened 2025-04-08
RI-7009: eui window event

pd-redis:
replace EuiWindowEvent component


redisinsight/ui/src/components/base/utilities/WindowEvent.spec.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react'
+import { render } from 'uiSrc/utils/test-utils'
+import { WindowEvent } from './WindowEvent'
+
+describe('WindowEvent', () => {
```
pawelangelow, line 5: Not sure how meaningful it would be, as we use the `window`, but do you think a test with setup for `click` and assert that the handler is called would be beneficial?

pawelangelow (approved):
🚀

redisinsight/ui/src/components/base/utilities/WindowEvent.spec.tsx
pd-redis: Added test for that

pawelangelow (approved):

PR #4498, opened 2025-04-15
Fix Add Key panel cancel regression

dantovska:
While fixing another bug in this PR - https://github.com/RedisInsight/RedisInsight/pull/4339, I introduced a regression.
When user has opened the add key form but clicks on the cancel button (right next to the Add button), an toast message appears:
image

It attempts to add the empty key object, or even if there is some input for the name to the state, not considering that it may be canceled. 

Applying this changes so  If add key panel is closed by clicking cancel, do not pass the key name to the add key panel handler above, otherwise, include it - as is.


KIvanow:
The comments with the isCancelled right under seem a bit redundant. Other than that it looks good

KIvanow (approved):

PR #4503, opened 2025-04-16
RI-7002: list components

pd-redis:
Re-implement list group and list item components with styled components

* all `$` prefixed props are for `styled-components` indication not to pass them to unerlying dom, so that we don't get `<div color="text" .../>` dom elements
* copied useInnerText hook
* left `list.module.scss` for the moment, until base theme values are included in `main` branch
* placed bunch of css properties in styled-components until the above happens
* replaced all occurrences of EuiListGroup and EuiListGroupItem


redisinsight/ui/src/components/base/layout/list/list.styles.ts
```diff
@@ -0,0 +1,398 @@
...
+export const SIZES = ['xs', 's', 'm', 'l'] as const
+export type ListGroupItemSize = (typeof SIZES)[number]
+
+export const COLORS = ['primary', 'text', 'subdued', 'ghost'] as const
+export type ListGroupItemColor = (typeof COLORS)[number]
+
+export type ListGroupItemProps = HTMLAttributes<HTMLLIElement> & {
+  /**
+   * Size of the label text
+   */
```
KrumTy, line 98: I think some of these belong in a separate file (types.ts)

redisinsight/ui/src/components/base/layout/list/Group.tsx
```diff
@@ -0,0 +1,38 @@
...
+  className,
+  style,
+  maxWidth = true,
+  gap,
+  flush,
+  ...rest
+}: ListGroupProps) => {
+  let newStyle = style
+
+  if (maxWidth && maxWidth !== true) {
```
KrumTy, line 20: why maxWidth can be true?

redisinsight/ui/src/components/base/layout/list/Item.tsx
```diff
@@ -0,0 +1,202 @@
...
+}: ListGroupItemProps) => {
+  const isClickable = !!onClick
+  let iconNode: ReactElement
+
+  if (iconType) {
+    // todo replace with redis-ui icon
+    iconNode = (
+      <EuiIcon
+        color="inherit" // forces the icon to inherit its parent color
+        {...iconProps}
```
KrumTy, line 44: this can be a separate component that you can pass props to and render bellow (as opposed to declaring it above hooks); same for itemContent

redisinsight/ui/src/components/base/layout/list/Item.tsx
```diff
@@ -0,0 +1,202 @@
...
+ *     <div ref={ref}>
+ *       {innerText}
+ *     </div>
+ *   )
+ * }
+ *
+ * @param innerTextFallback Value to return if `ref` is null or does not have an `innerText` property.
+ * @returns A tuple containing a function to update the `ref` and the current `innerText` value.
+ */
+export function useInnerText(
```
KrumTy, line 160: this exported but I don't see references in other files; might be worth considering moving to a separate file (hooks.ts)

redisinsight/ui/src/components/base/layout/list/list.styles.ts
```diff
@@ -0,0 +1,398 @@
...
+  listItem: 'RI-list-group-item',
+  listItemLabel: 'RI-list-group-item-label',
+  listItemButton: 'RI-list-group-item-button',
+  listItemText: 'RI-list-group-item-text',
+  listGroup: 'RI-list-group',
+  listItemActive: 'isActive',
+  listItemDisabled: 'isDisabled',
+}
+
+export const GAP_SIZES = ['none', 's', 'm'] as const
```
KrumTy, line 25: exported constants makes sense to be in a separate file (constants.ts), no?

redisinsight/ui/src/components/side-panels/panels/enablement-area/EnablementArea/components/InternalLink/InternalLink.tsx
```diff
@@ -48,28 +49,28 @@ const InternalLink = (props: Props) => {
...
-      )}
-      iconType={iconType}
-      size={size}
-      wrapText
-      color="subdued"
-      onClick={handleOpenPage}
-      label={content}
-      {...rest}
-    />
+    <>
```
KrumTy, line 58: I think you don't need this fragment

redisinsight/ui/src/components/side-panels/panels/enablement-area/EnablementArea/components/Navigation/Navigation.tsx
```diff
@@ -40,28 +53,41 @@ export interface Props {
...
 
-  useEffect(() => () => {
-    dispatch(setWbCustomTutorialsState())
-  }, [])
+  useEffect(
+    () => () => {
+      dispatch(setWbCustomTutorialsState())
+    },
+    [],
+  )
```
KrumTy, line 90: I like the above version more; maybe something is wrong with our formatting
what if you add {} in the body of first callback?

KrumTy (commented):
overall looks good, just some refactoring suggestions

pawelangelow:
I merged the big PR (https://github.com/RedisInsight/RedisInsight/pull/4500), and prepared a merge commit for this PR: https://github.com/RedisInsight/RedisInsight/commit/9f07d682db2c4f48db1cacecee7194c6c1f88505 if you want to cherry pick it.

redisinsight/ui/src/components/base/layout/list/list.styles.ts
pd-redis: Yes, but because they are quite common for many of the components, adding them in separate files now will lead either to conflicts or to many similar type files. 
So my thought is, that after these components are merged, to go around them all and move the common bits in one place

redisinsight/ui/src/components/base/layout/list/Item.tsx
pd-redis: Same as for types, after merging

redisinsight/ui/src/components/base/layout/list/Item.tsx
pd-redis: Until we start using redis ui icons, this should remain, after that, i think only the `icon` prop should be left and all this removed

redisinsight/ui/src/components/base/layout/list/list.styles.ts
pd-redis: Same as above

redisinsight/ui/src/components/side-panels/panels/enablement-area/EnablementArea/components/Navigation/Navigation.tsx
pd-redis: this is all result of the formatting rules, all comments should be made towards @pawelangelow :D

redisinsight/ui/src/components/base/layout/list/Group.tsx
pd-redis: updated

redisinsight/ui/src/components/base/layout/list/Item.tsx
KIvanow: btw, you can keep track of progress for the icons here -> https://www.figma.com/design/AAG5lMHFklYJcZy7kZGYTg/DS-Alignment?node-id=5-1390&p=f&m=dev

KIvanow (dismissed):
lgtm overall

redisinsight/ui/src/components/instance-header/components/instances-navigation-popover/components/instances-list/InstancesList.tsx
```diff
@@ -20,6 +15,10 @@ import {
   sendEventTelemetry,
 } from 'uiSrc/telemetry'
 import { getDbIndex } from 'uiSrc/utils'
+import {
+  Group as ListGroup,
+  Item as ListGroupItem,
+} from 'uiSrc/components/base/layout/list'
```
dantovska, line 21: I think it would be better if there is an index.ts file in layout and all layout components are imported from layout and not from the internal directories.

KIvanow (approved):

PR #4506, opened 2025-04-17
RI-7005: page and page body

pd-redis:


redisinsight/ui/src/components/base/layout/page/Page.tsx
```diff
@@ -0,0 +1,52 @@
...
+    $direction={direction}
+    $restrictWidth={restrictWidth}
+    $paddingSize={paddingSize}
+    style={restrictWidthSize(style, restrictWidth)}
+  />
+)
+
+export default Page
+
+export const PageContentBody = ({
```
KrumTy, line 34: why not in a separate file?

KrumTy (dismissed):
lgtm

KIvanow (approved):

redisinsight/ui/src/components/base/layout/page/Page.tsx
pd-redis: I don't think that separating components that don't do too much makes sense, but will move it for consistency with the other ones.

KrumTy (approved):

package.json
```diff
@@ -32,6 +32,7 @@
     "lint:e2e": "yarn --cwd tests/e2e lint",
     "prettier": "prettier --check .",
     "prettier:update": "prettier --write .",
+    "prettier:fix": "prettier --write",
```
pawelangelow, line 35: Is this a leftover? I don't mind using `prettier:fix` but let's keep just one of them - either `prettier:update` or `prettier:fix`

pd-redis: `"prettier:update": "prettier --write .", ` hardcodes which files to run on, which is the current folder.
`"prettier:fix": "prettier --write", ` can take list of file to format specifically

KrumTy (approved):

PR #4514, opened 2025-04-23
RI-7076: Use legacy ReJSON path

pawelangelow:
You can get more context on ReJSON path from the [official documentation.](https://redis.io/docs/latest/develop/data-types/json/path/#legacy-path-syntax).

But if you want just the TL;DR:

- ReJSON v2 uses `$.this.syntax`
- ReJSON v1 uses `.this.syntax` and this one is also supported in ReJSON v2, but it's considered legacy.

So with this PR:
- if ReJSON version is not reported - we fallback to legacy path option (a.k.a. `.`)
- if ReJSON version is 1 - again, the `.` path
- if ReJSON version is 2 - `$`


ArtemHoruzhenko (approved):
lgtm

PR #4518, opened 2025-04-24
RI-7081: database connection failed is sent multiple times

pd-redis:
When DB connection is not working, we're sending telemetry messages. Since `/overview` endpoint is called repeatedly every 5 s by default, we're sending a lot of redundand telemetry messages 

With this PR, overview call is made only if there is no connectivity error registered.

Also small connection lost indicator is added.
image

### DIFF
--- a/redisinsight/ui/src/components/database-overview/DatabaseOverview.tsx
+++ b/redisinsight/ui/src/components/database-overview/DatabaseOverview.tsx
@@ -22,11 +22,8 @@ import {
 import { ThemeContext } from 'uiSrc/contexts/themeContext'
 import { sendEventTelemetry, TelemetryEvent } from 'uiSrc/telemetry'
 import { toBytes, truncatePercentage } from 'uiSrc/utils'
-import {
-  appConnectivityError,
-  setConnectivityError,
-} from 'uiSrc/slices/app/connectivity'
-import ApiErrors from 'uiSrc/constants/apiErrors'
+import { appConnectivityError } from 'uiSrc/slices/app/connectivity'
+import WarningIcon from 'uiSrc/assets/img/warning.svg?react'
 import { getOverviewMetrics, IMetric } from './components/OverviewMetrics'
 
 import AutoRefresh from '../auto-refresh'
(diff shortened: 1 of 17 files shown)


KIvanow (commented):
Code changes look good! But the FE tests are failing

redisinsight/ui/src/pages/rdi/instance/InstancePage.spec.tsx
```diff
@@ -183,15 +184,15 @@ describe('InstancePage', () => {
       .fn()
       .mockReturnValue({ pathname: Pages.rdiPipeline(RDI_INSTANCE_ID_MOCK) })
 
-    await act(() => {
+    act(() => {
```
pawelangelow, line 187: Should these be awaited as well?

redisinsight/ui/src/components/database-overview/DatabaseOverview.tsx
```diff
@@ -130,105 +140,138 @@ const DatabaseOverview = () => {
...
-                </EuiButton>
-              </EuiFlexItem>
-            )}
-            {metrics?.map((overviewItem) => (
+                  <EuiFlexItem grow={false}>
+                    <EuiIcon size="m" type={WarningIcon} />
+                  </EuiFlexItem>
+                </EuiFlexGroup>
+              </EuiToolTip>
+            </EuiFlexItem>
```
pawelangelow, line 173: I'd introduce:
```
const Error = ({ // or ConnectivityError, or something else
  connectivityError,
}: {
  connectivityError: string | undefined
}) => (
  <EuiFlexItem
    key="connectivityError"
    className={styles.overviewItem}
    data-test-subj="connectivityError"
    grow={false}
  >
    <EuiToolTip
      position="bottom"
      className={styles.tooltip}
      content={connectivityError}
    >
      <EuiFlexGroup
        gutterSize="none"
        responsive={false}
        alignItems="center"
        justifyContent="center"
      >
        <EuiFlexItem grow={false}>
          <EuiIcon size="m" type={WarningIcon} />
        </EuiFlexItem>
      </EuiFlexGroup>
    </EuiToolTip>
  </EuiFlexItem>
)
```

It's not perfect, but IMO we should try to push in this direction (to split the big components into some smaller ones)

redisinsight/ui/src/components/database-overview/DatabaseOverview.tsx
```diff
@@ -87,13 +93,17 @@ const DatabaseOverview = () => {
...
+    if (connectedInstanceId && !connectivityError) {
       dispatch(getDatabaseConfigInfoAction(connectedInstanceId))
       setLastRefreshTime(Date.now())
     }
   }
+  useEffect(() => {
+    if (!connectivityError) {
+      loadData()
+    }
+  }, [connectivityError])
```
KrumTy, line 106: I don't like this approach as it can lead to issues and is hard to follow:
loadData - an arrow function that's generated on each render
the useEffect - triggers only when connectivityError changes so it can use stale loadData function (e.g. connectedInstanceId has now changed)

is it intentional that the dispatch shouldn't trigger if the connectedInstanceId changes? If not, a possible refactor could be:

```suggestion
useEffect(() => {
  if (!connectivityError && connectedInstanceId) {
    dispatch(getDatabaseConfigInfoAction(connectedInstanceId))
    setLastRefreshTime(Date.now())
  }
}, [connectivityError, connectedInstanceId])
```

pd-redis: `loadData` is used in AutoRefresh component. 
I don't see how it can be stale, as you are pointing out, it is recreated on each render.
The purpose of the effect is to re-trigger autorefresh once connectivity is no longer broken, because when there is no connection and you open the app, you get an error and no metrics. If you get no metrics, auto refresh is not set, and you never get metrics, unles you hard reload the app.

KrumTy: it is recreated but the new version will only be used if connectivityError changes; should any of it's other dependancies change (e.g. connectedInstanceId) it won't trigger the useEffect

I suppose in this case this works fine, it's just that this approach is prone to errors as the code grows. See a cool article by Dan Abramov on the topic ([link](https://overreacted.io/a-complete-guide-to-useeffect/#moving-functions-inside-effects))
There's a useful eslint rule that we have turned off for some reason - react-hooks/exhaustive-deps, it forbids such things

redisinsight/ui/src/components/database-overview/hooks/useDatabaseOverview.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+import { useDatabaseOverview } from './useDatabaseOverview'
+
+// Mock the telemetry function
+jest.mock('uiSrc/telemetry', () => ({
+  sendEventTelemetry: jest.fn(),
+  TelemetryEvent: {
+    OVERVIEW_AUTO_REFRESH_ENABLED: 'overview_auto_refresh_enabled',
+    OVERVIEW_AUTO_REFRESH_DISABLED: 'overview_auto_refresh_disabled',
+  },
+}))
```
pawelangelow, line 20: It's OK as is, but I'm wondering if it makes more sense to use the actual `TelemetryEvents`. In this case, if someone changes them, the test would potentially be more meaningful.
```suggestion
jest.mock('uiSrc/telemetry', () => {
  const actual = jest.requireActual('uiSrc/telemetry')
  return {
    ...actual,
    sendEventTelemetry: jest.fn(),
  }
})
```

redisinsight/ui/src/components/database-overview/hooks/useDatabaseOverview.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+  db: mockDb,
+})
+
+let mockedStore: ReturnType<typeof mockStore>
+// Set up fake timers
+jest.useFakeTimers()
+let mockDate: Date
+type HookReturnType = ReturnType<typeof useDatabaseOverview>
+
+const helper = (store: typeof mockedStore) => {
```
pawelangelow, line 68:
```suggestion
const render = (store: typeof mockedStore) => {
```

redisinsight/ui/src/components/database-overview/hooks/useDatabaseOverview.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+}
+
+describe('useDatabaseOverview', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockedStore = mockStore(initialState)
+
+    mockDate = new Date('2024-11-22T12:00:00Z')
+    jest.setSystemTime(mockDate)
+  })
```
pawelangelow, line 83:
```suggestion
  beforeEach(() => {
    jest.clearAllMocks()
    mockedStore = mockStore(initialState)

    mockDate = new Date('2024-11-22T12:00:00Z')
    jest.setSystemTime(mockDate)
  })
  
  afterEach(() => {
    jest.useRealTimers()
  })
```

redisinsight/ui/src/components/database-overview/hooks/useDatabaseOverview.spec.ts
pd-redis: I specifically did not use render, to not make it confusing with the regular RTL render function

pawelangelow: let's make it `renderHook` or `renderHelper` or something like that?

pd-redis: I changed it, but i find these types of discussion pointless. The name of a local test variable should not be the subject of multy day discussion

pawelangelow (approved):

KrumTy (approved):

KIvanow (approved):

redisinsight/ui/src/components/database-overview/hooks/useDatabaseOverview.spec.ts
pawelangelow: Well, I'm 50:50 here.

- yes, it's not the best use of our time to discuss the name of a variable
- and no. These suites tend to get bigger and bigger, and usually are copied and pasted. Having more explicit and verbose syntax, IMO, helps a lot. So, `helper` != `setup` or `renderHelper` or even `renderUseDatabaseOverview`

pawelangelow: Well, it's not marked with `Request changes`, but yes, we definitely can improve the PR review process in terms of expectations. "Good first issue" for the retrospective meetings :D

PR #4520, opened 2025-04-24
RI-7004: outside click detector

pd-redis:
- implement outside click detector
- EuiOutsideClickDetector replaced in project


KIvanow (approved):

KIvanow:
Code changes look good, but there is 1 failing FE test though

pd-redis:
> Code changes look good, but there is 1 failing FE test though

image
 it only fails in GH. And often tests even fail to start

pd-redis:
> Code changes look good, but there is 1 failing FE test though

And the test that fails in GH and the component it tests, do not use/reference OutsidClickDetector in any way

pd-redis:
> Code changes look good, but there is 1 failing FE test though

And now test passed in GH without any change to the branch!

KIvanow:
> And the test that fails in GH and the component it tests, do not use/reference OutsidClickDetector in any way

> it only fails in GH. And often tests even fail to start

I agree that our FE tests are unreliable. We'll need to come up with a way of updating them over time. However, this doesn't mean we should ignore randomly failing tests. At least we could flag them as ones that need reworking.

KIvanow:
> Code changes look good, but there is 1 failing FE test though
> 
> And now test passed in GH without any change to the branch!

I saw. Once again - we need to refactor some (if not all) of the FE tests at some point as they are far from reliable. This doesn't mean we should just ignore them though.

It's approved so you can merge it, but please flag such flaky tests, so we can group them up and plan  for their fixes

PR #4522, opened 2025-04-24
RI-6945: Reset editor to default every time a ReJSON is open

pawelangelow:
No tests, as we don't have a cool way to test it 😿 If you have something in mind - feel free to share it, if not, it's a good candidate for E2E.

Here's a preview:

https://github.com/user-attachments/assets/478c08b1-00e0-4e56-a1cb-b125fca1e0c4

And how it looked before:

https://github.com/user-attachments/assets/4b8aad8e-b105-4f80-a3e9-f838f9aba35f

### DIFF
--- a/redisinsight/ui/src/slices/tests/browser/rejson.spec.ts
+++ b/redisinsight/ui/src/slices/tests/browser/rejson.spec.ts
@@ -7,6 +7,7 @@ import {
   mockedStore,
   mockStore,
 } from 'uiSrc/utils/test-utils'
+import { EditorType } from 'uiSrc/slices/interfaces'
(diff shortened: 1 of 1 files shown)


KIvanow (approved):

dantovska (approved):

KIvanow:
> > @pawelangelow the FE tests have starting failing though.
> > I agree it would be a great candidate for e2e later on :)
> 
> Yes, it turns out there are store tests. IMO, this is not the way to test such functionality, as only the store gets tested, not the integration of the visual part with the store, but still, they needed to be fixed.

I agree. Overall our FE tests need rethinking. They produce too many false positives/negatives

KIvanow (approved):

PR #4527, opened 2025-04-28
RI-6950: Fetch whole json when Monaco editor is selected

pawelangelow:
3 major things are happening in this PR.

1. `forceRetrieve` flag used for the `ApiEndpoints.REJSON_GET` call is `true` when the selected editor is Monaco.
2. When the selected editor changes, the `ApiEndpoints.REJSON_GET` gets called (because, based on the editor, the result is going to be different).
3. When the selected Redis key is changed, the selected editor resets to the default value.

With these constraints, the logic inside `RejsonDetailsWrapper` gets a bit complicated, so we should rework the data flow to make the component dummy.

### DIFF
--- a/redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/RejsonDetailsWrapper.spec.tsx
+++ b/redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/RejsonDetailsWrapper.spec.tsx
@@ -1,14 +1,89 @@
 import React from 'react'
+import { useDispatch, useSelector } from 'react-redux'
 import { instance, mock } from 'ts-mockito'
 import { render } from 'uiSrc/utils/test-utils'
+import { fetchReJSON } from 'uiSrc/slices/browser/rejson'
+import { EditorType } from 'uiSrc/slices/interfaces'
+import { stringToBuffer } from 'uiSrc/utils'
+
 import { RejsonDetailsWrapper, Props } from './RejsonDetailsWrapper'
+
+jest.mock('react-redux', () => ({
+  useDispatch: jest.fn(),
+  useSelector: jest.fn(),
+}))
+
+jest.mock('uiSrc/slices/browser/rejson', () => ({
+  ...jest.requireActual('uiSrc/slices/browser/rejson'),
+  fetchReJSON: jest.fn((key) => ({ type: 'FETCH_REJSON', payload: key })),
+}))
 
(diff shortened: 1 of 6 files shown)


redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/RejsonDetailsWrapper.tsx
```diff
@@ -43,11 +49,25 @@ const RejsonDetailsWrapper = (props: Props) => {
...
+  // TODO: the whole workflow should be refactored
+  // in a way that this component will not be responsible for fetching data
+  // based on the editor type
+  useEffect(() => {
+    if (!selectedKey) return
+
+    if (prevEditorTypeRef.current !== editorType) {
+      prevEditorTypeRef.current = editorType
+
+      dispatch(fetchReJSON(selectedKey))
```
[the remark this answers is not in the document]
pawelangelow: Actually, it's a leftover. Currently in the main branch, the editorType reset happens inside `fetchReJSON`, but in this PR, I decided to change it, and I forgot to remove the ref fix. Thanks for noticing it!

redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/RejsonDetailsWrapper.spec.tsx
pd-redis, line 1: I am not sure that testing the state is good approach to testing. My opinion is that for UI tests, we should test what user can see/use, not what actions are trigerred or not.

This way the test is coupled with the state implementation used it tests wether some actions are called and not what the result is. What do you care if `fetchReJSON` is called or not, you should care if you see what you're supposed to see in different occassions

KrumTy (approved):

dantovska (approved):

redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/RejsonDetailsWrapper.spec.tsx
pawelangelow: I totally agree with this. The problem is that we are currently very coupled with the store. IMO, we should move the business logic to hooks, and then it will be easier to test such scenarios - if data changes, what happens, if the editor is changed, what happens, etc. So, at the current position, this is what I think is the best. If you have suggestions, just say the word 😸

KrumTy (approved):

PR #4539, opened 2025-05-05
RI-7106: additiona data in config databases open database

pd-redis:
Add data to: `CONFIG_DATABASES_OPEN_DATABASE`

Data fields added: 
  - redis_version
  - uptime_in_days
  - used_memory
  - connected_clients
  - maxmemory_policy
  - instantaneous_ops_per_sec
  - instantaneous_input_kbps:
  - instantaneous_output_kbps
  - totalKeys
  - numberOfKeysRange

### DIFF
--- a/redisinsight/api/src/modules/database/dto/redis-info.dto.ts
+++ b/redisinsight/api/src/modules/database/dto/redis-info.dto.ts
@@ -1,5 +1,39 @@
 import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 
+class RedisDatabaseStatsDto {
+  @ApiProperty({
+    type: String,
+  })
+  instantaneous_input_kbps: string | undefined;
+
+  @ApiProperty({
+    type: String,
+  })
+  instantaneous_ops_per_sec: string | undefined;
+
+  @ApiProperty({
+    type: String,
+  })
+  instantaneous_output_kbps: string | undefined;
+
+  @ApiProperty({
+    type: Number,
+  })
+  maxmemory_policy: string | undefined;
+
+  @ApiProperty({
+    description: 'Redis database mode',
+    type: String,
+  })
+  numberOfKeysRange: string | undefined;
+
+  @ApiProperty({
+    description: 'Redis database role',
+    type: String,
+  })
+  uptime_in_days: string | undefined;
+}
+
 export class RedisNodeInfoResponse {
   @ApiProperty({
     description: 'Redis database version',
@@ -24,9 +58,9 @@ export class RedisNodeInfoResponse {
 
   @ApiPropertyOptional({
     description: 'Various Redis stats',
-    type: Object,
+    type: RedisDatabaseStatsDto,
   })
-  stats?: any;
+  stats?: RedisDatabaseStatsDto;
 
   @ApiPropertyOptional({
     description: 'The number of Redis databases',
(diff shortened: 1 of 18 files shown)


redisinsight/api/src/modules/database/providers/database-info.provider.ts
```diff
@@ -177,6 +180,20 @@ export class DatabaseInfoProvider {
           parseInt(get(memoryInfo, 'number_of_cached_scripts'), 10) ||
           undefined,
         server: serverInfo,
+        stats: {
+          instantaneous_ops_per_sec:
+            get(statsInfo, 'instantaneous_ops_per_sec') || undefined,
```
ArtemHoruzhenko, line 185: there is no need for `|| undefined`
as I remember lodash `get` function has 3d argument which is `default` value which is  already `undefined` here `get(statsInfo, 'instantaneous_ops_per_sec')`

redisinsight/api/src/modules/database/dto/redis-info.dto.ts
```diff
@@ -22,6 +22,12 @@ export class RedisNodeInfoResponse {
   })
   server?: any;
 
+  @ApiPropertyOptional({
+    description: 'Various Redis stats',
+    type: Object,
+  })
+  stats?: any;
```
ArtemHoruzhenko, line 29: why `any`? does it make sense to have proper class for it and mapping will be done by class-transformer?

ArtemHoruzhenko (approved):
I can see changes on server side but there is no aprropriate tests modified/added
Approve to not block but please address this after merging please

redisinsight/ui/src/components/base/layout/list/Item.tsx
```diff
@@ -118,7 +118,6 @@ const Item = ({
       $isActive={isActive}
       $isDisabled={isDisabled}
       $color={color}
-      onClick={onClick}
```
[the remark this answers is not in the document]
pawelangelow: From what I saw, there are 2 inner components based on the props:
- if disabled = true and there is an onClick handler, it renders a button
- else - renders a span

Probably this is the reason that there were 2 invocations on the handler. The tricky part is the `if (isDisabled || onClick) {` condition, but if it's disabled, we probably don't want to invoke the handler, so we're safe. I'd guard it with a test, tho.

pawelangelow (dismissed):

redisinsight/ui/src/components/base/layout/list/Item.tsx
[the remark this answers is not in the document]
pd-redis: If there is on click handler provided, the button that handles it is rendered, if no onClick is provided, nothing needs to handle it

redisinsight/api/src/modules/database/dto/redis-info.dto.ts
pd-redis: Because server is using the same!?

pd-redis: `stats` is not used for anything else besides sending it for 1 telemetry event, and `server` is, so if we're to type it explicitly, we should type `server` explicitly as well

KIvanow: we have 400+ `any`s in our code base. It might be worthwile to type them all at some point

KIvanow (approved):

PR #4542, opened 2025-05-08
RI-7094: Recalculate provider on db connect

dantovska:
Currently, the provider is recalculated when a database is created or edited. However, instances that were added before this logic was introduced still have outdated or missing provider values.

This PR enhances the “connect database” logic to handle those legacy instances. When a user connects to such a database, the system will:
- Check if the stored provider is part of the current Host Provider list.
- If not, recalculate the provider to align it with the recognized list.
- Update the database instance with the new provider value.

This recalculation happens only once per instance, the first time it is connected after this logic is deployed.

Notes
- This ensures backward compatibility and smooth transition for older instances.
- Helps maintain consistency in provider metadata across all database entries.

### DIFF
--- a/redisinsight/api/src/modules/database/database-connection.service.spec.ts
+++ b/redisinsight/api/src/modules/database/database-connection.service.spec.ts
@@ -115,7 +115,7 @@ describe('DatabaseConnectionService', () => {
 
     it('should recalculate provider if not available in the list of providers', async () => {
       const testHost = 'localhost';
-      repository.get = jest.fn().mockResolvedValue({
+      repository.get.mockResolvedValue({
         host: testHost,
         provider: 'not-in-providers-enum',
       });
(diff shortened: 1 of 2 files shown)


redisinsight/api/src/modules/database/database-connection.service.spec.ts
```diff
@@ -107,6 +113,33 @@ describe('DatabaseConnectionService', () => {
       );
     });
 
+    it('should recalculate provider if not available in the list of providers', async () => {
+      const testHost = 'localhost';
+      repository.get = jest.fn().mockResolvedValue({
```
ArtemHoruzhenko, line 118: You don't need this `= jest.fn()`, repository already mocked.

ArtemHoruzhenko (approved):
Lgtm

KIvanow (approved):

redisinsight/api/src/modules/database/database-connection.service.spec.ts
pd-redis, line 1: Just to clarify, the only meaningful changes in this PR are for this test file?

pd-redis (approved):

PR #4548, opened 2025-05-14
RI-7039: replace eui

pd-redis:
this is the base branch where all ui updates should be merged initially before pushing back to main


redisinsight/ui/src/components/base/forms/buttons/index.tsx
KIvanow, line 1: Did you start seeing a point in the barrel files @pd-redis ?

pd-redis: Just to a reasnoable level. To group all buttons in one I find it acceptable, but to reexport again and again on each level, that is too much

package.json
```diff
@@ -234,6 +234,10 @@
   "dependencies": {
     "@elastic/datemath": "^5.0.3",
     "@elastic/eui": "34.6.0",
+    "@redis-ui/components": "^38.1.4",
```
valkirilov, line 240: `Badge` component in `@redis-ui/components` v38.6.0 supports custom color out of the box ([docs](https://redislabsdev.github.io/redis-ui/?path=/story/components-badge--with-custom-color)), something we need for the Vector Search feature.

Unfortunately, it's dependent on `@redis-ui/styles` which comes with some breaking changes in its latest version, so the upgrade process might need additional effort.

valkirilov, line 240: `Section` component in `@redis-ui/components` v38.7.1 comes with a bug fix that we need for the **Vector Search** page -> **Manage Indexes** drawer -> Index info boxes

When the box section is collapsed, it should display summary information about the index. However, there was a bug in Redis UI that makes this information not appear once the component is rendered inside a Drawer. They resolved the issue on their end, and we have to update Redis UI dependencies.

pd-redis: Update them in a branch and when everything is running, make a PR

PR #4556, opened 2025-05-16
RI-7091 -  add an environment variable to skip the eula screen

KIvanow:


configs/webpack.config.main.prod.ts
```diff
@@ -68,6 +68,7 @@ export default merge(baseConfig, {
       RI_APP_HOST: '127.0.0.1',
       RI_BUILD_TYPE: 'ELECTRON',
       RI_APP_VERSION: version,
+      RI_ACCEPT_TERMS_AND_CONDITIONS: process.env.RI_ACCEPT_TERMS_AND_CONDITIONS || 'false',
```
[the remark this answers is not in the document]
KIvanow: It actually isn't. Just tested it without it being set up and it works without issues. I was following the pattern of adding the env variables

redisinsight/api/src/modules/settings/settings.service.ts
```diff
@@ -56,16 +56,40 @@ export class SettingsService {
   ): Promise<GetAppSettingsResponse> {
     this.logger.debug('Getting application settings.', sessionMetadata);
     try {
-      const agreements =
-        await this.agreementRepository.getOrCreate(sessionMetadata);
       const settings =
         await this.settingsRepository.getOrCreate(sessionMetadata);
+
+      let agreements;
```
KrumTy, line 62: lacks type

redisinsight/api/src/modules/settings/repositories/local.agreements.repository.ts
```diff
@@ -14,12 +14,22 @@ export class LocalAgreementsRepository extends AgreementsRepository {
...
+        if (defaultOptions?.data) {
+          entity = await this.repository.save(
+            this.repository.create({
+              id: 1,
+              data: JSON.stringify(defaultOptions.data),
+            })
+          );
+        } else {
+          entity = await this.repository.save(this.repository.create({ id: 1 }));
+        }
```
ArtemHoruzhenko, line 32: We should use class transformer here to not do this `JSON.stringify(defaultOptions.data),` manually

redisinsight/api/src/modules/settings/settings.service.ts
```diff
@@ -56,16 +56,40 @@ export class SettingsService {
...
+      if (SERVER_CONFIG.acceptTermsAndConditions) {
+        const isEncryptionAvailable =
+          (await this.keyEncryptionStrategy.isAvailable()) ||
+          (await this.keytarEncryptionStrategy.isAvailable());
```
ArtemHoruzhenko, line 66: let's use this method https://github.com/RedisInsight/RedisInsight/blob/main/redisinsight/api/src/modules/encryption/encryption.service.ts#L28 and check it's output or even better to add additional function to that service e.g. `isEncryptionAvailable()` which will return boolean flag based on own method. so all logic will be in a single place

ArtemHoruzhenko:
@KIvanow could you please look on this. https://github.com/RedisInsight/RedisInsight/pull/4567/files
here how I see it but may be there another way to do this

redisinsight/api/src/modules/settings/repositories/local.agreements.repository.ts
KIvanow: Updated with the draft PR

redisinsight/ui/src/components/consents-settings/ConsentOption/ConsentOption.tsx
```diff
@@ -34,7 +55,7 @@ const ConsentOption = (props: Props) => {
             color="subdued"
             style={{ marginTop: '12px' }}
           >
-            {parse(consent.description)}
+            {getText()}
```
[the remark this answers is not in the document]
KIvanow: It would be a more React way. However I am not sure creating a component that will be used only in this case and prop drill it with the data from the parent one is optimal either. 
Is there another benefit of not having a function call, other than sticking to jsx?

redisinsight/api/src/modules/settings/settings.service.ts
ArtemHoruzhenko: Looks good

redisinsight/api/src/modules/settings/repositories/local.agreements.repository.ts
```diff
@@ -14,15 +15,22 @@ export class LocalAgreementsRepository extends AgreementsRepository {
...
+        entity = await this.repository.save(
+          classToClass(AgreementsEntity, plainToInstance(Agreements, {
+            ...defaultOptions,
+            id: 1,
+          })),
+        );
       } catch (e) {
         if (e.code === 'SQLITE_CONSTRAINT') {
-          return this.getOrCreate();
+          return this.getOrCreate(sessionMetadata);
```
pawelangelow, line 33: If this is a retry, shouldn't it pass the options as well?

```suggestion
          return this.getOrCreate(sessionMetadata, defaultOptions);
```

redisinsight/ui/src/components/consents-settings/ConsentOption/ConsentOption.tsx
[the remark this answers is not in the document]
KIvanow: > the produced code looks better and simpler, and maybe the most important part - generally sets the tone to not use it, especially on places where some things can be reused or simplified

I am not sure how the code looks better and simpler, but also sets the tone of not using it. This seems like a contradiction.
Anyway, I replaced a function call with 2 separate files and a folder. Even has an index.tsx to re-export it to follow all React patterns.

PR #4562, opened 2025-05-19
RI-7123: fix the order of commands stored in workbenchStorage.ts

pd-redis:


KIvanow:
@pd-redis why are so many random FE tests failing?

pd-redis:
not failing anymore

KIvanow (approved):

PR #4566, opened 2025-05-20
RI-7139: Fix json editor styles

pawelangelow:
With this PR https://github.com/RedisInsight/RedisInsight/commit/fe7b8af51b26991ffc945dba5e9a471a2714b05f, I've broken the styles a bit.

Generally, there are a few magics in the JSON editor HTML + CSS structure, like `calc(100%—X)`, `:global()`, or `position: absolute` on elements, that can be aligned without changing their position property. Cleaning this up properly would require a larger refactor. So for now, this PR introduces separate classes for the two editors to apply different styles as a workaround.

Note: header height changed from 115 to 108


KIvanow (approved):
Changes look good. Can confirm the bug doesn't occur any more

PR #4572, opened 2025-05-22
RI-7119 handle resisearch endpoints errors

ArtemHoruzhenko:


redisinsight/api/src/constants/error-messages.ts
```diff
@@ -102,8 +104,10 @@ export default {
     `Required ${module} module is not loaded.`,
   APP_SETTINGS_NOT_FOUND: () => 'Could not find application settings.',
   SERVER_INFO_NOT_FOUND: () => 'Could not find server info.',
-  INCREASE_MINIMUM_LIMIT: (count: string) =>
-    `Set MAXSEARCHRESULTS to at least ${count}.`,
+  INCREASE_MINIMUM_LIMIT: (count: number) =>
```
pd-redis, line 107: If it is possible that count is not provided, shouldn't it be optional?

KrumTy (approved):

redisinsight/api/src/constants/error-messages.ts
ArtemHoruzhenko: you are right! Thanks

PR #4574, opened 2025-05-22
RI-7006: Replace resize related components

dantovska:
Replaced the resize components:
Layout:
- EUI Resizable Container -> Resizable Container - Panel Group
- EUI Resizable Panel -> Resizable Panel - Panel
- Resizable Button -> Resizable Panel Handle - PanelResizeHandle
Where the new components are coming from this library -> react-resizable-panels
Util:
- EUI Resize Observer -> Locally defined Resize Observer

image


redisinsight/ui/src/pages/browser/BrowserPage.tsx
```diff
@@ -304,88 +305,52 @@ const BrowserPage = () => {
...
-            arePanelsCollapsed && isRightPanelOpen && !isBrowserFullScreen,
-        })}
-      >
-        <div className={styles.resizableContainer}>
-          <EuiResizableContainer
-            onPanelWidthChange={onPanelWidthChange}
-            style={{ height: '100%' }}
+      <div className={cx(styles.main)}>
+        <PanelGroup className={styles.panelGroup} direction="horizontal" onLayout={onPanelWidthChange}>
+          <Panel
```
pawelangelow, line 310: I see this is a draft PR, but yet, I wanted to point this out. So, IMO, let's have a custom component that exports `PanelGroup` and `Panel`, and `PanelResizeHandle` in `uiSrc/components/...`. This way we gain:
- common styling in a single place
- if, for some reason, we decide to switch `react-resizable-panels` with something else, it's going to be in a single place

Additionally, we can customise it as per our needs - we know that we have 2 panels and one of them + the handle can be "invisible", so that logic can live there.

jest.config.cjs
```diff
@@ -26,6 +26,8 @@ module.exports = {
     '^uuid$': require.resolve('uuid'),
     msgpackr: require.resolve('msgpackr'),
     'brotli-dec-wasm': '<rootDir>/redisinsight/__mocks__/brotli-dec-wasm.js',
+    'react-resizable-panels':
```
pd-redis, line 29: What does this do?

redisinsight/ui/src/components/base/layout/resize/handle/ResizablePanelHandle.tsx
pd-redis, line 1: Please use styled-components for styling

redisinsight/ui/src/components/base/utils/resize-observer/ResizeObserver.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect, useRef } from 'react'
+
+interface ResizeObserverProps {
+  children: (ref: React.Ref<HTMLDivElement>) => React.ReactNode
+  onResize: (dimensions: { height: number; width: number }) => void
+}
+
+export const ResizeObserver: React.FC<ResizeObserverProps> = ({
```
pd-redis: I would rename this somehow, there is built in class named ResizeObserver
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

redisinsight/ui/src/components/base/utils/resize-observer/ResizeObserver.tsx
```diff
@@ -0,0 +1,34 @@
...
+      onResize({ width, height })
+    })
+
+    observer.observe(element)
+
+    // eslint-disable-next-line consistent-return
+    return () => {
+      observer.disconnect()
+    }
+  }, [onResize])
```
pd-redis: Probably depending on when the ref has an element in it won't hurt
```suggestion
  }, [onResize, containerRef.current])
```

redisinsight/ui/src/components/base/utils/resize-observer/ResizeObserver.tsx
```diff
@@ -0,0 +1,34 @@
...
+    if (!element) return
+
+    const observer = new window.ResizeObserver(([entry]) => {
+      const { width, height } = entry.contentRect
+      onResize({ width, height })
+    })
+
+    observer.observe(element)
+
+    // eslint-disable-next-line consistent-return
```
pd-redis: You can avoid esllint disable by reversing the conditional, i like early returns as well, but in this case I might go the other way
```suggestion
    if (element) {

    const observer = new window.ResizeObserver(([entry]) => {
      const { width, height } = entry.contentRect
      onResize({ width, height })
    })

    observer.observe(element)
}
```

redisinsight/ui/src/pages/browser/components/onboarding-start-popover/OnboardingStartPopover.tsx
```diff
@@ -55,6 +55,7 @@ const OnboardingStartPopover = () => {
       panelClassName={styles.onboardingStartPopover}
       anchorPosition="upCenter"
       data-testid="onboarding-start-popover"
+      style={{ display: 'none' }}
```
pd-redis, line 58: Why do you need this?

redisinsight/ui/src/templates/instance-page-template/InstancePageTemplate.tsx
```diff
@@ -1,7 +1,6 @@
-import React, { useCallback, useEffect, useState } from 'react'
-import cx from 'classnames'
-import { EuiResizableContainer } from '@elastic/eui'
+import React, { useCallback, useEffect, useRef, useState } from 'react'
 import { useSelector } from 'react-redux'
+import { ImperativePanelGroupHandle } from 'react-resizable-panels'
```
pd-redis: Move this as re-export in the resize components please

redisinsight/ui/src/pages/browser/components/onboarding-start-popover/OnboardingStartPopover.tsx
dantovska: The popover adds an empty div to the layout, which was previously used as padding but I don't think we should use the padding from this component, so I'm hiding it. The onboarding popover still shows and it is not dependent on this div.

redisinsight/ui/src/components/base/utils/resize-observer/ResizeObserver.tsx
dantovska: makes sense. now I'm getting the "not all paths return a value" warning, but I guess it is better.

redisinsight/ui/src/components/base/utils/resize-observer/ResizeObserver.tsx
dantovska: maybe RIResizeObserver?

jest.config.cjs
dantovska: The tests were breaking before I moved the react-resizable-panels in a wrapper components. But now since they are not directly used, no need for mocking them. Removing this. Tnx!

redisinsight/ui/src/pages/browser/components/onboarding-start-popover/OnboardingStartPopover.tsx
pd-redis: hm, weird. But I guess this will be changed with replacing EuiPopover

dantovska: It kind of makes sense I think, because the popover is usually "over something", and that is why it expects an element over which to show the popover... Maybe the popover is not the right component here 🤔

PR #4588, opened 2025-05-27
Feature/ri 7091   add an environment variable to skip the eula screen

KIvanow:
This fix a few bugs found in the ticket validation - auto discovery not working and an extra 'privacy policy' link.

I've also updated some texts as an additional request on the ticket.

The autodiscovery is being triggered on `onApplicationBootstrap` because otherwise it completely freezes the process when it tries to start searching for processes for already running databases.

I've also managed to figure out why autodiscovery has never worked for me. It relies on [netstat](https://github.com/RedisInsight/RedisInsight/blob/1e8c0994aec2b0a8a0d2b8c53cfbc526eb5c5729/redisinsight/api/src/modules/database-discovery/utils/autodiscovery.util.ts#L26) and not all linux distributions come with it pre-installed. We also never show any warning or error to the user for that...


ArtemHoruzhenko (changes requested):
In general I think we should find another way to do this

ArtemHoruzhenko (approved):

PR #4593, opened 2025-05-28
RI-7056 replace eui health

pd-redis:
Before / After:

image  image

image  image

image  image

### DIFF
--- a/redisinsight/ui/src/components/base/layout/flex/flex.styles.ts
+++ b/redisinsight/ui/src/components/base/layout/flex/flex.styles.ts
@@ -174,22 +174,24 @@ export type FlexProps = PropsWithChildren &
     full?: boolean
   }
 
-export const StyledFlex = styled.div<FlexProps>`
+export const StyledFlex = styled.div<
+  Omit<FlexProps, 'direction'> & { $direction?: (typeof dirValues)[number] }
+>`
   display: flex;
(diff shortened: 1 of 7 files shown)


redisinsight/ui/src/components/base/layout/flex/flex.styles.ts
```diff
@@ -174,22 +174,24 @@ export type FlexProps = PropsWithChildren &
...
   ${({ align = 'stretch' }) => (align ? flexGroupStyles.align[align] : '')}
-  ${({ direction = 'row' }) =>
-    direction ? flexGroupStyles.direction[direction] : ''}
+  ${({ $direction = 'row' }) =>
+    $direction ? flexGroupStyles.direction[$direction] : ''}
   ${({ justify = 'start' }) =>
     justify ? flexGroupStyles.justify[justify] : ''}
   ${({ centered = false }) => (centered ? flexGroupStyles.centered : '')}
   ${({ responsive = false }) => (responsive ? flexGroupStyles.responsive : '')}
   ${({ wrap = false }) => (wrap ? flexGroupStyles.wrap : '')}
```
pawelangelow, line 191: Question: We prefix the props we don't want to go to the HTML with `$`, right? So why for example `wrap` or `responsive` are without it?

PR #4600, opened 2025-06-02
RI-7070: RI-7072 replace eui text, eui colortext

pd-redis:
First commit is the addition of new components. About 200 files will be added with next few commits :)


redisinsight/ui/src/pages/redis-cluster/RedisClusterDatabasesResult.tsx
```diff
@@ -75,7 +75,7 @@ const RedisClusterDatabasesResult = ({ columns, onBack, onView }: Props) => {
   }
 
   const SummaryText = () => (
-    <EuiText className={styles.subTitle}>
+    <Text className={styles.subTitle}>
```
dantovska: I was wondering what this style does to the text, so i checked the styles.module.scss and it is empty.. we can remove it as part of this PR or some following

dantovska (approved):

redisinsight/ui/src/pages/redis-cluster/RedisClusterDatabasesResult.tsx
pd-redis: I will remove it, but I expect to have lots of similar cases, especially with PR-s like this one. Checking if all the styles are "sane" or can be removed will take a lot of time, and I think it is better to leave it for later stage

PR #4603, opened 2025-06-02
RI-7047: replace eui combo box

pd-redis:
Before:
image

image

After:
image

image


KIvanow (approved):
Code looks good. It works well when tested.

Not sure what the todo comment means aregarding the keys in scope

pd-redis:
> Code looks good. It works well when tested.
> 
> Not sure what the todo comment means aregarding the keys in scope

@KIvanow right now, the `keyDown` handler uses `if (e.key === 'Enter')` which hardcodes the key value
In EUI there is the `keys` object - `import { keys } from '@elastic/eui'` which will also need to be replaced.
After we replace it with whatever works for us, and then replace hardcoded `Enter` with correct reference

PR #4605, opened 2025-06-03
RI-7154: Color Theme select box shown incorrectly

pd-redis:
fix: empty value set for theme if user has not configured it before

### DIFF
--- a/redisinsight/ui/src/pages/settings/components/theme-settings/ThemeSettings.spec.tsx
+++ b/redisinsight/ui/src/pages/settings/components/theme-settings/ThemeSettings.spec.tsx
@@ -42,7 +42,7 @@ describe('ThemeSettings', () => {
     expect(render(<ThemeSettings />)).toBeTruthy()
   })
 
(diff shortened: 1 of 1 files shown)


ArtemHoruzhenko (commented):
Looks like we must add test for this

KIvanow (approved):
Changes look good, but ignore tasks that are out of the sprint ;)

redisinsight/ui/src/pages/settings/components/theme-settings/ThemeSettings.spec.tsx
```diff
@@ -42,6 +42,17 @@ describe('ThemeSettings', () => {
     expect(render(<ThemeSettings />)).toBeTruthy()
   })
 
+  it('should render de default theme option selected when there is no previous config', async () => {
```
pawelangelow:
```suggestion
  it('should render the default theme option selected when there is no previous config', async () => {
```

PR #4607, opened 2025-06-03
RI-7050 replace EUI field number with NumericInput

dantovska:


redisinsight/ui/src/components/input-field-sentinel/InputFieldSentinel.tsx
```diff
@@ -76,12 +74,11 @@ const InputFieldSentinel = (props: Props) => {
...
-        <EuiFieldNumber
+        <NumericInput
           {...clearProp}
-          compressed
-          type="text"
-          value={value}
-          onChange={(e) => handleChange(validateNumber(e.target?.value))}
+          autoValidate
+          value={Number(value)}
+          onChange={(value) => handleChange(value ? value.toString() : '')}
```
pawelangelow, line 81: Question: I'm wondering why we store the value as string and when displaying it we convert it to number? I'm not suggesting editing it, just wondered why.

For example, what if:
```suggestion
          value={value}
          onChange={(value) => handleChange(Number(value ? value.toString() : ''))}
```

dantovska: up until the numeric input, even the eui field number had type text, and every field was represented as string, no matter if it was number. we were using the validateNumber, validatePort and similar methods which apply the validation to strings... 

with introducing the numeric input, we could no longer store the value as string (no option to set the type to text like before), that is why I am using the number. but the handleChange method accepts strings. that is why this conversion is happening

pawelangelow (approved):

redisinsight/ui/src/components/settings-item/SettingItem.tsx
```diff
@@ -94,22 +86,29 @@ const SettingItem = (props: Props) => {
...
+                  placeholder={placeholder}
+                  aria-label={testid?.replaceAll?.('-', ' ')}
+                  className={cx(styles.input, {
+                    [styles.inputEditing]: isEditing,
+                  })}
+                  readOnly={!isEditing}
+                  data-testid={`${testid}-input`}
+                  style={{ width: '100%' }}
+                />
+                {!isEditing && <EuiIcon type="pencil" color="subdued" />}
```
KrumTy: I think this should use base/icons instead of eui

I also think 1 component per icon name is huge boilerplate

KrumTy (approved):
LGTM

redisinsight/ui/src/components/settings-item/SettingItem.tsx
dantovska: Tnx! I changed it to use the redis ui edit icon.

PR #4610, opened 2025-06-05
E2e/ri 7131   е2е tests are failing for both app image and docker

KIvanow:
Not for merging - just for testing at this point

### DIFF
--- a/tests/e2e/pageObjects/browser-page.ts
+++ b/tests/e2e/pageObjects/browser-page.ts
@@ -948,8 +948,14 @@ export class BrowserPage extends InstancePage {
     */
     async selectIndexByName(index: string): Promise<void> {
         const option = Selector(`[data-test-subj="mode-option-type-${index}"]`);
+        const placeholder = Selector('[data-testid="select-index-placeholder"]');
+        const dropdown = Selector('[data-testid="select-search-mode"]');
+
+        // Click placeholder if it exists, otherwise click dropdown
+        const triggerElement = await placeholder.exists ? placeholder : dropdown;
+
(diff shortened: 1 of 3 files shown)


pawelangelow (dismissed):

tests/e2e/helpers/common.ts
```diff
@@ -38,7 +38,17 @@ export class Common {
     }
 
     static async waitForElementNotVisible(elm: Selector): Promise<void> {
```
KrumTy, line 40:
```suggestion
    static async waitForElementNotVisible(elm: Selector, retries = 1): Promise<void> {
```
recursion 💪

KIvanow: I am not sure we need the retries param, as we don't call the same function in the function

KIvanow:
> 121 tests skipped; probably deserves something in backlog

The .skip is a documentation on it's own. We'll handle them when we start porting the tests to playwright. The prioritization there will be dictate by criticality. I also don't think we need all of the e2e tests we already have. These are all things we need to discuss when we start focusing on the e2e testing (which will be after a few more initiatives)

tests/e2e/helpers/common.ts
KrumTy: with retries the logic in the catch can be updated - if retries == 0, then throw, otherwise call the same function with retries - 1

KIvanow: Yes, but at this stage I don't see a point in updating and enhancing the test suites that we'll deprecate and remove in a few months

PR #4611, opened 2025-06-05
RI-7049: replace eui menu with redis menu

dantovska:
- Export menu components
- Replace in Pagination component
image

TODO:
- Tests
- Finalize styling?


pd-redis (approved):
Looks good, but if we happen to need the menu in other places, I would like to simplify the interface and usage

PR #4620, opened 2025-06-09
RI-7059: Replace EUI Link with Redis Link

dantovska:
Replace Link with EuiLink in all components


redisinsight/ui/src/components/instance-header/components/user-profile/styles.module.scss
```diff
@@ -27,6 +27,27 @@
      border: 0 !important;
    }
 
+   .linkOption {
```
pd-redis: I would avoid adding custom scss, if you need to customize specific link, look into `styled(Link)` option instead. I know that we're going to fix up the UI after we finish replacing EUI, but there is no need to add more scss if we don't really need it

redisinsight/ui/src/pages/database-analysis/components/top-namespace/TopNamespace.tsx
```diff
@@ -78,13 +79,12 @@ const TopNamespace = (props: Props) => {
             <Title size="XS">No namespaces to display</Title>
             <p>
               {'Configure the delimiter in '}
-              <EuiLink
-                color="text"
+              <Text
```
pd-redis: I would look into EmptyButton for this case

pd-redis (commented):
In many instances, there is a "color" prop passed to the Link component, its value however is not handled in any way. See `useColorTextStyles` for example how it is handled for Text components and either use that, or implement something similar for Link

redisinsight/ui/src/components/instance-header/components/user-profile/styles.module.scss
dantovska: Since this is a link, which appears in the profile badge and does not have the standard look, I don't think I should add the styles to the component, as this is an exception to the usual usage. Most of the styles are applied to the whole option item, not the link element.

pd-redis: you are not supposed to add it to the component.
You should create specialized version of the link:
```typescript
const UserProfileLink = styled(Link)`
  padding: 8px 12px !important;
  width: 100%;
  color: ${({ theme }: { theme: ReturnType<typeof useTheme> }) =>
    theme.color.azure500} !important;

  &:not(:last-child) {
    border-bottom: 1px solid
      ${({ theme }: { theme: ReturnType<typeof useTheme> }) =>
        theme.color.azure500};
  }

  span {
    width: 100%;

    display: flex;
    align-items: center;
    justify-content: space-between;

    text-decoration: none !important;
    cursor: pointer;
  }
`
```

redisinsight/ui/src/components/base/link/UserProfileLink.tsx
```diff
@@ -0,0 +1,28 @@
+import styled from "styled-components"
+import { useTheme } from "@redis-ui/styles"
+import { Link } from "./Link"
+
+export const UserProfileLink = styled(Link)`
+  padding: 8px 12px !important;
```
pawelangelow, line 6: Question: do we still need these `!important` statements?

redisinsight/ui/src/components/messages/cli-output/cliOutput.tsx
```diff
@@ -105,14 +105,14 @@ export const cliTexts = {
   USE_PROFILER_TOOL: (onClick: () => void) => (
     <ColorText color="danger" key={Date.now()}>
       {'Use '}
-      <EuiLink
+      <Link
```
pd-redis: I don't like the idea of a link without href. If possible, replace with EmptyButton. If making it look right is too much work, leave it, but please leave a comment for the reason why

redisinsight/ui/src/components/instance-header/components/user-profile/UserProfileLink.tsx
pd-redis, line 1: Is this a left over?

redisinsight/ui/src/components/base/link/link.styles.ts
```diff
@@ -0,0 +1,74 @@
...
+        return colors.text.neutral600
+      case 'accent':
+        return colors.text.notice600
+      case 'warning':
+        return colors.text.attention600
+      case 'success':
+        return colors.text.success600
+      default:
+        return color // any supported color value e.g #fff
+    }
```
pawelangelow, line 58: Nit: I like the map syntax for such features:

```ts
const colorMap: Record<string, string> = {
    inherit: 'inherit',
    default: colors.text.primary500,
    primary: colors.text.primary500,
    text: colors.text.neutral700,
    subdued: colors.text.informative400,
    danger: colors.text.danger600,
    ghost: colors.text.neutral600,
    accent: colors.text.notice600,
    warning: colors.text.attention600,
    success: colors.text.success600,
  }

const getColorValue = (color?: ColorType) => {
  if (!color) {
    return colors.text.primary500
  }

  return colorMap[color] ?? color
}
```

IMO its less verbose. We can define it outside `useColorTextStyles` as well.

pawelangelow (approved):

redisinsight/ui/src/components/base/link/UserProfileLink.tsx
dantovska: Unfortunately we do in this case... I tried removing it but there are some inherited styles that break it.

pawelangelow (approved):

PR #4622, opened 2025-06-10
RI-7068: replace EuiSwitch with SwitchInput

KrumTy:


redisinsight/ui/src/components/auto-refresh/AutoRefresh.spec.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
 import { instance, mock } from 'ts-mockito'
+import userEvent from '@testing-library/user-event'
```
pd-redis: I would reexport it from test-utils, just for consistency

redisinsight/ui/src/components/base/inputs/SwitchInput.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+
+import { Switch } from '@redis-ui/components'
+
+const SwitchInput: React.FC<React.ComponentProps<typeof Switch>> = ({
+  style,
+  ...props
+}) => (
```
pd-redis: So far we follow the pattern to define props above the component and type them in the arguments list
```suggestion
type SwitchInputProps = React.ComponentProps<typeof Switch
const SwitchInput = ({
  style,
  ...props
}: SwitchInputProps) => (
```

redisinsight/ui/src/packages/redistimeseries-app/src/components/Chart/ChartConfigForm.tsx
```diff
@@ -49,17 +49,17 @@ export default function ChartConfigForm(props: ChartConfigFormProps) {
           selected={value.mode}
           onClick={(v) => onChange('mode', v)}
         />
-        <EuiSwitch
-          compressed
-          label={<span className="switch-staircase-label">Staircase</span>}
+        <SwitchInput
+          titleOn="Staircase"
```
pd-redis: Mybe it is worth wrapping redis-ui switch to avoid this repetition?

KrumTy: I think it makes sense to have an on/off label explicitly and adding such wrapper would introduce odd behaviour when combining props

pd-redis: How? You can have `title` and `titleOff` prop, and if the latter is not provided, just use the `title` for `titleOff`. How is that odd?

KrumTy: I've adjusted it as requested; we don't use the Off prop currently anyway

redisinsight/ui/src/components/base/text/Title.tsx
pd-redis, line 1: Is there a reason to delete this?

redisinsight/ui/src/components/consents-settings/ConsentsSettings.tsx
```diff
@@ -27,6 +21,7 @@ import { CallOut } from 'uiSrc/components/base/display/call-out/CallOut'
 import { Text } from 'uiSrc/components/base/text'
 import ConsentOption from './ConsentOption'
 
+import { SwitchInput } from '../base/inputs'
```
pd-redis: Why relative import?

KrumTy: auto imported it that way I guess and I missed it

PR #4625, opened 2025-06-10
[RI-7069]: Replace EuiTabs with Tabs

KrumTy:


redisinsight/ui/src/components/instance-header/components/instances-navigation-popover/InstancesNavigationPopover.tsx
```diff
@@ -99,6 +93,22 @@ const InstancesNavigationPopover = ({ name }: Props) => {
     )
   }
 
+  const tabs: TabInfo[] = useMemo(
```
pd-redis, line 96: I don't see much benefit in useMemo in this case. There is no calculation or expensive call made, that needs memoizing.

redisinsight/ui/src/components/home-tabs/HomeTabs.spec.tsx
```diff
@@ -114,7 +111,7 @@ describe('HomeTabs', () => {
     render(<HomeTabs />)
 
     expect(
-      screen.queryByTestId('home-tab-rdi-instances'),
+      screen.queryByTestId('Redis Data Integration'),
```
dantovska: why not use the id? or change to getByText?

dantovska (approved):

redisinsight/ui/src/components/base/layout/tabs/index.ts
```diff
@@ -0,0 +1,4 @@
+import { Tabs, TabInfo } from '@redis-ui/components'
+
+export default Tabs
```
pd-redis, line 3: Why export this as default export? Everywhere else we use named exports

redisinsight/ui/src/styles/elastic.css
pd-redis, line 1: I don't think there is much sense to edit this file as we go, the idea is to simply remove it once eui is replaced.
I don't mind, just noting.

PR #4626, opened 2025-06-10
RI-7166: ReJSON fixes

pawelangelow:
This PR changes when we show the Monaco editor icon - when the JSON is below a threshold, instead of when `envDependent` is off. That threshold was previously based on the root-level keys, but now it's based on the size and can be configured with env variable `RI_REJSON_MONACO_EDITOR_MAX_THRESHOLD`.

### DIFF
--- a/redisinsight/ui/src/config/default.ts
+++ b/redisinsight/ui/src/config/default.ts
@@ -84,6 +84,10 @@ export const defaultConfig = {
       'RI_DATABASE_OVERVIEW_MINIMUM_REFRESH_INTERVAL',
       1,
     ),
+    rejsonMonacoEditorMaxThreshold: intEnv(
+      'RI_REJSON_MONACO_EDITOR_MAX_THRESHOLD',
+      10_000,
+    ),
   },
   features: {
     envDependent: {
--- a/redisinsight/ui/src/slices/browser/keys.ts
+++ b/redisinsight/ui/src/slices/browser/keys.ts
@@ -2,6 +2,7 @@ import { createSlice, PayloadAction } from '@reduxjs/toolkit'
 import { cloneDeep, remove, get, isUndefined } from 'lodash'
 import axios, { AxiosError, CancelTokenSource } from 'axios'
 import { apiService, localStorageService } from 'uiSrc/services'
+import { getConfig } from 'uiSrc/config'
 import {
   ApiEndpoints,
   BrowserStorageItem,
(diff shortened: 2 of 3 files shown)


redisinsight/ui/src/slices/browser/keys.ts
```diff
@@ -808,6 +810,11 @@ export function fetchKeyInfo(
       if (data.type === KeyTypes.ReJSON) {
         dispatch<any>(fetchReJSON(key, '$', data.length, resetData))
         dispatch<any>(setEditorType(EditorType.Default))
+        dispatch<any>(
+          setIsWithinThreshold(
+            data.size <= riConfig.browser.rejsonMonacoEditorMaxThreshold,
+          ),
+        )
```
ArtemHoruzhenko, line 817: I don't think we must determine it here. You can see above `dispatch<any>(fetchReJSON(key, '$', data.length, resetData))` which, I assume, has own logic when to fetch entire JSON or partially. Not sure but seems like it is possible that we might show monaco editor partial JSON response from API. https://github.com/RedisInsight/RedisInsight/pull/4626/files#diff-ac43c1e42223492126a9f0acf6beb804db1905bfc66d3525967df9860ad125ebR15 

I feel like we must have something like this:
1. determine if we can fetch entire json and store this into flag. (E.g. length is less than 10_000 and size less than 1mb, and store flag `isForceLoaded`)
2. Then for monaco we can rely on `isForceLoaded` and any additional condition needed. But `isForceLoaded` flag will ensure that we won't show custom schema returned for partial responses

redisinsight/ui/src/config/default.ts
```diff
@@ -84,6 +84,10 @@ export const defaultConfig = {
       'RI_DATABASE_OVERVIEW_MINIMUM_REFRESH_INTERVAL',
       1,
     ),
+    rejsonMonacoEditorMaxThreshold: intEnv(
+      'RI_REJSON_MONACO_EDITOR_MAX_THRESHOLD',
+      10_000,
+    ),
```
ArtemHoruzhenko, line 90: nit but does it make sense to add keyword like SIZE or BYTES or something else to the env variable name?
also I thought that we want to make length threshold configurable also

KIvanow (approved):

redisinsight/ui/src/slices/browser/keys.ts
ArtemHoruzhenko: nvm, we are refetching data based on editor type

pawelangelow:
For the record, we discussed @ArtemHoruzhenko's suggestions, and they will be addressed in upcoming PRs.

PR #4631, opened 2025-06-17
RI-7060: Replace EUI Loading Spinner with Redis Loader

dantovska:


redisinsight/ui/src/components/base/display/index.ts
```diff
@@ -0,0 +1 @@
+export { default as Loader } from './loader/Loader'
```
pd-redis, line 1: Redundant export

redisinsight/ui/src/components/base/display/loader/Loader.tsx
```diff
@@ -0,0 +1,29 @@
...
+      return '32px'
+    default:
+      return '16px'
+  }
+}
+
+export default function Loader({ size, ...rest }: RedisLoaderProps) {
+  const sizeInPx = size ? convertSizeToPx(size) : '16px'
+  return <RedisLoader size={sizeInPx} {...rest} />
+}
```
pd-redis: Instead of hadcoding values, we can use theme values
```suggestion
const convertSizeToPx = (tShirtSize: string, theme) => {
  const {space} = theme.core
  switch (tShirtSize.toLowerCase()) {
    case 's':
      return space.space050
    case 'm':
      return space.space100
    case 'l':
      return space.space250
    case 'xl':
      return space.space300
    default:
      return space.space100
  }
}

export default function Loader({ size, ...rest }: RedisLoaderProps) {
  const theme = useTheme()
  const sizeInPx = size ? convertSizeToPx(size, theme) : '16px'
  return <RedisLoader size={sizeInPx} {...rest} />
}
```

redisinsight/ui/src/components/base/display/index.ts
dantovska: I see that we are doing this for some group components (layout, inputs) and not for others... I suggest we decide this at the end and make it consistent - either have the index.ts or not for all components.

KrumTy: imo we should ultimately unify all base components under a single barrel file as all libraries do (for convenience of use and no name collisions)

redisinsight/ui/src/components/base/display/loader/Loader.tsx
```diff
@@ -0,0 +1,45 @@
...
+  space150: string;
+  space200: string;
+  space250: string;
+  space300: string;
+  space400: string;
+  space500: string;
+  space550: string;
+  space600: string;
+  space800: string;
+}
```
KrumTy:
```suggestion
import { useTheme, theme } from '@redis-ui/styles'

type Space = typeof theme.core.space
```

this might work

pd-redis: or
```typescript
type Space = ReturnType<typeof useTheme>['core']['space']

```

redisinsight/ui/src/components/base/display/index.ts
pd-redis: IMO we should not use barrel files unless they make sense. Exporting single component deom barell file does not make sense

pd-redis (approved):

PR #4633, opened 2025-06-18
RI-7040: replace eui icon

pd-redis:
Sorry about number of changes, but we use a lot of icons


redisinsight/ui/src/components/base/icons/common/index.tsx
```diff
@@ -0,0 +1,260 @@
...
+export const CloudLinkIcon = (props: IconProps) => (
+  <Icon icon={CloudLinkSvg} {...props} isSvg />
+)
+
+export const ConnectionIcon = (props: IconProps) => (
+  <Icon icon={ConnectionSvg} {...props} isSvg />
+)
+
+export const CopilotIcon = (props: IconProps) => (
+  <Icon icon={CopilotSvg} {...props} isSvg />
```
KrumTy, line 91: props can have "isSvg" but it will always be overridden  by isSvg={true}

not sure if that's intentional

redisinsight/ui/src/components/base/icons/index.ts
```diff
@@ -1,31 +1,9 @@
...
-export { ThreeDotsIcon } from './ThreeDotsIcon'
-export { Trigger } from './Trigger'
-export { UserInCircle } from './UserInCircle'
 
+export * from './common'
+export * from './guides'
+export * from './metrics'
 export * from './modules'
 export * from './options'
 export * from './sidebar'
```
KrumTy, line 9: do we need to export all these (presumably internal) components from icons/?

redisinsight/ui/src/components/base/icons/modules/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+export const RedisGears2LightIcon = (props: IconProps) => (
+  <Icon icon={RedisGears2Light} {...props} isSvg />
+)
+
+export const RedisGearsDarkIcon = (props: IconProps) => (
+  <Icon icon={RedisGearsDark} {...props} isSvg />
+)
+
+export const RedisGearsLightIcon = (props: IconProps) => (
```
KrumTy, line 64: why not have a similar build icons script as done in [@redis-ui/icons](https://github.com/redislabsdev/redis-ui/blob/master/packages/icons/scripts/build-icons.js)? That should save a lot of boilerplate and will automate adding future icons (just adding an svg would be enough)

So the process would be: 
1. add an svg icon
2. run the build-icons script
3. the script will create a new component, based on svg icon's name and it will be included in the RiIcon + typed automatically

redisinsight/ui/src/components/base/icons/RiIcon.tsx
```diff
@@ -0,0 +1,76 @@
...
+  if (!IconType) {
+    console.warn(`Icon type "${type}" not found, rendering as image`)
+    // TODO - 17.06.25 - Replace with icon
+    //  There are a few cases where type is just imported image asset. In most cases, it seems
+    //  that the image is an svg in the plugins folder - http://localhost:5540/static/plugins/redisearch/./dist/table_view_icon_light.svg
+    //  we can either just scratch the plugins and move assets in to the main project, or look into dynamically loading as icons in runtime
+    return (
+      <img
+        {...(props as ImgHTMLAttributes<HTMLImageElement>)}
+        alt={props.title ? props.title : ''}
```
KrumTy, line 44:
```suggestion
        alt={props.title || ''}
```

redisinsight/ui/src/components/base/icons/RiIcon.tsx
```diff
@@ -0,0 +1,76 @@
...
+      iconSize = 'XL'
+      break
+    case 'original':
+      iconSize = null
+      break
+    case 'l':
+    default:
+      iconSize = 'L'
+  }
+  // @ts-ignore
```
KrumTy, line 74:
```suggestion
    case 'original':
      iconSize = undefined
      break
    case 'l':
    default:
      iconSize = 'L'
  }
```
any way to remove the ts-error?

redisinsight/ui/src/components/base/icons/common/index.tsx
pd-redis: For SVG icons it is intentional, yes

redisinsight/ui/src/components/base/icons/index.ts
pd-redis: they are not internal, they are used everywhere in RiIcon component, this is just some structure as they were found scattered  around the codebase

redisinsight/ui/src/components/base/icons/modules/index.tsx
pd-redis: It sounds like a good small project for the future

redisinsight/ui/src/components/base/icons/RiIcon.tsx
pd-redis: `null` here is also intentional.

PR #4640, opened 2025-06-20
[RI-7058] Replace EuiInMemoryTable with Table

KrumTy:
## Description

Replaces EuiInMemoryTable with Table component across the app


redisinsight/ui/src/components/shortcuts-flyout/ShortcutsFlyout.tsx
```diff
@@ -12,44 +10,42 @@ import {
...
 
-  const tableColumns: EuiBasicTableColumn<any>[] = [
+  const tableColumns: ColumnDefinition<any>[] = [
     {
-      name: '',
-      field: 'description',
-      width: '60%',
+      header: 'Description',
+      id: 'description',
+      accessorKey: 'description',
```
pd-redis, line 26: From what I see in tanstack table docs, `accessorKey` is optional, and from what I see in usages here, it is 100% of the time the same as `id` field. Have you tested if the tables work with just id?

redisinsight/ui/src/pages/rdi/statistics/data-streams/DataStreams.tsx
```diff
@@ -29,92 +29,104 @@ interface Props {
...
+      <EuiToolTip content={getValue<string>()}>
+        <span>{formatLongName(getValue<string>(), 30, 0, '...')}</span>
+      </EuiToolTip>
+    ),
+  },
+  {
+    header: 'Total',
+    id: 'total',
+    accessorKey: 'total',
+    enableSorting: true,
```
pd-redis, line 48: it seems that enableSorting is true by default

redisinsight/ui/src/components/shortcuts-flyout/ShortcutsFlyout.tsx
KrumTy: I've tested without and cell items disappear in columns
it's probably optional since you can also have accessorFn
anyway - doesn't seem to default to "access the field that corresponds to the id"

redisinsight/ui/src/pages/rdi/statistics/data-streams/DataStreams.tsx
KrumTy: it is but I prefer having it set explicitly, per column, as done previously
(I believe it was false by default with the previous implementation)

PR #4645, opened 2025-06-23
RI-7066: replace eui radio group and eui SuperSelect, RI-7067

pd-redis:
Replace RadioGroup and Select component.
Mixed together because there are like 3 instances of radio groups, and already starting updating Select in this branch and it is easier to review together than to split PRs

Before:
image
image
image
image

After:
image
image
image
image


redisinsight/ui/src/components/oauth/oauth-select-account-dialog/OAuthSelectAccountDialog.tsx
```diff
@@ -192,13 +191,18 @@ const OAuthSelectAccountDialog = () => {
...
+            value={formik.values.accountId ?? ''}
             onChange={(id) => handleChangeAccountIdFormat(id)}
-            name="radio accounts group"
-          />
+          >
+            {radios.map(({ id, label }) => (
+              <RiRadioGroup.Item.Compose value={id} key={id}>
+                <RiRadioGroup.Item.Indicator />
+                <RiRadioGroup.Item.Label>{label}</RiRadioGroup.Item.Label>
+              </RiRadioGroup.Item.Compose>
```
dantovska: I think that instead of following the Redis Component syntax, it would be good to simplify the component names where they are imported and exported (in this case  `base/forms/radio-group/RadioGroup.tsx`)
Looking at it, maybe this file should not be a tsx file but simply ts file (index.ts)?

A good example IMO is the case with the drawer component `base/layout/drawer/index.ts`:

```
import { Drawer } from '@redis-ui/components'

const DrawerHeader = Drawer.Header
const DrawerBody = Drawer.Body
const DrawerFooter = Drawer.Footer

export { Drawer, DrawerHeader, DrawerBody, DrawerFooter }
```

redisinsight/ui/src/components/base/forms/radio-group/RadioGroup.tsx
```diff
@@ -0,0 +1 @@
+export { RadioGroup as RiRadioGroup } from '@redis-ui/components'
```
pawelangelow, line 3: Question: do we need the `Ri` prefix?

pd-redis: Because autocomplete will directly import it and not ask me if i want to import from `@redis-ui/components` or `uiSrc/components/base/forms/radio-group/RadioGroup`

redisinsight/ui/src/components/oauth/oauth-select-account-dialog/OAuthSelectAccountDialog.tsx
pd-redis: I am inclined to think about some different approach, if we have more usage for this, but with just 2 places where we need the composition, I think it is just some overhead without much benefit.

pd-redis: @dantovska did what you proposed, please take a look and let me know if you are ok with it?

redisinsight/ui/src/components/base/forms/radio-group/RadioGroup.tsx
KrumTy: maybe we should do this for all components then

redisinsight/ui/src/components/command-helper/components/command-helper-search/CHSearchFilter/CHSearchFilter.tsx
```diff
@@ -41,56 +40,64 @@ const CHSearchFilter = ({ submitFilter, isLoading }: Props) => {
...
   return (
-    <OutsideClickDetector onOutsideClick={() => setIsSelectOpen(false)}>
-      <div className={cx(styles.container)}>
-        {!typeSelected && (
-          <div
-            className={styles.allTypes}
-            onClick={() => !isLoading && setIsSelectOpen(!isSelectOpen)}
-            role="presentation"
-          >
+    <div className={cx(styles.container)}>
```
KrumTy, line 75: don't need cx

redisinsight/ui/src/pages/home/components/add-database-screen/components/connectivity-options/ConnectivityOptions.tsx
```diff
@@ -26,6 +28,43 @@ export interface Props {
...
+  min-width: 160px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  text-decoration: none !important;
+  & * {
+    text-decoration: none !important;
+  }
+  position: relative;
+  width: 100%;
```
KrumTy, line 41: is all this style based on new design or taken from some scss file?

KrumTy (approved):
overall LGTM

redisinsight/ui/src/components/command-helper/components/command-helper-search/CHSearchFilter/CHSearchFilter.tsx
pd-redis: Fixed

redisinsight/ui/src/pages/home/components/add-database-screen/components/connectivity-options/ConnectivityOptions.tsx
pd-redis: It is from the css. Redis-ui buttons logically don't work as links, so some style updates were needed.

PR #4648, opened 2025-06-25
RI-7160 Stop fetching key details for key that was opened in another …

ArtemHoruzhenko:
…database after switch

RI-7132 fix duplicated calls on refresh on the details view for a json key

### DIFF
--- a/redisinsight/ui/src/pages/browser/modules/key-details/components/change-editor-type-button/useChangeEditorType.spec.ts
+++ b/redisinsight/ui/src/pages/browser/modules/key-details/components/change-editor-type-button/useChangeEditorType.spec.ts
@@ -2,7 +2,7 @@ import * as reactRedux from 'react-redux'
 import { renderHook, act } from '@testing-library/react-hooks'
 import { EditorType } from 'uiSrc/slices/interfaces'
 import { FeatureFlags } from 'uiSrc/constants'
-
+import { stringToBuffer } from 'uiSrc/utils'
 import { useChangeEditorType } from './useChangeEditorType'
 
 jest.mock('react-redux', () => ({
@@ -10,8 +10,14 @@ jest.mock('react-redux', () => ({
   useSelector: jest.fn(),
 }))
 
+jest.mock('uiSrc/slices/browser/rejson', () => ({
+  ...jest.requireActual('uiSrc/slices/browser/rejson'),
+  fetchReJSON: jest.fn((key) => ({ type: 'FETCH_REJSON', payload: key })),
+}))
+
 const mockedUseDispatch = reactRedux.useDispatch as jest.Mock
 const mockedUseSelector = reactRedux.useSelector as jest.Mock
+const mockKeyName = stringToBuffer('test-key')
 
 describe('useChangeEditorType', () => {
   const dispatchMock = jest.fn()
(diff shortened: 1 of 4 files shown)


redisinsight/ui/src/pages/browser/modules/key-details/components/change-editor-type-button/useChangeEditorType.tsx
```diff
@@ -2,22 +2,34 @@ import { useCallback } from 'react'
...
 
 export const useChangeEditorType = () => {
   const dispatch = useDispatch()
   const { editorType, isWithinThreshold } = useSelector(rejsonSelector)
   const { [FeatureFlags.envDependent]: envDependentFeature } = useSelector(
     appFeatureFlagsFeaturesSelector,
   )
+  const {
+    name: selectedKey,
+  } = useSelector(selectedKeyDataSelector) || {}
```
pawelangelow: I guess `|| {}` is to suppress a TS error. With this fallback, tho, we might hide some selector problems - if we change something and somehow cause it return `null` or other falsy values, `selectedKey` would be undefined, which is OK for this logic, but I'm a fan of a more straightforward approach - if something unexpected fails, it should fail. If we know that this can be `undefined`, let's wait for just it (the `name` property), not the whole selector.

```suggestion
  const selectedKey = useSelector(selectedKeyDataSelector)?.name
```

redisinsight/ui/src/pages/browser/modules/key-details/components/change-editor-type-button/useChangeEditorType.spec.ts
```diff
@@ -59,6 +65,50 @@ describe('useChangeEditorType', () => {
...
   })
 
+  it('should fetch json when type switched', async () => {
+    mockedUseSelector.mockReturnValue({
+      editorType: EditorType.Default,
+    }).mockReturnValue({
+      [FeatureFlags.envDependent]: { flag: false },
+    }).mockReturnValue({
+      name: mockKeyName,
+    })
```
pawelangelow, line 75: This is a nitpicky comment, but I just wanted to share it 😄 

I know that I introduced this, but probably we need to migrate it to something like:

```
const setupUseSelectorMock = (overrides = {}) => {
  mockedUseSelector.mockImplementation((selector) => {
    if (selector === rejsonSelector) {
      return overrides.rejsonSelector ?? { editorType: EditorType.Default, isWithinThreshold: true }
    }
    if (selector === appFeatureFlagsFeaturesSelector) {
      return overrides.featureFlags ?? { [FeatureFlags.envDependent]: { flag: false } }
    }
    if (selector === selectedKeyDataSelector) {
      return overrides.selectedKeyData ?? { name: mockKeyName }
    }
    return undefined
  })
}
```
The whole effort would be to not rely on execution order, but selector, making the tests a bit more robust.

pawelangelow (dismissed):

redisinsight/ui/src/pages/browser/modules/key-details/components/change-editor-type-button/useChangeEditorType.spec.ts
ArtemHoruzhenko: I would say that we can go even further have jest.fn() in each if realization and have this `setupUseSelectorMock` global per file or even better per app and rewrite responses on tests runtime when needed

pawelangelow (approved):

PR #4652, opened 2025-06-30
Bugfix/cluster info handle ipv6

Sylvain-Royer:
This is a renamed branch for [this PR](https://github.com/RedisInsight/RedisInsight/pull/4632).

Previously, the code assumed that all hosts contained only 1 :, where the left side was the host and the right was the port. Unfortunately, this breaks for IPv6. Instead, this code will find the last : and have the host be the last side, port be right.

This is related to https://github.com/RedisInsight/RedisInsight/issues/1429.

I've tested this on my local machine using docker compose set up in ipv6 only, as well as on my company's ipv6 only Kubernetes cluster. This is confirmed working on my end.


pawelangelow (approved):
Thanks, again @Sylvain-Royer and sorry for the extra work with the branch!

Sylvain-Royer:
> Thanks, again @Sylvain-Royer and sorry for the extra work with the branch!

np, thank you for the quick turnaround!

PR #4659, opened 2025-07-01
[RI-7074] Replace EuiToolTip with RiTooltip

KrumTy:


redisinsight/ui/src/packages/ri-explain/src/Node.tsx
```diff
@@ -99,9 +100,9 @@ export function ProfileNode(props: INodeProps) {
     <div className="ProfileContainer" id={`node-${id}`}>
       <div className="Main">
         <div className="InfoData">
-          <EuiToolTip delay="long" content={infoData}>
+          <RiTooltip delay={500} content={infoData}>
```
valkirilov: I see this magic number on a few places and I wonder whether the new component comes with some predefined delays, like EUI?

Otherwise, we may put this magic number behind a shared constant (something like `TOOLTIP_DELAY_LONG=500`) so we can easily update it in the future, if necessary.

valkirilov (approved):

redisinsight/ui/src/components/base/tooltip/index.tsx
pd-redis, line 1: Why this file is named `index.tsx`? Its not very consistent with other components

redisinsight/ui/src/components/base/tooltip/tooltip.styles.ts
```diff
@@ -0,0 +1,14 @@
+import styled from 'styled-components'
+import { Text } from 'uiSrc/components/base/text'
+
+export const ContentWrapper = styled.div`
```
pd-redis: This looks like it could be the `Col` component from flex-layouts

redisinsight/ui/src/components/base/tooltip/tooltip.styles.ts
```diff
@@ -0,0 +1,14 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
```
pd-redis: And if it is not the `Col` component, then use the `theme` and use some of the spacing values there

redisinsight/ui/src/components/base/tooltip/index.tsx
KrumTy: we already have a bunch of <folder>/index.tsx files in /base/*

I'd say we handle it (decide on name pattern) separately in the main branch for all components, along with adding "Ri" prefix to them.
Another thing I'd suggest (open for debate) would be exposing all usable base ui components straight from /base/index.ts

pd-redis: Re-exporting a bunch of components in their index.ts, and then all toghether from another index.ts means, that each time we add a new component, we should remember to re-export it from all indexes. In my opinion this is just added overhead without much benefit. But I am open for discussion about this as well

redisinsight/ui/src/components/base/tooltip/HoverContent.tsx
```diff
@@ -1,15 +1,16 @@
...
 interface RiTooltipContentProps {
   title?: React.ReactNode
   content: React.ReactNode
 }
 
 export const HoverContent = ({ title, content }: RiTooltipContentProps) => (
-  <ContentWrapper>
-    {title && <TitleWrapper>{title}</TitleWrapper>}
+  <Col>
+    {title && <Title size="S">{title}</Title>}
```
pd-redis, line 13: I wouldn't wrap it in Title, unless it is a string. Title uses H1-2-3-... tags and whatever is passed to HoverContent might not be compatible or intended to be in a heading. Just something to consider

redisinsight/ui/src/components/base/tooltip/index.tsx
KrumTy: imagine what using an external (ui) library would be like if each component had a different namespace/import path

single path, e.g. 'uiSrc/components/base' would provide 
- easier import (all base components, under one path)
- not having ~10 extra randomly placed 'uiSrc/' imports on top of each component file
- guarantee of no naming collisions since all needs to go through 1 barrel file

redisinsight/ui/src/components/base/tooltip/HoverContent.tsx
KrumTy: I've tested and it works fine even with nested jsx; Title's children are typed as ReactNode too

dantovska (approved):

PR #4661, opened 2025-07-02
Update Node.js default runtime

valkirilov:
# Description

Update the default Node.js runtime constraint to v22

* updated the [wiki docs](https://github.com/RedisInsight/RedisInsight/wiki/How-to-build-and-contribute#prerequisites) with the instructions for the prerequisite software before getting started with the installation
```
$ yarn install

error @nestjs/core@11.0.20: The engine "node" is incompatible with this module. 

Expected version ">= 20". Got "18.20.8"
error Found incompatible module.
```

* update the default Node.js runtime version in the GitHub Actions workflow
* update the default Node.js runtime version in the `package.json`
* update the engine check to actually use the official way (according to [npm](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) and [yarn](https://classic.yarnpkg.com/en/docs/package-json/#toc-engines), it should be `engines`, not `devEngines`)
* added [.nvmrc](https://github.com/nvm-sh/nvm?tab=readme-ov-file#nvmrc) with default Node.js version for easier setup


.nvmrc
```diff
@@ -0,0 +1 @@
+22
```
KrumTy: worth specifying exact version here as well? If we target `22.11.0` in all places it will be easier to update it in future without missing something

valkirilov: For the GitHub Action workflow I kept it with fixed version number, because I know it's a good practice to reduce the risk of issues in the pipeline in case a new version is published and automatically picked up.

For the `.nvmrc` I was not sure whether we want the exact version, since it's for development purposes and we can risk it with the latest available one. Of course, it's not a problem, I just updated it for the sake of consistency.

pawelangelow (dismissed):
🚀

KrumTy (approved):

dantovska (approved):

package.json
pd-redis, line 1: I would keep devEngines and simply add engines field. With this complicated setup you don't know which tool might be using this

valkirilov: Yes, I was a bit hesitant to touch it too, that's why I asked whether it was safe to change it.

Anyway, I've applied your suggestion, and now we have both approaches in place:
- the official way of limiting the Node.js runtime using `engines`
- and the custom `devEngines` fallback, just in case it's still used by some tooling we're not aware of

Thanks for the input 🙂

pawelangelow (approved):

KrumTy (approved):

francislavoie:
The Dockerfile wasn't updated to match, so building the Docker image now fails. 

```
$ docker build -t redis-insight .
...
...
=> ERROR [build  8/18] RUN SKIP_POSTINSTALL=1 yarn install    0.5s
------
 > [build  8/18] RUN SKIP_POSTINSTALL=1 yarn install:
0.329 yarn install v1.22.22
0.438 [1/5] Validating package.json...
0.442 error redisinsight@: The engine "node" is incompatible with this module. Expected version ">=22.x". Got "20.19.4"
0.449 error Found incompatible module.
```

#4864 

https://github.com/redis/RedisInsight/blob/1a9f5703ced74bd2417ec34304245e6c149a179a/Dockerfile#L9

My workaround is to stick with `2.70.0` for the time being.

valkirilov:
> The Dockerfile wasn't updated to match, so building the Docker image now fails.
> 
> ```
> $ docker build -t redis-insight .
> ...
> ...
> => ERROR [build  8/18] RUN SKIP_POSTINSTALL=1 yarn install    0.5s
> ------
>  > [build  8/18] RUN SKIP_POSTINSTALL=1 yarn install:
> 0.329 yarn install v1.22.22
> 0.438 [1/5] Validating package.json...
> 0.442 error redisinsight@: The engine "node" is incompatible with this module. Expected version ">=22.x". Got "20.19.4"
> 0.449 error Found incompatible module.
> ```
> 
> #4864
> 
> https://github.com/redis/RedisInsight/blob/1a9f5703ced74bd2417ec34304245e6c149a179a/Dockerfile#L9
> 
> My workaround is to stick with `2.70.0` for the time being.

Thanks for pointing this out @francislavoie, the issue should be resolved now, the Docker build is fixed by https://github.com/redis/RedisInsight/pull/4878

PR #4663, opened 2025-07-03
RI-7065: Replace EUI Progress with custom Progress Bar Loader

dantovska:
Create a custom Progress Bar Loader with animation and replace the EUI Progress. 

The component is used everywhere as a basic loader, just has a vision like a progress bar (the blue-ish line below the key results):
Screenshot 2025-07-04 at 13 56 47


redisinsight/ui/src/components/base/layout/progress-bar/ProgressBarLoader.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
```
pd-redis: Nitpick, but I don't see the progress bar as a `layout` component, maybe more of `display` component?

redisinsight/ui/src/components/base/layout/progress-bar/progress-bar-loader.styles.ts
```diff
@@ -0,0 +1,45 @@
+import { theme } from '@redis-ui/styles'
+import styled, { css, keyframes } from 'styled-components'
+
+export const LoaderContainer = styled.div<
+  React.HtmlHTMLAttributes<HTMLDivElement>
```
pd-redis: I don't see import for React, doesn't eslint complain about that?

pd-redis: You only use (allow) style and className for props to this component, maybe it is worth to narrow down the props?

redisinsight/ui/src/components/base/layout/progress-bar/progress-bar-loader.styles.ts
```diff
@@ -0,0 +1,45 @@
...
+>`
+  position: relative;
+  height: 3px;
+  background-color: #e6e6e6;
```
pd-redis: I am guessing that this will be migrated to the "theme"?

valkirilov (approved):

redisinsight/ui/src/components/base/layout/progress-bar/progress-bar-loader.styles.ts
dantovska: Do you mean reuse the useTextStyle logic ?

redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  const barBackgroundColors: Map<ColorType, string> = new Map([
+    ['inherit', 'inherit'],
+    ['default', themeColors.background.primary300],
+    ['primary', themeColors.background.primary300],
+    ['danger', themeColors.background.danger600],
+    ['warning', themeColors.background.attention600],
+    ['success', themeColors.background.success600],
+  ])
```
pd-redis: This seems way too complicated. Is there any reason to use map?
```suggestion
  const barBackgroundColors: Record<ColorType, string> = {
    inherit: 'inherit',
    default: themeColors.background.primary300,
    primary: themeColors.background.primary300,
    danger: themeColors.background.danger600,
    warning: themeColors.background.attention600,
    success: themeColors.background.success600,
  }
```

redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  return barBackgroundColors.get(color) ?? color
```
pd-redis:
```suggestion
  return barBackgroundColors[color] ?? color
```

redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
```diff
@@ -0,0 +1,90 @@
...
+interface LoaderContainerProps {
+  children?: ReactNode
+  style?: React.CSSProperties
+  className?: string
+}
+
+export const LoaderContainer = styled.div<LoaderContainerProps>`
+  position: relative;
+  height: 3px;
+  background-color: #e6e6e6;
```
pd-redis: this is still hardcoded, if it is irrelevant for some reason, just remove it, if it is seen by the user, please use theme colors

redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
```diff
@@ -0,0 +1,90 @@
...
+  overflow: hidden;
+  border-radius: 2px;
+`
+
+export const LoaderBar = styled.div<MapProps>`
+  ${useColorBackgroundStyles}
```
KrumTy: I've never seen such hook usage before

dantovska: Is there something wrong with this? I've seen it in our code, so please suggest what better thing to use.

redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
dantovska: As we spoke offline, I plan to get this logic outside in a util, and pass this map as a parameter. This map would be different for link, text, backgrounds etc. Why is it complicated, it is pretty much the same as the switch case...

redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
KrumTy: Ok so I've read more on the topic and it turns out this is not actually a hook, so the "use.." prefix is misleading. It is not attached to React's lifecycle, useColorBackgroundStyles triggers once - during render. You can't use hooks inside (like useEffect/useState). It just so happens that useTheme, which you use inside, is synchronous and simply access the theme context so it works.

tl;dr; if you inject a function like that in a styled component, it is not a hook. It should be synchronous and should return valid css. I'd replace the "use" prefix with "get".

redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
dantovska: turns out i have misunderstood - you were talking about why not Record but Map, I was thinking why not switch case, since we are using it in other places for the same functionality. changed to Record. thx!

redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
dantovska: Thanks a lot @KrumTy ! 

I tried changing it to getColorBackgroundStyles instead of starting with use, but I'm getting an ESlint error:
`React Hook "useTheme" is called in function "getColorBackgroundStyles" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use".eslintreact-hooks/rules-of-hooks`

Turns out, since in the function I am calling useTheme(), which is a hook, I need to name the function appropriately by react hook rules, or, pass the theme as a param from the outside, which would look like this:

```
export const LoaderBar = styled.div<MapProps>`
  ${({ $color, theme = useTheme() }) => getColorBackgroundStyles({ $color, theme })}

  position: absolute;
  height: 100%;
  width: 100%;
  border-radius: 2px;

  animation: ${loading} 1s ease-in-out infinite;
`
```

Or I guess I could use getColorBackgroundStyles, but disable the ESlint rule?

pawelangelow: What if we access the theme directly like this
https://github.com/redis/RedisInsight/pull/4670/files#diff-37b2790429c11d5d6b8c38b2967de36b1afd7c2f1a51134ce61ef972fc9b2b2fR6

```
gap: ${({ theme }) => theme.core.space.space050};
```

This way, we can avoid using `useTheme`, and `useColorBackgroundStyles` won't need to look like a React hook anymore.

It's also safe to use `${({ theme }) => ...` multiple times across styled components - styled-components caches the styles, so this doesn’t introduce performance issues or unnecessary theme lookups.

pd-redis: I don't mind using the theme in any way possible, but on other hand I don't see what the problem is about with custom hooks? They are simply a funciton with specific name.
And if we happen to change the theme - like for example user changing a setting, it will trigger useTheme.
Just something to test before deciding to rename it

pawelangelow: I'm okay with this being `use*`, but let's not disable eslint for it as it's a hook if it uses `useTheme` inside or `{theme} => ...` if it satisfies what we need (when the user changes the theme setting, all changes to be applied run-time, withot need of a refresh).

pawelangelow (approved):

redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
```diff
@@ -0,0 +1,89 @@
+import { Theme, theme } from '@redis-ui/styles'
+import { ReactNode } from 'react'
+import styled, { css, keyframes } from 'styled-components'
+
+export type EuiColorNames =
```
valkirilov, line 5: Just a question: Why do we want to keep this prefixed with the EUI dependency we're currently removing?

pd-redis: Because we had 100s of components using these color names, and this is just a reminder where they come from.
Hopefully some day, some designer will come up with "redis" color names and we can change them.
Or we can just drop the Eui part

dantovska: The names of the value options are the ones defined in EUI, we are just giving them the closest values coming from Redis UI styles. TBH I don't know if we need to have the EUI in the name, I'm following an already set pattern. I'll mention this in some discussion when I'm working on the refactoring, which I plan to do regarding the coloring functionality of several components.

redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
dantovska: Will bring this up again when refactoring takes place. Tnx!

PR #4670, opened 2025-07-08
RI-7191 Selector tabs component

valkirilov:
# Description

Created base wrapper for the "Create Index" step part of the new Vector Search
    
* Integrate the tab component from Redis UI to serve as a base wrapper
* Created a minimal implementation for the "Use preset index" scenario
* Prepared a minimal setup for the "Build new index" scenario, but disabled for now

# How to test it

This is a base component that will be included in the new wizard for the Vector Search, but it's currently a standalone component. So, you won't be able to find it integrated anywhere in the flow so far.

## Preview

image


redisinsight/ui/src/components/new-index/CreateIndexStep/BuildNewIndex/index.ts
```diff
@@ -0,0 +1 @@
+export * as BuildNewIndex from './TabBarTrigger'
```
pd-redis: Why another barrel file for single export????

redisinsight/ui/src/components/new-index/CreateIndexStep/CreateIndexStepWrapper.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+const initSUT = () => render(<CreateIndexStepWrapper />)
+
+describe('CreateIndexStepWrapper', () => {
+  it('should render', () => {
+    const { container } = initSUT()
+
+    expect(container).toBeTruthy()
+
+    // Check if the tabs are rendered
+    const buildNewIndexTabTrigger = screen.getByText('Build new index')
```
pd-redis, line 14: Won't this fail if the text is not in the document?
If so, then the next expectations are redundant

redisinsight/ui/src/components/new-index/CreateIndexStep/BuildNewIndex/index.ts
valkirilov: We're starting with a simpler version of this feature that supports only the "Use preset index" option ("Build new index" is marked as coming soon). I intended to set up the structure in a way that would make it easier to extend in the future.

That's why I replicated the folder structure here for the "Build New Index" scenario - it should make it more straightforward to plug in the additional functionality later when we decide to implement it.

redisinsight/ui/src/components/new-index/CreateIndexStep/CreateIndexStepWrapper.spec.tsx
valkirilov: With this test, I want to check whether the component is rendered as it should, meaning:
- the tabs are rendered ("Build new index" and "Use preset index" are rendered in the document)
- the default tab content is rendered (although it is just a temporary text for now, it will change soon)

So, yes, I want to check them all and cause the test to fail even if one of them is missing?

redisinsight/ui/src/components/new-index/CreateIndexStep/BuildNewIndex/TabBarTrigger.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Badge } from '@redis-ui/components'
+
+const StyledTabBarTrigger = styled.div`
```
KrumTy: I think it's better to keep the styled components inside a xyz.styles.ts file

redisinsight/ui/src/components/new-index/CreateIndexStep/CreateIndexStepWrapper.tsx
```diff
@@ -0,0 +1,43 @@
...
+    VectorIndexTab.UsePresetIndex,
+  )
+
+  const handleTabChange = (newValue: string) => {
+    setActiveTab(newValue as VectorIndexTab)
+  }
+
+  return (
+    <Tabs
+      defaultValue={activeTab}
```
KrumTy:
```suggestion
      defaultValue={activeTab}
      value={activeTab}
```
I think you should also pass y̶o̶u̶r̶ n̶a̶m̶e̶ value

redisinsight/ui/src/components/new-index/CreateIndexStep/BuildNewIndex/TabBarTrigger.tsx
valkirilov: Thanks for letting me know, I'll try to keep them separate from now on.

I just pushed the necessary changes 👍

redisinsight/ui/src/components/new-index/CreateIndexStep/CreateIndexStepWrapper.tsx
valkirilov: Nice catch, my bad. Actually, we should add `value` and remove `defaultValue` if we want to be a truly controlled component and not a mixture of the two approaches. 

For now, let's keep it controlled, since it will be easier to maintain its state for the wizard that will wrap the flow, changes applied and pushed.

```
// uncontrolled
<Tabs tabs={tabs} defaultValue={tabs[0].value} />;

// controlled
const [selectedTab, setSelectedTab] = useState(tabs[0].value);

<Tabs tabs={tabs} value={selectedTab} onChange={setSelectedTab} />;
```

Reference: https://redislabsdev.github.io/redis-ui/?path=/docs/components-tabs--docs#usage

KrumTy (approved):

redisinsight/ui/src/components/new-index/CreateIndexStep/CreateIndexStepWrapper.tsx
```diff
@@ -0,0 +1,30 @@
...
+  },
+]
+
+export const CreateIndexStepWrapper = () => (
+  <Tabs
+    tabs={VECTOR_INDEX_TABS}
+    defaultValue={VectorIndexTab.UsePresetIndex}
+    variant="sub"
+  />
+)
```
pawelangelow:
```suggestion
export const CreateIndexStepWrapper = (props) => (
  <Tabs
    tabs={VECTOR_INDEX_TABS}
    defaultValue={VectorIndexTab.UsePresetIndex}
    variant="sub"
    {...props}
  />
)
```
to empower the consumers of it:

```ts
const [selectedTab, setSelectedTab] = useState(tabs[0].value)

<CreateIndexStepWrapper value={selectedTab} onChange={setSelectedTab} />
```

valkirilov: My bad, the change that was responsible for this "prop drill" remained uncommitted 🤦‍♂️ 

Now it's pushed, and you can pass any of the Redis UI tab component props, or leave it to rely on the defaults.


```
// Default behavior with two tabs and "Use preset index" tab selected by default
<CreateIndexStepWrapper />

// Custom default value preselected by default
<CreateIndexStepWrapper defaultValue={VectorIndexTab.UsePresetIndex} />
```

redisinsight/ui/src/components/new-index/CreateIndexStep/CreateIndexStepWrapper.tsx
```diff
@@ -0,0 +1,36 @@
...
+  },
+  {
+    value: VectorIndexTab.UsePresetIndex,
+    label: 'Use preset index',
+    content: <div>TODO: Add content later</div>,
+  },
+]
+
+export const CreateIndexStepWrapper = (props: Partial<TabsProps>) => {
+  const { tabs, defaultValue, ...restProps } = props
```
pawelangelow: nit: I think `rest` is fine instead of `restProps` (react community practice)

redisinsight/ui/src/components/new-index/CreateIndexStep/CreateIndexStepWrapper.spec.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react'
+import { fireEvent, render, screen } from 'uiSrc/utils/test-utils'
+import { CreateIndexStepWrapper } from './CreateIndexStepWrapper'
+
+const initSUT = () => render(<CreateIndexStepWrapper />)
```
pawelangelow: nit: `initSystemUnderTesting`? IMO, more practical is `renderComponent` or `renderStepper` or something like this

redisinsight/ui/src/components/new-index/CreateIndexStep/CreateIndexStepWrapper.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+    const usePresetIndexTabTrigger = screen.getByText('Use preset index')
+
+    expect(buildNewIndexTabTrigger).toBeInTheDocument()
+    expect(usePresetIndexTabTrigger).toBeInTheDocument()
+
+    // Check if the "Use preset index" tab content is selected by default
+    const usePresetIIndexTabContent = screen.queryByText(
+      'TODO: Add content later',
+    )
+    expect(usePresetIIndexTabContent).toBeInTheDocument()
```
pawelangelow: nit: is there a way to check for some html attribute, as this test would require change when the text changes

redisinsight/ui/src/components/new-index/CreateIndexStep/CreateIndexStepWrapper.tsx
valkirilov: I saw it as a pattern in `redis-ui` and thought it was the convention I must follow. 
Anyway, `restProps` is updated to `rest` now ✌️

pawelangelow (approved):

redisinsight/ui/src/components/new-index/CreateIndexStep/CreateIndexStepWrapper.spec.tsx
valkirilov: tl;dr renamed it to `renderComponent`, I agree that this name better reflects what the function does.

I originally used `initSUT`, thinking it was a common pattern for setting up the component under test. I've seen it used in a few projects and believe it was introduced by Kent C. Dodds (the author of React Testing Library). But if it's not a familiar convention in this codebase, I'm happy to stick with something clearer for everyone, and don't intend to force it 🙂

redisinsight/ui/src/components/new-index/CreateIndexStep/CreateIndexStepWrapper.spec.tsx
valkirilov: Sure thing, I've added a `data-testid` to help reliably check for the presence of this element.

Initially, I was using a text-based query, following the default [query rules priority](https://testing-library.com/docs/queries/about/#priority) in React Testing Library, knowing that we'd replace it in a future PR once this component is updated with real content. But you're absolutely right - we'll test the actual contents later when we introduce it, most probably in a separate component.

For now, since we're just verifying that the Tabs component works as expected, using a testid makes perfect sense, and is enough to serve the job. Change applied, thanks for the input 🙌

pd-redis: The last response reads like its an LLM responding 😄

PR #4671, opened 2025-07-08
[RI-7064]Replace EuiPopover with RiPopover

KrumTy:


redisinsight/ui/src/components/base/popover/types.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export type RiPopoverProps = Omit<
+  PopoverProps,
+  | 'open'
+  | 'onClickOutside'
+  | 'autoFocus'
+  | 'content'
+  | 'className'
+  | 'placement'
+  | 'align'
```
KrumTy, line 17: I've changed the props a bit to preserve the previous structure and minimize the changes

valkirilov (approved):

redisinsight/ui/src/components/base/popover/RiPopover.tsx
```diff
@@ -0,0 +1,37 @@
...
+  <Popover
+    {...props}
+    open={isOpen}
+    onClickOutside={closePopover}
+    content={children}
+    // Props passed to the children wrapper:
+    className={panelClassName}
+    maxWidth={maxWidth}
+    style={{
+      padding: panelPaddingSize && panelPaddingSizeMap[panelPaddingSize],
```
pd-redis, line 29: Why inline style and not styled component?

redisinsight/ui/src/utils/test-utils.tsx
```diff
@@ -251,11 +251,18 @@ const waitForRiTooltipHidden = async () => {
   })
 }
 
-const waitForEuiPopoverVisible = async (timeout = 500) => {
+const waitForRiPopoverVisible = async (timeout = 500) => {
   await waitFor(
     () => {
-      const tooltip = document.querySelector('.euiPopover__panel-isOpen')
+      const tooltip = document.querySelector(
+        'div[data-radix-popper-content-wrapper]',
```
pd-redis, line 258: Isn't this exactly the same check as in `waitForRiTooltipVisible` and `waitForRedisUiSelectVisible` ?

KrumTy: looks like yes, internally redis ui has the same wrapper for both
doesn't mean we should use a single helper method though for popovers and tooltips though

redisinsight/ui/src/components/base/popover/RiPopover.tsx
KrumTy: it's a composed component and wrapping it in styled component yields a lot of ts issues
Moreover, these props under the comment, are passed to the wrapper of the content by the Popover internally ([ref](https://github.com/redislabsdev/redis-ui/blob/master/packages/components/src/Popover/Popover.tsx#L21))

dantovska (approved):

redisinsight/ui/src/utils/test-utils.tsx
pd-redis: I think it renders exactly the same thing underneeth, it is a popper instance. But anyway, mo methods, mo problems 🤓

PR #4718, opened 2025-07-11
[RI-7194] Add "Create vector search" wizard

KrumTy:


redisinsight/ui/src/constants/pages.ts
```diff
@@ -49,6 +50,8 @@ export const Pages = {
   sentinelDatabases: `${sentinel}/databases`,
   sentinelDatabasesResult: `${sentinel}/databases-result`,
   browser: (instanceId: string) => `/${instanceId}/${PageNames.browser}`,
+  vectorSearch: (instanceId: string) =>
```
pawelangelow, line 53: Why not just `search`?

pawelangelow:
Change the target branch to `feature/RI-6855/vector-search`

redisinsight/ui/src/pages/vector-search/create-index/styles.ts
```diff
@@ -0,0 +1,34 @@
+import styled from 'styled-components'
+import { FlexGroup, FlexItem } from 'uiSrc/components/base/layout/flex'
+
+export const CreateIndexWrapper = styled(FlexGroup)`
+  width: 95%;
+  background-color: white;
+  margin-top: 20px;
+  margin-bottom: 20px;
```
valkirilov: I think it's better to use the spacing provided by Redis UI ([docs](https://redislabsdev.github.io/redis-ui/?path=/docs/theme--docs)) instead of hardcoding the values.

```
${({ theme }) => theme.core.space.space050};
```

However, I don't know what the convention is that you usually follow - using the design system as a main source of truth or you try to do a pixel-perfect results, based on the Figma designs.

redisinsight/ui/src/pages/vector-search/create-index/styles.ts
```diff
@@ -0,0 +1,34 @@
...
+  margin-left: auto;
+  margin-right: auto;
+`
+
+export const CreateIndexHeader = styled(FlexItem)`
+  justify-content: space-between;
+  padding: 24px;
+  border: 1px solid #b9c2c6;
```
valkirilov: I think it's better to use the theme colors provided by Redis UI ([docs](https://redislabsdev.github.io/redis-ui/?path=/docs/theme--docs)), instead of using hardcoded colors. You can see that `#b9c2c6` is already there under `variants.secondary-ghost.disabled.borderColor`.

redisinsight/ui/src/pages/vector-search/create-index/styles.ts
```diff
@@ -0,0 +1,34 @@
+import styled from 'styled-components'
+import { FlexGroup, FlexItem } from 'uiSrc/components/base/layout/flex'
+
+export const CreateIndexWrapper = styled(FlexGroup)`
```
valkirilov, line 5: I've noticed a few weird design issues:
1. The border radius is not properly applied to the parent container, and you can see how it overflows with the background color, on light theme.

image

image

2. For some reason, the theming is not working properly when using dark theme
image

redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.tsx
```diff
@@ -0,0 +1,98 @@
...
+type VectorSearchCreateIndexProps = {
+  initialStep?: number
+}
+
+export const VectorSearchCreateIndex = ({
+  initialStep = 1,
+}: VectorSearchCreateIndexProps) => {
+  const { instanceId } = useParams<{ instanceId: string }>()
+  const [step, setStep] = useState(initialStep)
+  const [createSearchIndexParameters, setCreateSearchIndexParameters] =
```
valkirilov, line 37: Tip: It's ok for now, since it's the first PR in a series, so I'll leave a note only to think about it. 

I like the idea you went for a state variable to control this (instead of Redux), but we'll need to prop drill it to every component, in order to access some useful information. Instead of deep prop drilling, we may use React Context, and let it be a "central state storage" for the family of components related to the Vector Search setup.

But we may change this in the next PRs, once we have the other components in place and more vision on the data model itself.

redisinsight/ui/src/constants/pages.ts
KrumTy: because that's the name of the epic and the target branch
I thought that was the name of the feature

redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.tsx
KrumTy: yes, context will be the way to go if deep sharing of state is needed
this is just a skeleton/wrapper component w/o actual functionality after all

redisinsight/ui/.eslintrc.js
```diff
@@ -93,6 +93,7 @@ module.exports = {
     'sonarjs/no-duplicate-string': 'off',
     'sonarjs/cognitive-complexity': [1, 20],
     'sonarjs/no-identical-functions': [0, 5],
+    'sonarjs/no-nested-template-literals': 'off',
```
KrumTy, line 96: turned off so we can we can add theme related css inside styled components without an extra wrapper function

valkirilov (approved):

PR #4720, opened 2025-07-14
RDI Tab and check

KIvanow:
Added a simple util that checks the URL of the page and if it matches, returns true. Used to enable a third tab in the copilot chat window. Currently limited to the page for initial set up after connecting to an RDI instance.

[https://www.loom.com/share/720faf8aa7424f40b351b2a45b6f59cd](https://www.loom.com/share/720faf8aa7424f40b351b2a45b6f59cd)

### DIFF
--- a/redisinsight/ui/src/components/side-panels/panels/ai-assistant/components/chats-wrapper/ChatsWrapper.tsx
+++ b/redisinsight/ui/src/components/side-panels/panels/ai-assistant/components/chats-wrapper/ChatsWrapper.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react'
+import React, { useEffect, useState } from 'react'
 
 import { EuiTab, EuiTabs } from '@elastic/eui'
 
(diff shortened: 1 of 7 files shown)


valkirilov (approved):

redisinsight/ui/src/components/side-panels/panels/ai-assistant/components/chats-wrapper/ChatsWrapper.tsx
```diff
@@ -92,6 +109,15 @@ const ChatsWrapper = () => {
                 My Data
               </EuiTab>
             )}
+            {isOnRdiPipelineConfig && (
```
valkirilov, line 112: If we're not sure whether this feature will make it, we could also consider hiding it behind a feature flag as well? This way, we'll able to toggle the visibility of the new tab for testing purposes only.

KIvanow: It should've been in a forked repo, unclear how the PR ended up here tbh

PR #4723, opened 2025-07-16
RI-6570 Verify read operations for all key types in the browsers module

valkirilov:
# Description

Add new E2E tests to check the read functionality for different key types in the **Browsers** module. These tests make sure that when you click on a key, all its details show up correctly in the Details Panel (name, value(s), TTL, etc.).

### How the tests work
* First, prepare the data for the test using the Redis API (it's easier and faster and also, the UI flow for creating keys is already covered in another test suite)
* Then we open the **Details Panel** by clicking on the key we just added (in the table in the UI)
* Finally, we check that all the key information is displayed correctly, like "**Key name**", "**Value(s)**", "**TTL**", etc.

image

## Code Changes

* Extend the API to support adding various key types, enabling us to easily assemble test data
* Extend browser page locators to provide helpers for checking the state of the keys details drawer
* Add e2e tests to verify that the information related to the various key types is properly displayed in the details drawer

# How to run the tests

You can always refer to the README, but simply running the following command should do the trick for you

```
# From the root directory
yarn dev:api

yarn dev:ui

yarn test:chromium:local-web browser/keys-read 

# The, check the detailed report an all the video recordings
yarn playwright show-report
``` 

image

### DIFF
--- a/tests/playwright/helpers/api/api-keys.ts
+++ b/tests/playwright/helpers/api/api-keys.ts
@@ -164,6 +164,32 @@ export class APIKeyRequests {
         }
     }
 
+    async addJsonKeyApi(
+        keyParameters: { keyName: string; value: any; expire?: number },
+        databaseParameters: AddNewDatabaseParameters,
+    ): Promise<void> {
+        const databaseId = await this.databaseAPIRequests.getDatabaseIdByName(
+            databaseParameters.databaseName,
+        )
+        const requestBody: any = {
+            keyName: Buffer.from(keyParameters.keyName, 'utf-8'),
+            data: JSON.stringify(keyParameters.value),
+        }
+
+        if (keyParameters.expire) {
+            requestBody.expire = keyParameters.expire
+        }
+
+        const response = await this.apiClient.post(
+            `/databases/${databaseId}/rejson-rl?encoding=buffer`,
+            requestBody,
+        )
+
+        if (response.status !== 201) {
+            throw new Error('The creation of new JSON key request failed')
+        }
+    }
+
     async searchKeyByNameApi(
         keyName: string,
         databaseName: string,
--- a/tests/playwright/helpers/api/api-keys.ts
+++ b/tests/playwright/helpers/api/api-keys.ts
@@ -7,6 +7,7 @@ import {
     SetKeyParameters,
     StreamKeyParameters,
 } from '../../types'
+import { stringToBuffer } from '../utils'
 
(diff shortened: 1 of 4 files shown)


tests/playwright/helpers/api/api-keys.ts
```diff
@@ -84,6 +135,62 @@ export class APIKeyRequests {
...
+        },
+        databaseParameters: AddNewDatabaseParameters,
+    ): Promise<void> {
+        const databaseId = await this.databaseAPIRequests.getDatabaseIdByName(
+            databaseParameters.databaseName,
+        )
+        const requestBody = {
+            keyName: Buffer.from(keyParameters.keyName, 'utf-8'),
+            members: keyParameters.members.map((member) => ({
+                name: Buffer.from(member.name, 'utf-8'),
```
pd-redis: `Buffer.from(member.name, 'utf-8')` can be probably abstracted as a helper function

tests/playwright/pageObjects/browser-page.ts
```diff
@@ -1282,4 +1286,320 @@ export class BrowserPage extends BasePage {
...
+            return false
+        } catch (error) {
+            return false
+        }
+    }
+
+    async isKeyDetailsClosed(): Promise<boolean> {
+        try {
+            // Give a small moment for UI transitions to complete
+            await this.page.waitForTimeout(100)
```
pd-redis: this is flaky. Better wait for an element to show

tests/playwright/pageObjects/browser-page.ts
```diff
@@ -1282,4 +1286,320 @@ export class BrowserPage extends BasePage {
...
+            }
+
+            const actualValue = await valueLocator.textContent()
+            return actualValue?.includes(fieldValue) || false
+        } catch {
+            return false
+        }
+    }
+
+    async getKeyTTL(): Promise<string | null> {
```
pd-redis: there is function declared in constructor with very similar name.

valkirilov: Nice catch, refactored.

tests/playwright/helpers/api/api-keys.ts
valkirilov: Done 👍

tests/playwright/pageObjects/browser-page.ts
valkirilov: Totally agree, fixed.

pawelangelow (approved):

PR #4725, opened 2025-07-16
Decouple ui/* from api/* via openapi types

KrumTy:
## Issue
Currently we have a lot of typescript issues on the client side. Many of them are due to references to the nestjs api and/or mismatch between api contracts.

## What this PR does
It addresses decoupling the `ui/*` from `api/*` be leveraging auto generated types based on openapi (swagger).
It shrinks the tsc errors by about 50% (from ~2500 to ~1300) and sets the path for further ts fixes.

## Review guide
This PR appears large at first but that's because of the large amount of autogenerated types. Do not review the `ui/src/api-client` folder

## Notes
- When an api controller changes (param/dtos/etc), `generate-client` script needs to be rerun in order to generate the new types for the client. Not sure if this can be automated.
- There's an option to split the autogenerated types into multiple files (one per type) but searching through the codebase with so many overlapping file names becomes too hard, so I kept them in one (`api/clinet/api.ts`)


redisinsight/api/package.json
```diff
@@ -172,7 +174,6 @@
     ],
     "moduleNameMapper": {
       "src/(.*)": "<rootDir>/$1",
-      "apiSrc/(.*)": "<rootDir>/$1",
```
pd-redis, line 175: was this unused?

valkirilov:
Great job on this, using an OpenAPI contract to keep services in sync is a solid and widely used approach, especially when different tech stacks are involved or when the code lives in separate repos. In those situations, generating types from Swagger docs makes it easier to keep services aligned and follow the correct communication protocol.

That said, in our case, we're working in a monorepo, and both the backend and frontend use TypeScript. I don't want to sound too negative, but adding a (semi-automatic) code generation tool here feels like extra complexity and might hide the real issue that we don't have a single source of truth for our types.

Instead of adding another layer, I think it would be more effective to focus on fixing the type mismatches directly. We could do this by adjusting how the frontend uses the types, or better yet, by introducing a shared types package (or something similar) so both sides rely on the same definitions.

Another thing to consider is that OpenAPI doesn't always match perfectly with the actual TypeScript code, so I think it's better to stick to the code interfaces as the main source of truth.

That said, if there are plans to move away from the monorepo and split the services, then generating types from Swagger would absolutely make sense, and it will probably be the best approach. I'm not against it overall, just think we should focus on solving the core issue first. 

So, I'll be happy to chat more about this so we can agree on which is the best way to reduce these TypeScript errors and resolve the type mismatches between the BE and the FE.

pd-redis:
> Great job on this, using an OpenAPI contract to keep services in sync is a solid and widely used approach, especially when different tech stacks are involved or when the code lives in separate repos. In those situations, generating types from Swagger docs makes it easier to keep services aligned and follow the correct communication protocol.
> 
> That said, in our case, we're working in a monorepo, and both the backend and frontend use TypeScript. I don't want to sound too negative, but adding a (semi-automatic) code generation tool here feels like extra complexity and might hide the real issue that we don't have a single source of truth for our types.
> 
> Instead of adding another layer, I think it would be more effective to focus on fixing the type mismatches directly. We could do this by adjusting how the frontend uses the types, or better yet, by introducing a shared types package (or something similar) so both sides rely on the same definitions.
> 
> Another thing to consider is that OpenAPI doesn't always match perfectly with the actual TypeScript code, so I think it's better to stick to the code interfaces as the main source of truth.
> 
> That said, if there are plans to move away from the monorepo and split the services, then generating types from Swagger would absolutely make sense, and it will probably be the best approach. I'm not against it overall, just think we should focus on solving the core issue first.
> 
> So, I'll be happy to chat more about this so we can agree on which is the best way to reduce these TypeScript errors and resolve the type mismatches between the BE and the FE.

I don't think BE types should be used at all on FE, besides in the apiService. 
Many type problems are caused because `RedisString` can be either string or `RedisStringBuffer`, and its being used as a string in most places, but in a few it is converted to string or to buffer. However, it is not actually a Buffer, but a data structure with `type` and `data` properties - so in this example, I don't think it should be typed as Buffer (it is send like one from BE, but serialized to JSON as plain object), but as:
```
{
type: string
data: number[]
}
```

Since we're passing via the API, using the same types is a lie for anything that is not supported by JSON and converting from json to buffer for example is a waste since we don't use it as buffer anywhere

redisinsight/api/package.json
KrumTy: I don't think so

pawelangelow:
What are we going to do here?

KIvanow:
> What are we going to do here?

Were the comments from @pd-redis and @valkirilov  adressed @KrumTy ?

Before all of the conflicts, were the tests (despite being flaky) passing?

Is anyone against these changes (@ArtemHoruzhenko and @dantovska as only ones not tagged so far)?

redisinsight/api/package.json
ArtemHoruzhenko: I think we didn't use apiSrc from api folder. so it should be safe to remove

redisinsight/api/src/common/decorators/redis-string-schema.decorator.ts
```diff
@@ -0,0 +1,32 @@
...
+        data: {
+          type: 'array',
+          items: { type: 'number' },
+          example: [61, 101, 49],
+        },
+      },
+      required: ['type', 'data'],
+    },
+  ],
+};
```
ArtemHoruzhenko, line 20: Nice! I assume before we had "string" in swagger, right?

redisinsight/api/src/common/decorators/redis-string-schema.decorator.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const ApiRedisString = (
+  description: string = undefined,
+  isArray = false,
+  required = true,
+) =>
```
ArtemHoruzhenko, line 26: I feel like it will be better to join `isArray` and `required` and send as a second argument `{isArray: string, required: boolean}`. It will be more flexible/scalable approach.

redisinsight/api/src/modules/browser/hash/dto/delete.fields-from-hash.dto.ts
```diff
@@ -1,15 +1,14 @@
...
+} from 'src/common/decorators';
 import { RedisString } from 'src/common/constants';
 
 export class DeleteFieldsFromHashDto extends KeyDto {
-  @ApiProperty({
-    description: 'Hash fields',
-    type: String,
-    isArray: true,
-  })
+  @ApiRedisString('Hash fields', true)
```
ArtemHoruzhenko, line 11: and here it will be clear what we are doing: e.g. `@ApiRedisString('Hash fields', { isArray: true })`

redisinsight/api/src/modules/cloud/user/models/cloud-user-account.ts
```diff
@@ -1,16 +1,33 @@
...
+    type: String,
+  })
   name: string;
 
   @Expose({ groups: [TransformGroup.Secure] })
+  @ApiProperty({
+    description: 'Cloud API key',
+    type: String,
+  })
   capiKey?: string; // api_access_key
```
ArtemHoruzhenko, line 25: fields under 'secure' group are not sent to UI not sure we must include them into swagger

redisinsight/desktop/src/lib/cloud/cloud-oauth.handlers.ts
```diff
@@ -5,22 +5,36 @@ import { UrlWithParsedQuery } from 'url'
...
+} from 'uiSrc/api-client'
+
 import { createAuthStrategy } from '../auth/auth.factory'
 import { getWindows } from '../window/browserWindow'
 
 const authStrategy = createAuthStrategy()
+// ref: /api/src/common/constants/user.ts
+const DEFAULT_USER_ID = '1'
+const DEFAULT_SESSION_ID = '1'
+// ref: /api/src/modules/cloud/auth/exceptions/cloud-oauth.unexpected-error.exception.ts
```
ArtemHoruzhenko, line 21: what is ref here? do we have circular deps here?

redisinsight/ui/src/api-client/common.ts
```diff
@@ -0,0 +1,150 @@
+/* tslint:disable */
+/* eslint-disable */
+/**
+ * Redis Insight Backend API
+ * Redis Insight Backend API
```
ArtemHoruzhenko, line 5: twice for those who didn't get 😄

redisinsight/ui/src/api-client/common.ts
```diff
@@ -0,0 +1,150 @@
...
+
+import type { Configuration } from "./configuration";
+import type { RequestArgs } from "./base";
+import type { AxiosInstance, AxiosResponse } from 'axios';
+import { RequiredError } from "./base";
+
+/**
+ *
+ * @export
+ */
```
ArtemHoruzhenko, line 24: why do we need this?

redisinsight/ui/src/api-client/docs/AIApi.md
ArtemHoruzhenko, line 1: I'm not a big fan of having millions of .md files in our source code. Why we need it?

redisinsight/api/src/common/decorators/redis-string-schema.decorator.ts
KrumTy: It was mostly string yes

redisinsight/api/src/common/decorators/redis-string-schema.decorator.ts
KrumTy: yeah, I figured I'll get that comment
it was just 2 params until the very end when a 3rd was needed for a few service methods

redisinsight/desktop/src/lib/cloud/cloud-oauth.handlers.ts
KrumTy: this is where the class/constant was ported over since it can't be shipped via swagger
(rare places required this)

redisinsight/ui/src/api-client/common.ts
KrumTy: those are all autogenerated helpers (everything inside api-client/)
common.ts is imported in api.ts

redisinsight/ui/src/api-client/docs/AIApi.md
KrumTy: they come by default via the auto generator cli. I suppose there will be a config to omit markdown

PR #4739, opened 2025-07-21
RI-7212: replace EuiFormFieldset with FormFieldset

KrumTy:


valkirilov (dismissed):
Tip: It will be awesome if you provide screenshots with visual comparison between the "old" and "new" version of the components, as part of the description of the pull request.

| Old | New |
|-|-|
image|image

KrumTy:
@valkirilov I do that when applicable but in this case there's no visual change really; just a fieldset wrapper. In the screenshots you've shown it's actually hidden too.

redisinsight/ui/src/components/base/forms/fieldset/FormFieldset.styles.ts
```diff
@@ -0,0 +1,25 @@
...
+`
+
+export interface StyledLegendProps extends HTMLAttributes<HTMLLegendElement> {
+  display?: 'visible' | 'hidden'
+}
+
+export const StyledLegend = styled.legend<StyledLegendProps>`
+  ${({ theme, display }: { theme: Theme } & StyledLegendProps) => css`
+    margin-bottom: ${theme.core.space.space100};
+    display: ${display === 'hidden' ? 'none' : 'block'};
```
pawelangelow: I'm wondering if we need to change the value from `hidden` to `none` as `hidden` looks like an invalid choice for `display`, and then we won't have this check. I see only 2 usages - `CreateRedisearchIndex.tsx` and `AddKeyCommonFields.tsx`

pd-redis: This may conflict with any class passed that alters display - flex, inline-flex whatever

KrumTy: I've moved the check to component-render level; let me know if you have another idea

pd-redis (approved):
lgtm

PR #4740, opened 2025-07-21
electron upgraded from 33.* to 36.*

KIvanow:
We have multiple different conflicts when trying to update to the latest 37.*, so I ended up with 36, which is still supported and still newer that what we are using. There have been changes to the electron builder API/config that were needed. One of these is the way mac is being notarized [https://www.electron.build/app-builder-lib.interface.macconfiguration#notarize](https://www.electron.build/app-builder-lib.interface.macconfiguration#notarize)

### DIFF
--- a/redisinsight/api/yarn.lock
+++ b/redisinsight/api/yarn.lock
@@ -2972,7 +2972,7 @@ buffer@^6.0.3:
     base64-js "^1.3.1"
     ieee754 "^1.2.1"
 
-busboy@^1.6.0:
+busboy@^1.0.0, busboy@^1.6.0:
   version "1.6.0"
   resolved "https://registry.yarnpkg.com/busboy/-/busboy-1.6.0.tgz#966ea36a9502e43cdb9146962523b92f531f6893"
   integrity sha512-8SFQbg/0hQ9xy3UNTB0YEnsNBbWfhf7RtnzpL7TkBiTBRfrQ9Fxcnz7VJsleJpyp6rVLvXiuORqjlHi5q+PYuA==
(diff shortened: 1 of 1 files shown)


electron-builder.json
```diff
@@ -24,9 +27,7 @@
         "arch": ["x64", "arm64"]
       }
     ],
-    "notarize": {
-      "teamId": "UUK47G4BAZ"
-    },
+    "notarize": true,
```
ArtemHoruzhenko, line 30: not sure it will work for mas when removed

redisinsight/api/yarn.lock
ArtemHoruzhenko, line 1: why we have changes in api lock file? is it out of scope for this PR?

electron-builder.json
KIvanow: It's not removed. It is changed to be a boolean field https://www.electron.build/app-builder-lib.interface.macconfiguration#notarize

redisinsight/api/yarn.lock
KIvanow: It was removed by mistake. I've removed all node modules and yarn.lock files to download and regenerate everything after the updates and have update the api's yarn.lock by mistake. Added another [commit](https://github.com/redis/RedisInsight/pull/4740/commits/0fac2f9e13d2ad5b0dafa9f54da93b31f9dc71b2) to revert the change and tested it with a [build](https://github.com/redis/RedisInsight/actions/runs/16436678645) after. Continues working normally.

ArtemHoruzhenko (approved):

KrumTy (approved):

PR #4742, opened 2025-07-21
RI-7226: fix RiTooltip when content is empty

KrumTy:
## Description

| Before  | After |
| ------------- | ------------- |
| Screenshot 2025-07-21 at 17 31 57  | Screenshot 2025-07-21 at 17 31 48 |

### DIFF
--- a/redisinsight/ui/src/components/base/tooltip/RITooltip.tsx
+++ b/redisinsight/ui/src/components/base/tooltip/RITooltip.tsx
@@ -23,7 +23,7 @@ export const RiTooltip = ({
   <TooltipProvider>
     <Tooltip
       {...props}
-      content={<HoverContent title={title} content={content} />}
+      content={content && <HoverContent title={title} content={content} />}
       placement={position}
       openDelayDuration={delay}
     >


redisinsight/ui/src/components/base/tooltip/RiTooltip.spec.tsx
```diff
@@ -0,0 +1,189 @@
...
+    render(
+      <RiTooltip {...defaultProps} title="Test Title" content="Test content" />,
+    )
+
+    await act(async () => {
+      fireEvent.focus(screen.getByTestId('tooltip-trigger'))
+    })
+    await waitForRiTooltipVisible()
+
+    expect(screen.getAllByText('Test Title')[0]).toBeInTheDocument()
```
valkirilov, line 49: Question: Why not the simpler query function `getByText`, in this case, when we know there will be only one text element rendered? It looks cleaner to me, and I wonder whether you know any perks of using the `getAllByText` variant?

```suggestion
    expect(screen.getByText('Test Title')).toBeInTheDocument()
```

KrumTy: in many cases (such as this) there's another hidden element with redis ui; probably for accessibility but I'm not sure

KIvanow (approved):

redisinsight/ui/src/components/base/tooltip/RiTooltip.spec.tsx
valkirilov: Good to know, thanks! I might face a similar issue at some point and will remember your note (hopefully) 😀

PR #4746, opened 2025-07-22
RI-7205 replace eui breadcrumbs

pd-redis:
Breadcrumbs are not used anywhere in the pro


KIvanow:
@pd-redis please use fe/feature prefix, so the BE tests don't have to run. Just something to keep in mind for the future, no need to change the branch name
The FE tests are failing though and should be checked

PR #4747, opened 2025-07-22
RI-7226: fix RiTooltip when title/content is empty

KrumTy:
## Description

If only title or only content is provided, the tooltip should render

### DIFF
--- a/redisinsight/ui/src/components/base/tooltip/RITooltip.tsx
+++ b/redisinsight/ui/src/components/base/tooltip/RITooltip.tsx
@@ -23,7 +23,9 @@ export const RiTooltip = ({
   <TooltipProvider>
     <Tooltip
       {...props}
-      content={content && <HoverContent title={title} content={content} />}
+      content={
+        (content || title) && <HoverContent title={title} content={content} />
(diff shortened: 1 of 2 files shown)


redisinsight/ui/src/components/base/tooltip/RITooltip.tsx
```diff
@@ -23,7 +23,9 @@ export const RiTooltip = ({
   <TooltipProvider>
     <Tooltip
       {...props}
-      content={content && <HoverContent title={title} content={content} />}
+      content={
+        (content || title) && <HoverContent title={title} content={content} />
```
pawelangelow, line 27: Looking at the test - `content && title`?

KrumTy: there's a test that cover this already; I 've renamed the current test

pawelangelow: yup, that changes my expectations :D

pawelangelow (approved):
lgtm

PR #4748, opened 2025-07-22
RI-7197 Add API endpoint for deleting Redis indexes

valkirilov:
# Description

Added a simple API endpoint to allow deleting existing Redis indexes.

_Note: We're gonna need it in the new Vector Search interface later._

image

# How to use it

## Request

```
DELETE {{API_URL}}/{{DB_INTANCE}}/redisearch

{
  "index": "idx:smpl_bicycle"
}
```

### When the index is deleted successfully

```
HTTP Status Code: 204 No Content
```

### When there is no such index to be deleted

```
{
  "message": "idx:smpl_bicycle: no such index",
  "error": "Not Found",
  "statusCode": 404
}
```

### When something else fails
```
{
  "message": "Lorem ipsum dolor sit amet...",
  "error": "Internal Server Error",
  "statusCode": 500
}
```

### DIFF
--- a/redisinsight/api/test/api/redisearch/POST-databases-id-redisearch-info.test.ts
+++ b/redisinsight/api/test/api/redisearch/POST-databases-id-redisearch-info.test.ts
@@ -146,11 +146,11 @@ describe('POST /databases/:id/redisearch/info', () => {
         data: {
           index: 'Invalid index',
         },
-        statusCode: 500,
+        statusCode: 404,
         responseBody: {
+          statusCode: 404,
           message: INVALID_INDEX_ERROR_MESSAGE_V1,
-          error: 'Internal Server Error',
-          statusCode: 500,
+          error: 'Not Found',
         },
       },
     ].forEach(mainCheckFn);
@@ -177,11 +177,11 @@ describe('POST /databases/:id/redisearch/info', () => {
         data: {
           index: 'Invalid index',
         },
-        statusCode: 500,
+        statusCode: 404,
         responseBody: {
+          statusCode: 404,
           message: INVALID_INDEX_ERROR_MESSAGE_V2,
-          error: 'Internal Server Error',
-          statusCode: 500,
+          error: 'Not Found',
         },
       },
     ].forEach(mainCheckFn);
@@ -204,11 +204,11 @@ describe('POST /databases/:id/redisearch/info', () => {
         data: {
           index: 'Invalid index',
         },
-        statusCode: 500,
+        statusCode: 404,
         responseBody: {
-          message: INVALID_INDEX_ERROR_MESSAGE_V2,
-          error: 'Internal Server Error',
-          statusCode: 500,
+          statusCode: 404,
+          message: 'Invalid index: no such index',
+          error: 'Not Found',
         },
       },
     ].forEach(mainCheckFn);


redisinsight/api/src/modules/browser/redisearch/redisearch.service.ts
```diff
@@ -269,6 +271,34 @@ export class RedisearchService {
     }
   }
 
+  public async deleteIndex(
+    clientMetadata: ClientMetadata,
+    dto: IndexDeleteRequestBodyDto,
+  ): Promise<void> {
+    const { index } = dto;
+    this.logger.debug(`Deleting redisearch index ${index}.`, clientMetadata);
```
ArtemHoruzhenko: Let's not add index into log message string

redisinsight/api/src/modules/browser/redisearch/redisearch.service.ts
```diff
@@ -269,6 +271,34 @@ export class RedisearchService {
...
+
+      await client.sendCommand(['FT.DROPINDEX', index]);
+
+      this.logger.debug(
+        `Successfully deleted redisearch index ${index}.`,
+        clientMetadata,
+      );
+    } catch (error) {
+      this.logger.error(
+        `Failed to delete redisearch index ${index}.`,
```
ArtemHoruzhenko: the same

redisinsight/api/src/modules/browser/redisearch/redisearch.service.ts
```diff
@@ -269,6 +271,34 @@ export class RedisearchService {
...
+    this.logger.debug(`Deleting redisearch index ${index}.`, clientMetadata);
+
+    try {
+      const client: RedisClient =
+        await this.databaseClientFactory.getOrCreateClient(clientMetadata);
+
+      await client.sendCommand(['FT.DROPINDEX', index]);
+
+      this.logger.debug(
+        `Successfully deleted redisearch index ${index}.`,
```
ArtemHoruzhenko: and here

redisinsight/api/src/modules/browser/redisearch/redisearch.service.ts
```diff
@@ -269,6 +271,34 @@ export class RedisearchService {
     }
   }
 
+  public async deleteIndex(
```
ArtemHoruzhenko, line 274: it seems like there is no appropriate unit test added for this method

ArtemHoruzhenko (commented):
looks good
there are few minor things and seems like tests are failing.

redisinsight/api/src/modules/browser/redisearch/redisearch.service.ts
valkirilov: tl;dr - removed, thanks for your input.

As discussed over chat, my intention was to provide some helpful clues in case we have to debug issues related to that feature, and with the intention that there is some chance that the index name doesn't categorize fully as "user data". I tried to move it as an n-th parameter, not part of the message itself, but there is another issue with the fact we work with buffers and the index name is not always a string. 

So, as a conclusion, in order to avoid any potential issues, I have removed it completely from the logs.

redisinsight/api/src/modules/browser/redisearch/redisearch.service.ts
valkirilov: Thanks for the reminder, I went for the integration tests and forgot about the unit tests.

Now, we have coverage for this method as well.

ArtemHoruzhenko (approved):
lgtm

PR #4752, opened 2025-07-23
RI-7187 show deployed pipeline by default

ArtemHoruzhenko:
# Description

Show deployed pipeline from RDI server by default otherwise show a popup as now

## When pipeline exists:
Screenshot 2025-07-23 at 17 05 22
 
## When pipeline doesn't exist:
Screenshot 2025-07-23 at 17 05 33

### DIFF
--- a/redisinsight/ui/src/pages/rdi/pipeline-management/PipelineManagementPage.tsx
+++ b/redisinsight/ui/src/pages/rdi/pipeline-management/PipelineManagementPage.tsx
@@ -5,6 +5,7 @@ import { useDispatch, useSelector } from 'react-redux'
 import { IRoute, PageNames, Pages } from 'uiSrc/constants'
 import { connectedInstanceSelector } from 'uiSrc/slices/rdi/instances'
 import {
+  fetchRdiPipeline,
   fetchRdiPipelineJobFunctions,
   fetchRdiPipelineSchema,
 } from 'uiSrc/slices/rdi/pipeline'
@@ -43,6 +44,7 @@ const PipelineManagementPage = ({ routes = [] }: Props) => {
   setTitle(`${rdiInstanceName} - Pipeline Management`)
 
   useEffect(() => {
+    dispatch(fetchRdiPipeline(rdiInstanceId))
     dispatch(fetchRdiPipelineSchema(rdiInstanceId))
     dispatch(fetchRdiPipelineJobFunctions(rdiInstanceId))
   }, [])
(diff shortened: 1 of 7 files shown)


redisinsight/ui/src/pages/rdi/pipeline-management/components/source-pipeline-dialog/SourcePipelineModal.spec.tsx
```diff
@@ -111,4 +123,20 @@ describe('SourcePipelineDialog', () => {
       },
     })
   })
+
+  it('should not show dialog when there is deployed pipeline on a server', () => {
+    const sendEventTelemetryMock = jest.fn()
+    ;(sendEventTelemetry as jest.Mock).mockImplementation(
```
pawelangelow, line 129: Nit: due to our eslint config this looks very ugly 😄 So if you want to avoid it:
```
const mockedSendEventTelemetry = sendEventTelemetry as jest.Mock
...
mockedSendEventTelemetry.mockImplementation(...)
```

redisinsight/ui/src/pages/rdi/pipeline-management/PipelineManagementPage.tsx
```diff
@@ -43,6 +44,7 @@ const PipelineManagementPage = ({ routes = [] }: Props) => {
   setTitle(`${rdiInstanceName} - Pipeline Management`)
 
   useEffect(() => {
+    dispatch(fetchRdiPipeline(rdiInstanceId))
```
pawelangelow, line 47: I'm not a fan of this - not saying changes are required, but it does break the single responsibility principle. This function is essentially just a trigger to populate Redux with the right data, which is then used by `SourcePipelineModal`. So now the logic is split across two places.

We’re not tackling the broader Redux-related issues in this PR, but what if we move this logic into the modal as well? At least that way, it’s all in one place.

What do you think?

ArtemHoruzhenko: I understand your point but disagree with it.
pipeline redux state used across many components on this page for me it is not logical to put it into `SourcePipelineModal`. This component might behave differently based on a state but for sure shouldn't be one who fetching that...
For now `pipeline` should be treated like something `global` the same as schema which fetched in the line below.

redisinsight/ui/src/pages/rdi/pipeline-management/components/source-pipeline-dialog/SourcePipelineModal.spec.tsx
ArtemHoruzhenko: come on :smile:

redisinsight/ui/src/pages/rdi/pipeline-management/PipelineManagementPage.tsx
pawelangelow: Won't block the PR, but I'm ready to bet it's a matter of time for some weird `useEffects` to start living in the codebase :D

PR #4757, opened 2025-07-25
RI-7210 replace eui file picker

pd-redis:
Before:
image

After:
image

### DIFF
--- a/redisinsight/ui/src/components/base/forms/file-picker/RiFilePicker.tsx
+++ b/redisinsight/ui/src/components/base/forms/file-picker/RiFilePicker.tsx
@@ -13,6 +13,7 @@ import {
 } from 'uiSrc/components/base/forms/file-picker/styles'
 import { CommonProps } from 'uiSrc/components/base/theme/types'
 import ProgressBarLoader from 'uiSrc/components/base/display/progress-bar/ProgressBarLoader'
+import { ColorText } from 'uiSrc/components/base/text'
 
 export type RiFilePickerProps = CommonProps &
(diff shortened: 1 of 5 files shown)


KIvanow:
Couple of questions:
1 - Why are the styles done through sass, instead of Styled components to keep in line with the current Redis UI version and approach?
2 - not related to this PR, but is reduced spacing between the icon and label in the tabs by design?

pd-redis:
> Couple of questions: 1 - Why are the styles done through sass, instead of Styled components to keep in line with the current Redis UI version and approach? 2 - not related to this PR, but is reduced spacing between the icon and label in the tabs by design?
1. Because it was the fastest way to do it. I have not seen actual design for the "File picker" component, so the simplest was to reuse existing styles
2. Can't say, the tabs are not part of this PR, maybe something to put the list to tighten for the UI?

KIvanow (dismissed):

redisinsight/ui/src/components/import-file-modal/styles.module.scss
```diff
@@ -57,11 +57,11 @@
     width: 300px;
 
     :global {
-      .euiFilePicker__showDrop .euiFilePicker__prompt, .euiFilePicker__input:focus + .euiFilePicker__prompt {
+      .RI-File-Picker__showDrop .RI-File-Picker__prompt, .RI-File-Picker__input:focus + .RI-File-Picker__prompt {
         background-color: var(--euiColorEmptyShade);
       }
```
KrumTy, line 62: are these styled used? Or just euiFilePicker* class was replaced with RI-File-Picker*?

PR #4760, opened 2025-07-25
RI-7213: replace EuiImage with img tag

dantovska:
Simply replacing EuiImage with img tag, no difference visually.

### DIFF
--- a/redisinsight/ui/src/pages/rdi/statistics/empty/Empty.tsx
+++ b/redisinsight/ui/src/pages/rdi/statistics/empty/Empty.tsx
@@ -1,4 +1,3 @@
-import { EuiImage } from '@elastic/eui'
 import React from 'react'
 import { useHistory } from 'react-router-dom'
 
@@ -21,7 +20,7 @@ const Empty = ({ rdiInstanceId }: Props) => {
   return (
     <Panel>
       <div className={styles.emptyPipelineContainer} data-testid="empty-pipeline">
-        <EuiImage src={EmptyPipelineIcon} alt="empty" size="s" />
+        <img src={EmptyPipelineIcon} alt="empty" size="s" />
         <Spacer size="xl" />
         <Text>No pipeline deployed yet</Text>
         <Text className={styles.subTitle}>
--- a/redisinsight/ui/src/components/oauth/oauth-sign-in-button/OAuthSignInButton.tsx
+++ b/redisinsight/ui/src/components/oauth/oauth-sign-in-button/OAuthSignInButton.tsx
@@ -1,6 +1,4 @@
 import React from 'react'
-import { EuiImage } from '@elastic/eui'
-
 import { OAuthSsoHandlerDialog } from 'uiSrc/components'
 
 import RedisLogo from 'uiSrc/assets/img/logo_small.svg'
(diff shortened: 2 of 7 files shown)


redisinsight/ui/src/pages/rdi/statistics/empty/Empty.tsx
```diff
@@ -21,7 +20,7 @@ const Empty = ({ rdiInstanceId }: Props) => {
   return (
     <Panel>
       <div className={styles.emptyPipelineContainer} data-testid="empty-pipeline">
-        <EuiImage src={EmptyPipelineIcon} alt="empty" size="s" />
+        <img src={EmptyPipelineIcon} alt="empty" size="s" />
```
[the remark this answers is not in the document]
pawelangelow: @dantovska that sounds like a good comment 😸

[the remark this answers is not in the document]
valkirilov: If we had this component used in many places and we wanted to keep it backwards compatible, potentially we could wrap it into our own component, like `RiImage`, and keep the interface compatible and later on control the size internally, via CSS. However, I see this param is used in a single place, and this sounds like overkill. 

Going for the CSS-only approach sounds reasonable to me as well.

valkirilov (dismissed):

redisinsight/ui/src/pages/rdi/statistics/empty/Empty.tsx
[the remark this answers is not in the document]
dantovska: RiImage sounds good. Implemented it. Tnx!

valkirilov (dismissed):

redisinsight/ui/src/components/base/display/image/RiImage.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+import { RiImageProps, StyledImage } from './image.styles'
```
valkirilov, line 2: Question: Do you usually put the component props interface (in this case `RiImageProps`) into the styles files? I'm used to keeping it closer to the component, just before its definition, and I wonder whether you have a different convention, so I don't break it.

valkirilov (approved):

redisinsight/ui/src/components/base/display/image/RiImage.tsx
dantovska: I've been doing this every time - not a particular reason for it. It is a matter of preference I think, but we don't have a convention.

redisinsight/ui/src/components/base/display/image/image.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+  fullWidth: css`
+    width: 100%;
+  `,
+}
+
+export type RiImageSize = (typeof SIZES)[number]
+
+export interface RiImageProps extends HTMLAttributes<HTMLImageElement> {
+  size?: RiImageSize
+  src?: string
```
pd-redis: I would make at least `src` required, and probably `alt` too.

redisinsight/ui/src/components/base/display/image/image.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+  alt?: string
+}
+
+export const StyledImage = styled.img<
+  Omit<RiImageProps, 'size'> & {
+    size?: RiImageSize
```
pd-redis: this does not do anything, you omit `size` and add it back. It will still be passed to `img` tag.
Use `$size` to tell `styled-components` you don't want to pass this prop to the underlying component.

pawelangelow (approved):

PR #4765, opened 2025-07-28
add playground

pd-redis:
Visit `/playground` to see various theme properties of currently used theme


redisinsight/ui/src/components/main-router/constants/commonRoutes.ts
```diff
@@ -44,4 +46,11 @@ const ROUTES: IRoute[] = [
   },
 ]
 
+if (LOAD_PLAYGROUND) {
+  ROUTES.push({
+    path: '/playground',
```
KrumTy, line 51: theme-playground would be more descriptive imo (also for the components folder)

valkirilov (approved):
Nice!

PR #4767, opened 2025-07-28
RI-7193: Introduce useCreateIndex()

pawelangelow:
When creating an index, we need two key components:
- The data that the index will be built upon
- The index creation command that defines how the data should be indexed (i.e., which fields are involved and in what way)

In the current implementation, both are hardcoded:
- The data comes from a predefined dataset (bikes)
- The index creation command is a fixed string

This PR introduces a hook that encapsulates the logic for index creation, decoupling it from the UI layer.

Key considerations:
- The bikes dataset is large (~1MB), so bundling it with the application needs to be handled carefully.
- ~~Serving it as a static file from the public directory is currently the cleanest approach.
For Electron, this requires additional handling: a plugin to copy the file into the build output and some runtime path resolution logic.~~
- Preset data is added in the same manner we use for other static presets - via bulk import (with new endpoint)
- The commands (both bulk inserts and index creation) do not create duplicates when re-run - they fail silently if the index or keys already exist. This makes repeated runs safe by default.

### DIFF
--- a/redisinsight/ui/src/services/hooks/useDispatchWbQuery.ts
+++ b/redisinsight/ui/src/services/hooks/useDispatchWbQuery.ts
@@ -1,3 +1,4 @@
+import { useCallback } from 'react'
 import { useDispatch } from 'react-redux'
 import { sendWbQueryAction } from 'uiSrc/slices/workbench/wb-results'
 
@@ -10,23 +11,23 @@ export interface UseDispatchWbQueryOptions {
 export const useDispatchWbQuery = () => {
   const dispatch = useDispatch()
 
-  return (
-    data: string | null | undefined,
-    options?: UseDispatchWbQueryOptions,
-  ) => {
-    if (!data) return
+  return useCallback(
+    (data: string | null | undefined, options?: UseDispatchWbQueryOptions) => {
+      if (!data) return
 
-    dispatch(
-      sendWbQueryAction(
-        data,
-        undefined,
-        undefined,
-        {
-          afterAll: options?.afterAll,
-          afterEach: options?.afterEach,
-        },
-        options?.onFail,
-      ),
-    )
-  }
+      dispatch(
+        sendWbQueryAction(
+          data,
+          undefined,
+          undefined,
+          {
+            afterAll: options?.afterAll,
+            afterEach: options?.afterEach,
+          },
+          options?.onFail,
+        ),
+      )
+    },
+    [dispatch],
+  )
 }


redisinsight/ui/vite.config.mjs
```diff
@@ -37,6 +38,17 @@ export default defineConfig({
     svgr({ include: ['**/*.svg?react'] }),
     reactClickToComponent(),
     ViteEjsPlugin(),
+    // Copy public directory contents for Electron builds
+    ...(isElectron ? [
```
valkirilov: I wonder do we need this copy mechanics, and why we can't use the existing approach for filling the test data?

So far, I have used the sample data for my tests, and it seeds both the test data and the indexes related to it.

image

redisinsight/ui/src/pages/vector-search/create-index/hooks/useCreateIndex.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+jest.mock('uiSrc/utils/index/generateFtCreateCommand', () => ({
+  generateFtCreateCommand: () => 'FT.CREATE idx:bikes_vss ...',
+}))
+
+describe('useCreateIndex', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  const defaultParams = {
```
valkirilov: We can type-hint it :) 

```suggestion
  const defaultParams: CreateSearchIndexParameters = {
```

valkirilov (approved):

redisinsight/api/src/modules/bulk-actions/bulk-import.service.ts
```diff
@@ -280,4 +280,38 @@ export class BulkImportService {
...
+   */
+  public async importVectorCollection(
+    clientMetadata: ClientMetadata,
+    dto: ImportVectorCollectionDto,
+  ): Promise<IBulkActionOverview> {
+    try {
+      const collectionFilePath = join(
+        PATH_CONFIG.dataDir,
+        'vector-collections',
+        dto.collection,
```
KrumTy: is the plan to have other hardcoded collections as well? If so, we should change the collection type to prefixed enum

redisinsight/ui/src/pages/vector-search/create-index/hooks/useCreateIndex.ts
```diff
@@ -0,0 +1,67 @@
...
+const collectionNameByPresetDataChoiceMap = {
+  [PresetDataType.BIKES]: 'bikes',
+}
+
+export const useCreateIndex = (): UseCreateIndexResult => {
+  const [success, setSuccess] = useState(false)
+  const [loading, setLoading] = useState(false)
+  const [error, setError] = useState<Error | null>(null)
+
+  const { load } = useLoadData()
```
KrumTy, line 22: useLoadData already exposes loading/error states, can we use them here?

redisinsight/ui/src/pages/vector-search/create-index/hooks/useCreateIndex.ts
```diff
@@ -0,0 +1,67 @@
...
+  const dispatchCreateIndex = useDispatchWbQuery()
+
+  const run = useCallback(
+    async ({ instanceId }: CreateSearchIndexParameters) => {
+      setSuccess(false)
+      setError(null)
+      setLoading(true)
+
+      try {
+        const collectionName =
```
KrumTy, line 32: probably meant to be a parameter for the hook/component but I guess we can figure it our later

redisinsight/ui/src/services/hooks/useDispatchWbQuery.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const useDispatchWbQuery = () => {
+  const dispatch = useDispatch()
+
+  return (
+    data: string | null | undefined,
+    options?: UseDispatchWbQueryOptions,
+  ) => {
+    if (!data) return
+
```
KrumTy: a hook that returns non memoized (in this cate it generates a function each render) is not a good practice. 

It's later used in [useCreateIndex.ts](https://github.com/redis/RedisInsight/pull/4767/files#diff-1d44ce3e1bbf7ed41c8bfbb24bc8d41ac0ca235a9bfac6013e4e56a18b3531c8) as a dependency of the useCallback, rendering it useless

I suggest dropping this hook entirely (used only in 1 place anyway) and moving it's implementation to useCreateIndex; would be easier to understand and manage that way

redisinsight/api/src/modules/bulk-actions/dto/upload-import-file-by-path.dto.ts
```diff
@@ -10,3 +10,14 @@ export class UploadImportFileByPathDto {
...
+
+export class ImportVectorCollectionDto {
+  @ApiProperty({
+    type: 'string',
+    description: 'Collection name to load vector data',
+    example: 'bikes',
+  })
+  @IsString()
+  @IsNotEmpty()
+  collection: string;
```
pd-redis: In my mind `collection` is some structure, when I saw this code, my first thought was - This is string-ified JSON. Not a blocker, but I would change it to `collectionName` or something similar.

redisinsight/ui/src/pages/vector-search/create-index/hooks/useCreateIndex.ts
```diff
@@ -0,0 +1,67 @@
...
+          collectionNameByPresetDataChoiceMap[PresetDataType.BIKES]
+
```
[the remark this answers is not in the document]
pawelangelow: Correct! Will be introduced as we implement the UI changes (a.k.a. the preset buttons) and set up the types for them.

redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.tsx
```diff
@@ -65,6 +66,14 @@ export const VectorSearchCreateIndex = ({
     setStep(step - 1)
   }
 
+  if (loading) {
+    return <>Loading...</>
+  }
+
+  if (success) {
+    return <>Success!</>
```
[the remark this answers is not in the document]
pawelangelow: This is just to showcase the functionality

redisinsight/api/src/modules/bulk-actions/bulk-import.service.ts
pawelangelow: yes - basically, from the current design, there are 3 collections of preset datasets

redisinsight/ui/src/pages/vector-search/create-index/hooks/useCreateIndex.ts
pawelangelow: Well, no. 😄 loading/error for the loadData hook are reflecting if the data is loaded (a.k.a. did we seed everything in the database), while the useCreateIndex loading/error are reflecting the whole operation (seed + index creating). Are they connected - yes, but for this version, i decided to skip them (as we handle hardcoded stuff only). Since this is within try/catch, we get the error and since we await the load, we internally respect the loading as well.

redisinsight/ui/src/pages/vector-search/create-index/hooks/useCreateIndex.ts
pawelangelow: Exactly - as soon as we have all the collection choices, we should handle it here as well.

redisinsight/ui/src/services/hooks/useDispatchWbQuery.ts
pawelangelow: Oh, yeah, good eye 😄 I'll fix the memoization. Can't think of any valid point to leave it as a standalone hook, apart from keeping the logic simple.

redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.tsx
```diff
@@ -65,6 +66,14 @@ export const VectorSearchCreateIndex = ({
     setStep(step - 1)
   }
 
+  if (loading) {
+    return <>Loading...</>
```
pd-redis, line 70: Probably not important at this stage, but we have some "loader" components that you could use

redisinsight/api/src/modules/bulk-actions/bulk-import.service.ts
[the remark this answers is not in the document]
pawelangelow: handled this with whitelisting the colletctions

redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.tsx
pawelangelow: I believe the next PR would totally remove this, I just used it as a showcase.

redisinsight/api/src/modules/bulk-actions/dto/upload-import-file-by-path.dto.ts
pawelangelow: Done

redisinsight/ui/src/services/hooks/useDispatchWbQuery.ts
pawelangelow: @KrumTy, after both functionalities get together, the testing of this hook becomes a total mess, so I'll keep it this way. We can revisit after we build the whole flow and address issues.

PR #4769, opened 2025-07-28
RI-7224 - in messages - ACK / CLAIM buttons need space between

KIvanow:
Removed the old sass styling.
Added a new horizontal spacer (similar to the existing spacer). The same thing could've been done with a margin, but I think we'll need some space here and there as well anyway.
Specifically for @pawelangelow there are a few small tests :) Including validating the parsing of the sizes

### DIFF
--- a/redisinsight/ui/src/pages/browser/modules/key-details/components/stream-details/messages-view/MessageAckPopover/MessageAckPopover.tsx
+++ b/redisinsight/ui/src/pages/browser/modules/key-details/components/stream-details/messages-view/MessageAckPopover/MessageAckPopover.tsx
@@ -6,6 +6,7 @@ import {
   SecondaryButton,
 } from 'uiSrc/components/base/forms/buttons'
 import { RiPopover } from 'uiSrc/components/base'
+import { HorizontalSpacer } from 'uiSrc/components/base/layout'
 import styles from './styles.module.scss'
 
 export interface Props {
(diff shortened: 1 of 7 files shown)


redisinsight/ui/src/pages/browser/modules/key-details/components/stream-details/messages-view/MessageAckPopover/MessageAckPopover.tsx
```diff
@@ -35,15 +36,18 @@ const AckPopover = (props: Props) => {
...
+          <SecondaryButton
+            size="s"
+            aria-label="Acknowledge pending message"
+            onClick={showPopover}
+            className={styles.ackBtn}
+            data-testid="acknowledge-btn"
+          >
+            ACK
+          </SecondaryButton>
+          <HorizontalSpacer size="s" />
```
dantovska, line 49: shouldn't this be a vertical spacer? or are we having the ACK and CLAIM buttons on separate rows?

KIvanow: horizontal, in the sense we are separating things on X / Horizontal axis. We can name it something else if it is confusing (the original spacer is called just that, so I picked what made sense to me, which might be wrong for others)

valkirilov (approved):

PR #4771, opened 2025-07-29
RI-7200 add request metadata to session metadata and exclude it from …

ArtemHoruzhenko:
…logs

### DIFF
--- a/redisinsight/api/src/common/models/session.ts
+++ b/redisinsight/api/src/common/models/session.ts
@@ -1,6 +1,7 @@
 import { IsNotEmpty, IsObject, IsOptional, IsString } from 'class-validator';
 import { BadRequestException } from '@nestjs/common';
 import ERROR_MESSAGES from 'src/constants/error-messages';
+import { Expose } from 'class-transformer';
 
 export interface ISessionMetadata {
   userId: string;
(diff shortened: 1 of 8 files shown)


redisinsight/api/src/common/models/session.ts
```diff
@@ -9,21 +10,31 @@ export interface ISessionMetadata {
...
+  @Expose()
   @IsObject()
   data?: Record<string, any> = {};
 
+  @Expose({ groups: ['security'] })
+  @IsObject()
+  @IsOptional()
+  requestMetadata?: Record<string, any> = {};
+
+  @Expose()
```
pawelangelow, line 27: Should we expose all of those? For example `sessionId`?

redisinsight/api/src/common/models/session.ts
```diff
@@ -9,21 +10,31 @@ export interface ISessionMetadata {
...
   @IsNotEmpty()
   @IsString()
   userId: string;
 
+  @Expose()
   @IsObject()
   data?: Record<string, any> = {};
 
+  @Expose({ groups: ['security'] })
```
pawelangelow, line 22: Nit: this should become `Groups.Security` constant as it's on a lot of places

redisinsight/api/src/common/models/session.ts
ArtemHoruzhenko: yes

redisinsight/api/src/common/models/session.ts
ArtemHoruzhenko: Correct but it is not a single place. Didn't want to introduce more changes in bunch of not related to this PR files

PR #4772, opened 2025-07-29
Introduce factories for mocking test data

valkirilov:
# Description

Introduce factories for dynamic reusable mocks (primarily used in the tests)
- integrate [fishery](https://github.com/thoughtbot/fishery) to provide simple and handy tools for building factories
- integrate [faker](https://fakerjs.dev/) to provide utilities for generating dynamic and realistic-looking data
- created basic example factories for the vector search unit tests

## Why Factories?

Factories allow us to define a single source of truth for how our mock data should look, based on the actual interfaces and types used in our application. This means we can reuse the same factory across multiple test files and scenarios, reducing duplication and maintenance overhead.

Additionally, by using factories, all test data is generated in a consistent way, ensuring that changes to the data structure only need to be made in one place.

And this approach comes with the bonus that factories can encapsulate any "business logic" required to create valid objects, keeping this complexity out of individual test cases. This makes tests easier to read and maintain.

## Why Faker?

Faker generates realistic random values for names, numbers, sentences, and more. This helps us avoid "magic values" and makes our tests more robust by simulating real-world scenarios. This way, we can easily customize the shape and content of our test data, including edge cases, without manual effort.

# How to use it

## How to build factories

You can always refer to the [official docs](https://github.com/thoughtbot/fishery?tab=readme-ov-file#define-and-use-factories) (and it's recommended to go through it and learn more about the built-in concepts), but here is a quick example:

```js
export const exampleFactory = Factory.define<ExampleInterface>(() => ({
  value: faker.string.alpha({ length: { min: 5, max: 12 } }),
  label: faker.word.noun(),
  text: faker.lorem.sentence(),
  tag: faker.helpers.enumValue(FieldTypes),
  disabled: faker.datatype.boolean(),
}))
```

## How to use factories in tests 

Here are the most common scenarios for building mock data using the factories you built.

```js
// Create new objects that conform to the interface, but with dynamic and realistic data
const example = exampleFactory.build();

// Create multiple, different objects
const examples = example.buildList(3);

// Create new objects following the interface, but explicitly override some of the fields to match our needs
const example = exampleFactory.build({
  value: 'id',
  label: 'ID',
  text: 'Primary Key field',
  tag: FieldTypes.TAG,
});
```


redisinsight/ui/src/components/new-index/create-index-step/field-box/FieldBox.spec.tsx
```diff
@@ -27,14 +26,7 @@ describe('CreateIndexStepWrapper', () => {
...
     const props: FieldBoxProps = {
-      box: {
-        ...MOCK_VECTOR_SEARCH_BOX,
-        value: 'id',
-        label: 'id',
-        text: 'Unique product identifier',
-        tag: FieldTypes.TAG,
-        disabled: false,
-      },
+      box: vectorSearchBoxFactory.build(),
```
valkirilov, line 29: This is how we usually create objects with the factory. It will return dynamic and "realistic" data, based on the factory implementation. 

Every time, the generated data will be different, which helps us test various test scenarios, and sometimes we may even handle use cases we forgot to cover in our tests (like how our system behaves when we pass longer text, but we have a limit... and so on).

redisinsight/ui/src/components/new-index/create-index-step/field-box/FieldBox.spec.tsx
```diff
@@ -53,23 +45,31 @@ describe('CreateIndexStepWrapper', () => {
     expect(checkbox).toBeInTheDocument()
   })
 
-  it('should select the box when clicked', () => {
-    renderFieldBoxComponent()
+  it('should select the box when clicked', async () => {
+    const props: FieldBoxProps = {
+      box: vectorSearchBoxFactory.build({
+        disabled: false,
```
valkirilov, line 51: We can always pass our preferences to the factory and make sure the end object will contain our desired values. In this case, we need to ensure the box is not disabled, so we can click it.

redisinsight/ui/src/mocks/factories/redisearch/IndexInfo.factory.ts
```diff
@@ -0,0 +1,108 @@
...
+export const indexInfoFactory = Factory.define<IndexInfoDto>(() => ({
+  index_name: `idx:${faker.word.noun()}`,
+  index_options: {},
+  index_definition: {
+    key_type: 'JSON',
+    prefixes: [`$${faker.word.noun()}:`],
+    default_score: '1',
+    indexes_all: 'false',
+  },
+  attributes: indexInfoAttributeFactory.buildList(3),
```
valkirilov, line 21: Factories can be used with composition, and extract some common logic, and make it reusable:
Basically, it reduces the size of the original "object" by removing the repetitive lines for the different types of attributes. 

_PS: This factory produces big object, so you can't see the power of this composition, but usually it makes it way simpler and more readable._

redisinsight/ui/src/mocks/factories/redisearch/IndexInfoTableData.factory.ts
```diff
@@ -0,0 +1,36 @@
+import { Factory } from 'fishery'
+import { faker } from '@faker-js/faker'
+import { IndexInfoTableData } from 'uiSrc/pages/vector-search/manage-indexes/IndexAttributesList'
+import { FieldTypes } from 'uiSrc/pages/browser/components/create-redisearch-index/constants'
+
+export const INDEX_INFO_SEPARATORS: string[] = [',', ';', '|', ':']
+
+type IndexInfoTableDataFactoryTransientParams = {
```
valkirilov, line 8: A bit more complex example, for a handy feature. Consider the scenario where you want to create a specific type of object, and you need a way to instruct the factory to do so. 

In this case, we have two optional fields (weight and separator), and here is how to build objects that have them prefilled:

```js
const example = indexInfoTableDataFactory.build({}, {
    includeWeight: true,
    includeSeparator: true
})
```

Result

```js
{
    attribute: "lorem-ipsum",
    type: "TEXT",
    weight: "1",
    separator: ","
}
```
Of course, if you don't pass the transient params, you rely on the default behavior of the factory, which will treat these two fields as optional. So you may end up without them defined in the result object. 


You can always check the official docs on [transient params](https://github.com/thoughtbot/fishery?tab=readme-ov-file#params-that-dont-map-to-the-result-object-transient-params) for more in-depth examples..

redisinsight/ui/src/pages/vector-search/manage-indexes/IndexAttributesList.spec.tsx
```diff
@@ -1,30 +1,14 @@
...
-        attribute: 'content',
-        type: 'TEXT',
-      },
-      {
-        attribute: 'tags',
-        type: 'TAG',
-        weight: '1.0',
-      },
-    ],
+    data: indexInfoTableDataFactory.buildList(3),
```
valkirilov, line 11: Instead of copy/pasting the same object N times and changing its values, we can simply ask the factory to generate an array of objects for us, and know they'll match the interface we expect.

yarn.lock
```diff
@@ -2414,7 +2419,7 @@
   resolved "https://registry.yarnpkg.com/@redis-ui/icons/-/icons-4.16.1.tgz#87fc6ab073a203de6b313a4868c0a84f076c8580"
   integrity sha512-ZRQWzambHq9A/5zWx4HXEBNU8uCFwf6Ap0NKZryoMq5iGfygBF8k8A89RmwUXuSlJ99NRxW+RvmbcMd+Tyosmg==
 
-"@redis-ui/styles@^11.4.0":
+"@redis-ui/styles@^11.0.2", "@redis-ui/styles@^11.4.0":
```
pawelangelow, line 2422: I'm pretty sure this isn't part of this PR, as I saw it on my branches as well 😸

pawelangelow (dismissed):

yarn.lock
valkirilov: It's the `yarn.lock` file, simply because I ran `yarn install` for the new packages, and the fact that this package is marked with `^`, maybe that's why it was bumped as well.

dantovska (approved):

redisinsight/ui/src/components/new-index/create-index-step/field-box/FieldBox.spec.tsx
```diff
@@ -1,16 +1,15 @@
...
 import { cleanup, fireEvent, render, screen } from 'uiSrc/utils/test-utils'
+import { vectorSearchBoxFactory } from 'uiSrc/mocks/factories/redisearch/VectorSearchBox.factory'
 
 import { FieldBox, FieldBoxProps } from './FieldBox'
 import { VectorSearchBox } from './types'
 
 const renderFieldBoxComponent = (props?: FieldBoxProps) => {
   const defaultProps: FieldBoxProps = {
-    box: MOCK_VECTOR_SEARCH_BOX,
+    box: vectorSearchBoxFactory.build(),
```
pd-redis, line 12: Isn't this going to add overhead, generating new data on each execution?

valkirilov: To be honest, I didn't experience issues related to that, but in case we feel it's adding unnecessary overhead, we can always create new objects once, in the beginning of the test suite.

```js
const mockVectorSearchBox = vectorSearchBoxFactory.build();

const renderFieldBoxComponent = (props?: FieldBoxProps) => {
  const defaultProps: FieldBoxProps = {
    box: mockVectorSearchBox,
  }
  
  return render(
    <BoxSelectionGroup.Compose>
      <FieldBox {...defaultProps} {...props} />
    </BoxSelectionGroup.Compose>,
  )
}
```

PR #4778, opened 2025-07-30
RI-7263: Replace EUI empty prompt

dantovska:
A wrapper, created for icons and some content... 
Wraps the bouncy logo and the error indicating component:
image
image

### DIFF
--- a/redisinsight/ui/src/components/base/layout/empty-prompt/RiEmptyPrompt.tsx
+++ b/redisinsight/ui/src/components/base/layout/empty-prompt/RiEmptyPrompt.tsx
@@ -1,5 +1,7 @@
 import React, { HTMLAttributes } from 'react'
 import styled from 'styled-components'
+import { useTheme } from '@redis-ui/styles'
+import { Spacer } from '../spacer'
 
 interface RiEmptyPromptProps extends Omit<HTMLAttributes<HTMLDivElement>, 'title'> {
   body?: React.ReactNode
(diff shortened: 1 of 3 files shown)


redisinsight/ui/src/components/base/layout/empty-prompt/RiEmptyPrompt.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+const StyledEmptyPrompt = styled.div`
+  max-width: 36em;
+  text-align: center;
+  padding: 24px;
+  margin: auto;
+`
+
+const Spacer = styled.div`
```
pd-redis: nit, but can't you use the `Spacer` component?

pd-redis:
The tests fail

redisinsight/ui/src/components/base/layout/empty-prompt/RiEmptyPrompt.tsx
dantovska: I guess I could, but the Spacer is not using the theme spacing but the hard coded variables (--size-xs, --size-s etc.) Should I refactor them along with this?

pd-redis: It would be perfect if you do :)

pd-redis: Or:
see https://styled-components.com/docs/api#createglobalstyle 
add global style, where we can setup those variables from the theme

KIvanow: @pd-redis is this similar to what you meant? https://github.com/redis/RedisInsight/pull/4778/commits/cc6c1413353bc7d3ab97370d938e9c97f251fe95 It allows for the spacer to be used both with old sizes (strings such as 'l', 'm', etc), the theme sizes as strings ('space050', 'space000', etc), as well as the theme sizes directly (theme.core.space.space100)

PR #4784, opened 2025-08-01
Fe/feature/ri 7269   fix major discrepancies between the ds previews and the implementation around the main content under the new navigation

KIvanow:
Fixed the differences between the previews we have with the current state around the general layout.
Went from this
image
to this
image
List of changes:

- fixed paddings for the CLI row
- updated the CLI row to be closer to the previews - fixed border size and positioning
- added spacing under the navigation 
- removed background for the bottom buttons (add, cancel, save, etc). In the new DS there is no different BG and it only makes it mroe complciated to keep track
- added borders to the keys list and key details
- updated main content brackgrounds to use the proper colors
- added border above the new navigation

### DIFF
--- a/redisinsight/ui/src/pages/browser/BrowserPage.tsx
+++ b/redisinsight/ui/src/pages/browser/BrowserPage.tsx
@@ -4,7 +4,7 @@ import { useDispatch, useSelector } from 'react-redux'
 import cx from 'classnames'
 
 import { isNumber } from 'lodash'
-import { theme } from '@redis-ui/styles'
+import { useTheme } from '@redis-ui/styles'
 import {
   formatLongName,
   getDbIndex,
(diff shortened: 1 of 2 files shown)


redisinsight/ui/src/pages/browser/BrowserPage.tsx
```diff
@@ -4,6 +4,7 @@ import { useDispatch, useSelector } from 'react-redux'
 import cx from 'classnames'
 
 import { isNumber } from 'lodash'
+import { theme } from '@redis-ui/styles'
```
pd-redis: `useTheme` is better option

redisinsight/ui/src/templates/instance-page-template/InstancePageTemplate.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { useCallback, useEffect, useRef, useState } from 'react'
 import { useSelector } from 'react-redux'
+import { theme } from '@redis-ui/styles'
```
pd-redis: same, `useTheme`

pd-redis (approved):

PR #4788, opened 2025-08-01
RI-7195: Add search screen

pawelangelow:
This PR introduces the `Search` screen, which is intended to be displayed when there are existing indexes.

Since the functionality is quite similar to Workbench, I attempted to reuse some of its logic. However, due to structural differences, full reuse isn't straightforward at this stage. Here's a breakdown of what's included in this change:

- useQuery – This is the core piece of the PR and deserves the most attention during review. It replicates the same command execution flow as Workbench, but implemented as a React hook rather than going through Redux.
  _Note: Unit tests are not included in this PR but will be added later in the flow once the behavior stabilizes._
- VectorSearchQuery – A resizable layout wrapper combining the Query and Results panes.
- Extended API for QueryCardHeader – Added a hideFields prop to control the visibility of specific elements like "Profiling" and the "View type" toggle.
- Temporary duplication of QueryCard and CommandsView components – Similar to #4794, this is a workaround to avoid breaking Workbench while the Search feature evolves. Once we finalize some open design decisions, we can revisit this and consolidate the components properly.

If you want to test with the provided data, use these commands:
- `FT.SEARCH idx:bikes_vss "@brand:Nord" SORTBY price ASC`
- `FT.SEARCH idx:bikes_vss "@material:{alloy} @weight:[0 20]"`

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
+++ b/redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
@@ -47,7 +47,7 @@ export const VectorSearchQuery = () => {
             resultsMode={resultsMode}
             setQuery={setQuery}
             setQueryEl={() => {}}
-            onSubmit={onSubmit}
+            onSubmit={() => onSubmit()}
             onQueryChangeMode={onQueryChangeMode}
             onChangeGroupMode={onChangeGroupMode}
             queryProps={{ useLiteActions: true }}


redisinsight/ui/src/components/query/query-lite-actions/QueryLiteActions.tsx
```diff
@@ -0,0 +1,81 @@
...
+        position="right"
+        content={
+          isLoading
+            ? 'Please wait while the commands are being executed…'
+            : 'Clear query'
+        }
+        data-testid="clear-query-tooltip"
+      >
+        <EmptyButton
+          onClick={() => {
```
valkirilov: nit: Why not?

```suggestion
          onClick={onClear}
```

PS: Same for the `onSubmit` bellow.

pawelangelow: 👍 , but I extracted that change in different PR

redisinsight/ui/src/pages/vector-search/components/commands-view/CommandsView/styles.module.scss
```diff
@@ -0,0 +1,43 @@
+.wrapper {
```
valkirilov, line 1: _I noticed that this component is a replica of its counterpart in [Wokbench](https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/workbench/components/wb-results/WBResults/styles.module.scss), so this comment doesn’t apply to this specific case._ 

Question: However, I’m curious, do we have a general strategy for using styled components over SASS modules, or are we still not clear about the preferred way to style our custom components?

redisinsight/ui/src/pages/vector-search/components/commands-view/CommandsViewWrapper.tsx
```diff
@@ -0,0 +1,37 @@
...
+  onQueryProfile: (
+    query: string,
+    commandId?: Nullable<string>,
+    executeParams?: CodeButtonParams,
+  ) => void
+}
+
+const CommandsViewWrapper = (props: Props) => <CommandsView {...props} />
+
+export default React.memo(CommandsViewWrapper
```
valkirilov: nit: You can fix the formatting.

PS: We can think of enforcing the prettier formatting rules as a pre-commit hook, to avoid such minor discrepancies, and unify the codebase "look and feel".

redisinsight/ui/src/pages/vector-search/VectorSearchPage.tsx
```diff
@@ -4,13 +4,11 @@ import { VectorSearchCreateIndex } from './create-index/VectorSearchCreateIndex'
 import { VectorSearchQuery } from './query/VectorSearchQuery'
 
 export const VectorSearchPage = () => {
-  const hasIndexes = false
+  const hasIndexes = true
```
valkirilov, line 7: Question: Is that flip intentional (you had discussions with @KrumTy), or simply you forgot to stash it before pushing, since it was easier to test your changes?

redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
```diff
@@ -1,11 +1,92 @@
...
+
+      <StyledResizableContainer direction="vertical">
+        <ResizablePanel id="top-panel" minSize={20} defaultSize={30}>
+          <QueryWrapper
+            query={query}
+            activeMode={activeMode}
+            resultsMode={resultsMode}
+            setQuery={setQuery}
+            setQueryEl={() => {}}
+            onSubmit={onSubmit}
```
valkirilov: There is something odd happening here, but submitting the query, it doesn't work for me. The query is not passed properly to the `handleSubmit` in the `useQuery` hook, and nothing happens. 

However, if I make this simple change, it works.

```
onSubmit={() => onSubmit()}
```

redisinsight/ui/src/pages/vector-search/query/useQuery.ts
```diff
@@ -0,0 +1,378 @@
...
+    })
+
+    await addCommands(reverse(data))
+
+    if (restCommands.length > 0) {
+      const nextCommands = restCommands[0]
+      if (nextCommands?.length) {
+        setTimeout(() => {
+          handleSubmit(nextCommands.join('\n'), undefined, executeParams)
+        }, 100)
```
valkirilov: It looks like a fake wait, do you know why we need it, what we're trying to bypass here?

redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
```diff
@@ -1,11 +1,92 @@
...
+            scrollDivRef={scrollDivRef}
+            hideFields={[HIDE_FIELDS.profiler, HIDE_FIELDS.viewType]}
+            onQueryReRun={onQueryReRun}
+            onQueryProfile={onQueryProfile}
+            onQueryOpen={onQueryOpen}
+            onQueryDelete={onQueryDelete}
+            onAllQueriesDelete={onAllQueriesDelete}
+            noResultsPlaceholder={
+              <StyledNoResultsWrapper>
+                TODO: Not sure yet what to put here
```
valkirilov, line 84: https://www.lipsum.com/ will look great! 😀

redisinsight/ui/src/pages/vector-search/components/commands-view/CommandsView/styles.module.scss
pawelangelow: Usually yes, but if this happens to be our final implementation (a.k.a. Workbench vs Search), I plan to refactor the common components and fix the styles. Consider this a "hack" because I don't want to spend much time on something that may be deleted very soon.

redisinsight/ui/src/pages/vector-search/components/commands-view/CommandsViewWrapper.tsx
pawelangelow: Oh, G! I missed that. Thanks! We are kind of ready with this, but it's on the main branch. I wanted to merge it in the eui branch, but decided maybe there would be too many conflicts, so as soon as we merge the eui branch back to main, we'll fix all these and will have everything needed to achieve a consistent code style.

redisinsight/ui/src/pages/vector-search/VectorSearchPage.tsx
pawelangelow: There's no particular reason for it - just wanted to make it easier for someone who does the review to be able to see the page. It really doesn't matter how we are going to merge it IMO, as the next PR should be to determine this, based on the actual database. Maybe I'm missing something, so please, elaborate if needed.

redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
pawelangelow: In the designs, we have a suggestion for a link to the create index wizard, so probably that would be it, but as we don't have it, I decided not to block the PR.

redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
```diff
@@ -1,11 +1,92 @@
...
+  return (
+    <CreateIndexWrapper direction="column" justify="between">
+      <HeaderActions />
+
+      <StyledResizableContainer direction="vertical">
+        <ResizablePanel id="top-panel" minSize={20} defaultSize={30}>
+          <QueryWrapper
+            query={query}
+            activeMode={activeMode}
+            resultsMode={resultsMode}
```
valkirilov, line 47: I'm not sure whether it's an issue with the way I test it, but I've noticed the following issue.

If we run queries on the **Workbench** page, I see a button with the option to toggle the way we display the results - via text, or render them in a table.

Image

However, when I go to the new **Vector Search** page and run a query, I don't see the button for toggling the view.

Image

redisinsight/ui/src/pages/vector-search/query/useQuery.ts
pawelangelow: Yes, it is a leftover. My first implementation was in memory, without sending it to the API, so I forgot to remove it. I spent some time checking if we need to do it anyways, but for some reasonable size of commands, without it, it works just as fine.

redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
pawelangelow: Yes, this is intended - the designs use a very lite version of that component (the one that displays the command + the controls for it - re-reun, delete, expand, etc..). So I just copied the one from the workbench and disabled some stuff - the view type changer + profile button. If you see the last methods of the `useQuery` hook - we don't support them at all for this page.

redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
pawelangelow: Something strange happened here - maybe something related to the rebase for the other PR. Bsically the problem was sending the native HTML event directly `onSubmit={onSubmit}`, which overrides the expected command parameter. This signature is needed, because we manually trigger it when there are multiline commands. So I just wrapped it in an empty callback to make sure we use proper inputs.

redisinsight/ui/src/pages/vector-search/query/useQuery.ts
```diff
@@ -0,0 +1,380 @@
...
+        setIsLoaded(true)
+      } catch (error) {
+        setIsLoaded(true)
+      }
+    }
+
+    loadHistory()
+  }, [instanceId])
+
+  const prepareNewItems = (
```
KrumTy: can extract from hook

KrumTy (commented):
overall the useQuery hook seems really complex, it lacks tests as well
can it be restructured/split in a way or the functionality has to be together?

redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
valkirilov: Thanks for the information. I asked because we have telemetry events planned for these buttons, and `SEARCH_RESULT_VIEW_CHANGED` is also included. Anyway, if it's not part of the design, I can omit this one for now, and cover the rest.

valkirilov (dismissed):

KrumTy (approved):

PR #4794, opened 2025-08-01
RI-7195: Extend Query component

pawelangelow:
This is a prerequisite PR for https://github.com/redis/RedisInsight/pull/4788

In order to match the design, we need a lite version of the `<Query />` actions. Since the final state isn't clear, I decided to "hack" the current component. After we get the bigger picture, we can think of proper refactoring.


redisinsight/ui/src/pages/workbench/components/query/QueryWrapper.tsx
```diff
@@ -18,6 +18,7 @@ export interface Props {
   query: string
   activeMode: RunQueryMode
   resultsMode?: ResultsMode
+  queryProps?: any
```
valkirilov: Is there any chance to type-hint this, or it's a legacy and is not possible at the moment

valkirilov (approved):

PR #4797, opened 2025-08-04
Fe/feature/ri 7252   rdi   broken layout for the list of jobs

KIvanow:
The [InlineItemEditor](https://github.com/redis/RedisInsight/blob/d2f6b9a613f707d135fcf21c6b8f734ad4ea7236/redisinsight/ui/src/components/inline-item-editor/InlineItemEditor.tsx) is overengineered and used in too many places, doing too many things.
This PR introduces props for custom stylings of different parts of the component, so we can tweak it for the various places it is used in.
In the JSON inline editing, it fixes this
image
turning it to this
image
And in the RDI jobs from
image
to
image

A proper refactoring and separating the component into a different one will be done in a separate task during the UX updates later

### DIFF
--- a/redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/rejson-scalar/RejsonScalar.tsx
+++ b/redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/rejson-scalar/RejsonScalar.tsx
@@ -96,6 +96,17 @@ const RejsonScalar = (props: JSONScalarProps) => {
               {editing ? (
                 <div className="jsonItemEditor">
                   <InlineItemEditor
+                    styles={{
+                      inputContainer: {
+                        height: `24px`,
+                      },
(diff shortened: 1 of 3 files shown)


redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/rejson-scalar/RejsonScalar.tsx
```diff
@@ -96,6 +96,17 @@ const RejsonScalar = (props: JSONScalarProps) => {
               {editing ? (
                 <div className="jsonItemEditor">
                   <InlineItemEditor
+                    styles={{
+                      inputContainer: {
+                        height: `24px`,
+                      },
+                      input: {
+                        height: `24px !important`
```
pd-redis, line 104: If style is used inline, I don't think `important` does anything

PR #4803, opened 2025-08-05
RI-7196: add SavedQueriesScreen

KrumTy:
## Description

Adds UI component for Saved queries section

Screenshot 2025-08-05 at 18 02 48


redisinsight/ui/src/pages/vector-search/saved-queries/mocks.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+
+import { SavedIndex } from './types'
+import { SavedQueriesScreen } from './SavedQueriesScreen'
+
+export const savedIndexes: SavedIndex[] = [
```
valkirilov: These mocks look like a perfect candidate for test factories and Faker.

You can check for examples here:

redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.spec.tsx
```diff
@@ -0,0 +1,187 @@
...
+import React from 'react'
+import { render, screen, fireEvent } from '@testing-library/react'
+
+import { SavedQueriesScreen } from './SavedQueriesScreen'
+import { SavedIndex } from './types'
+
+const mockOnIndexChange = jest.fn()
+const mockOnQueryInsert = jest.fn()
+
+const mockSavedIndexes: SavedIndex[] = [
```
valkirilov, line 13: If you define a factory, you can reuse it in this test suite as well, instead of copying/pasting the same mocks.

redisinsight/ui/src/pages/vector-search/saved-queries/mocks.tsx
KrumTy: I will be deleting this whole file once we link it to real data; it's just for component showcase

redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.spec.tsx
KrumTy: I'm mocking the component in mocks.tsx, will be deleting the whole file once integrated and linked to actual data

redisinsight/ui/src/pages/vector-search/query/HeaderActions.spec.tsx
```diff
@@ -10,11 +10,20 @@ jest.mock('uiSrc/components/base/layout/drawer', () => ({
   DrawerHeader: jest.fn().mockReturnValue(null),
 }))
 
-const renderComponent = () => render(<HeaderActions />)
+const mockProps = {
```
valkirilov:
```suggestion
const mockProps: HeaderActionsProps = {
```

redisinsight/ui/src/pages/vector-search/query/HeaderActions.spec.tsx
```diff
@@ -10,11 +10,20 @@ jest.mock('uiSrc/components/base/layout/drawer', () => ({
...
+  isManageIndexesDrawerOpen: false,
+  setIsManageIndexesDrawerOpen: jest.fn(),
+  isSavedQueriesOpen: false,
+  setIsSavedQueriesOpen: jest.fn(),
+}
 
-describe('ManageIndexesDrawer', () => {
+const renderComponent = (props = mockProps) =>
```
valkirilov, line 20:
```suggestion
const renderComponent = (props: HeaderActionsProps = mockProps) =>
```

redisinsight/ui/src/pages/vector-search/query/HeaderActions.tsx
```diff
@@ -1,24 +1,36 @@
...
 
-  return (
-    <>
-      <StyledHeaderAction data-testid="vector-search-header-actions">
-        <StyledTextButton>Saved queries</StyledTextButton>
-        <StyledTextButton onClick={() => setIsManageIndexesDrawerOpen(true)}>
-          Manage indexes
-        </StyledTextButton>
-      </StyledHeaderAction>
+export const HeaderActions = ({
```
valkirilov, line 12: nit: It's not a big deal at the moment, we're prop-drilling these only one level deep, but personally, I would encapsulate these into context, only for these components, so we can share these state variables in a clean way. 

Also, this way, it will stand as a good base for future extension, in case we want to add more state variables.

```typescript
type HeaderActionsContextType = {
  isManageIndexesDrawerOpen: boolean
  setIsManageIndexesDrawerOpen: (value: boolean) => void
  isSavedQueriesOpen: boolean
  setIsSavedQueriesOpen: (value: boolean) => void
}

const HeaderActionsContext = createContext<HeaderActionsContextType | undefined>(undefined)

export const HeaderActionsProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
  const [isManageIndexesDrawerOpen, setIsManageIndexesDrawerOpen] = useState(false)
  const [isSavedQueriesOpen, setIsSavedQueriesOpen] = useState(false)

  return (
    <HeaderActionsContext.Provider
      value={{
        isManageIndexesDrawerOpen,
        setIsManageIndexesDrawerOpen,
        isSavedQueriesOpen,
        setIsSavedQueriesOpen,
      }}
    >
      {children}
    </HeaderActionsContext.Provider>
  )
}

export const useHeaderActions = () => {
  const context = useContext(HeaderActionsContext)
  if (!context) {
    throw new Error('useHeaderActions must be used within a HeaderActionsProvider')
  }
  return context
}
```

And then, we can simply use it as a hook, instead of prop-drilling it:
```typescript
const MyComponent = () => {
  const {
    isManageIndexesDrawerOpen,
    setIsManageIndexesDrawerOpen,
  } = useHeaderActions()

  // ...use these as needed
}
```

redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.spec.tsx
```diff
@@ -0,0 +1,187 @@
...
+  onIndexChange: mockOnIndexChange,
+  onQueryInsert: mockOnQueryInsert,
+}
+
+describe('SavedQueriesScreen', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('should render', () => {
```
valkirilov, line 52: nit: I know (and agree that tests should be atomic), but I think all of these fall into the same category of verifying the rendered UI, so we can combine them into one:
- it should render the main title
- it should render the main title
- it should render the index selector label
- it should render all queries for the selected index

redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.tsx
```diff
@@ -0,0 +1,84 @@
...
+                    <FieldTag key={tag} tag={tag as any} />
+                  ))}
+                </TagsWrapper>
+              )
+            }
+          />
+        </VectorSearchSavedQueriesSelectWrapper>
+        {selectedIndex?.queries.map((query) => (
+          // @ts-expect-error: key prop
+          <VectorSearchScreenBlockWrapper key={query.value}>
```
valkirilov: You can remove the `@ts-expect-error`
```suggestion
          <VectorSearchScreenBlockWrapper key={query.value} as="div">
```

valkirilov: nit: I would put these boxes into a separate component, so we can have it isolated, and keep the main wrapper component simpler. Also, it will have its own tests to verify whether the query is rendered properly and the run button works as expected, instead of flooding the current test suite of the wrapper.

redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.spec.tsx
```diff
@@ -0,0 +1,187 @@
...
+  })
+
+  it('should render insert buttons for each query', () => {
+    render(<SavedQueriesScreen {...defaultProps} />)
+
+    const insertButtons = screen.getAllByText('► Insert')
+    expect(insertButtons).toHaveLength(2) // 2 queries in the selected index
+  })
+
+  it('should call onQueryInsert when insert button is clicked', () => {
```
valkirilov, line 74: If you reference this [comment](https://github.com/redis/RedisInsight/pull/4803/files#r2256229088), you can move these tests into a separate suite that belongs to the new component;

redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.spec.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+      expect(screen.queryByText('► Insert')).not.toBeInTheDocument()
+    })
+  })
+
+  describe('Component Structure', () => {
+    it('should render the main wrapper components', () => {
+      const { container } = render(<SavedQueriesScreen {...defaultProps} />)
+
+      expect(container.firstChild).toBeInTheDocument()
```
valkirilov: Why do we need this test, and what do we actually verify with it?

redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.spec.tsx
```diff
@@ -0,0 +1,187 @@
...
+    })
+
+    it('should have proper button text for insert actions', () => {
+      render(<SavedQueriesScreen {...defaultProps} />)
+
+      const insertButtons = screen.getAllByText('► Insert')
+      insertButtons.forEach((button) => {
+        expect(button).toBeInTheDocument()
+        expect(button.tagName).toBe('BUTTON')
```
valkirilov: Why do we care about the type of element? I think we shouldn't test implementation details, I don't see the value of such tests.

redisinsight/ui/src/pages/vector-search/useRedisearchListData.ts
```diff
@@ -0,0 +1,31 @@
...
+import { useSelector, useDispatch } from 'react-redux'
+
+import {
+  redisearchListSelector,
+  fetchRedisearchListAction,
+} from 'uiSrc/slices/browser/redisearch'
+import { connectedInstanceSelector } from 'uiSrc/slices/instances/instances'
+import { isRedisearchAvailable } from 'uiSrc/utils'
+
+export const useRedisearchListData = () => {
```
valkirilov, line 11: Can you add a few tests for this hook as well 😬

redisinsight/ui/src/pages/vector-search/query/HeaderActions.tsx
```diff
@@ -1,24 +1,36 @@
...
+}: HeaderActionsProps) => (
+  <>
+    <StyledHeaderAction data-testid="vector-search-header-actions">
+      <StyledTextButton
+        variant="primary"
+        onClick={() => setIsSavedQueriesOpen(!isSavedQueriesOpen)}
+      >
+        Saved queries
+      </StyledTextButton>
+      <StyledTextButton onClick={() => setIsManageIndexesDrawerOpen(true)}>
```
valkirilov, line 26: There is something wrong with this state, for some reason, we can no longer open the drawer.

https://github.com/user-attachments/assets/ae33b0f7-1a2a-42e3-8475-420fbc4435bb

redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.tsx
KrumTy: I will leave it as it is for simplicity and avoid prop drilling; we can refactor in the main branch if needed

redisinsight/ui/src/pages/vector-search/query/HeaderActions.tsx
KrumTy: Initially I had the HeaderActions component removed entirely but I reverted it to simplify the PR. I don't think context is needed in this case. Prop drilling is just one level and context just for this section would overcomplicate stuff. 

If we ever need context we'd better add it as a top-level provider/wrapper to the whole feature.

redisinsight/ui/src/pages/vector-search/useRedisearchListData.ts
KrumTy: it's a functionality I obviously extracted over from another place, which took it from a 3rd one. 

It's also not really working as intended imo so lets revisit this in the main branch when tackling bugs? In this specific hook, the loading state does not behave as expected (reload the page to reproduce), indicating an issue related to the redux slice.

valkirilov (approved):

pawelangelow (approved):

PR #4809, opened 2025-08-06
fix: resolve redux store circular dependencies

KrumTy:
## Description

Issue: In jest ui unit tests we use configureMockStore instead of the real Redux store. The mock store doesn't actually use the reducers - it just holds state. However, when the test runs, some component or hook is trying to access the actual Redux store configuration, which expects the reducers to be properly defined. Hence we get a lot of errors in the console, that usually happen after a test finishes successfully.

Solution: Import only types from store where possible. Use store-dynamic wrapper of the store methods where not possible.

A better solution would require large refactor of the code base.


valkirilov (approved):

jest.config.cjs
```diff
@@ -11,6 +11,7 @@ module.exports = {
     '\\.svg': '<rootDir>/redisinsight/__mocks__/svg.js',
     '\\.(css|less|sass|scss)$': 'identity-obj-proxy',
     '\\.scss\\?inline$': '<rootDir>/redisinsight/__mocks__/scssRaw.js',
+    'uiSrc/slices/store$': '<rootDir>/redisinsight/ui/src/utils/test-store.ts',
```
valkirilov, line 14: Why do we need this alias? Can't we use the one we already have (below)?

redisinsight/ui/src/utils/test-utils.tsx
```diff
@@ -94,6 +96,7 @@ const initialStateDefault: RootState = {
     csrf: cloneDeep(initialStateAppCsrfReducer),
     init: cloneDeep(initialStateAppInit),
     connectivity: cloneDeep(initialStateAppConnectivity),
+    dbSettings: cloneDeep(initialStateAppDbSettings),
```
valkirilov, line 99: Yeah, I forgot that this one was missing. Kudos for resolving it 👌

jest.config.cjs
KrumTy: this is to override and redirect all imports from 'uiSrc/slices/store' to test-store.ts (during jest tests only)

valkirilov (dismissed):

pawelangelow (approved):

pd-redis (approved):

valkirilov (approved):

PR #4811, opened 2025-08-06
RI-7218 Add telemetry collection for Vector Search page

valkirilov:
# Description

Added telemetry collection events to track users' interaction with the Vector Search page:
- when users run a query
- when users copy a query result
- when users toggle the full-screen mode for a query result
- when users expand/collapse a query result
- when users delete a query result
- when users delete all results
- when users delete the query in the editor

_PS: The diff may look huge, but actually, we have the same changes applied on multiple places + copy/paste of a big component._

# How to test it

1. Go to Vector Search page - `http://localhost:8080/{databaseId}/vector-search`
2. Enter a query in the main editor and hit the "**Run**" button (dispatches `SEARCH_COMMAND_SUBMITTED` event)
3. Click on the "**Copy**" button when you hover over a result of a query (dispatches `SEARCH_COMMAND_COPIED` event)
4. Click on the "**Run**" button on a result of a query, to re-run it (dispatches `SEARCH_COMMAND_RUN_AGAIN` event)
6. Click on the "**Toggle Full-Screen**" button on a result of a query (dispatches `SEARCH_RESULTS_IN_FULL_SCREEN` event)
7. Click on the "**Expand**" button on a result of a query to see the command results (dispatches `SEARCH_RESULTS_COLLAPSED` event)
8. Click on the "**Collapse**" button on a result of a query to hide the command results (dispatches `SEARCH_RESULTS_EXPANDED` event)
9. Click on the "**Delete**" button on a result of a query to remove it (dispatches `SEARCH_CLEAR_RESULT_CLICKED` event)
10. Click on the "**Clear Results**" button to remove all commands (dispatches `SEARCH_CLEAR_ALL_RESULTS_CLICKED` event)
11. Click on the "**Clear**" button to remove the current query from the editor (dispatches `SEARCH_CLEAR_EDITOR_CLICKED` event)

https://github.com/user-attachments/assets/2812440a-79e2-419b-a7f1-1846b7177474

## Verify Telemetry events
We have the new events for the Vector Search page collected successfully.

_Note: Telemetry is collected only if the user decides to opt-in to that.
You control it from **Settings** page -> **Privacy** section -> **Usage Data** toggle._

image

image

image

image

image

image

image

image

image

Notes: Since the introduced changes reflect the Workbench telemetry as well, I have checked that we still emit properly the following events:
- `WORKBENCH_RESULTS_COLLAPSED`
- `WORKBENCH_RESULTS_EXPANDED`
- `WORKBENCH_COMMAND_COPIED`
- `WORKBENCH_CLEAR_RESULT_CLICKED`


redisinsight/ui/src/pages/vector-search/components/commands-view/CommandsView/CommandsView.tsx
```diff
@@ -76,6 +79,16 @@ const CommandsView = (props: Props) => {
     }
   }
 
+  const collectTelemetryQueryReRun = (query: string) => {
+    sendEventTelemetry({
+      event: TelemetryEvent.SEARCH_COMMAND_RUN_AGAIN,
+      eventData: {
+        databaseId: instanceId,
```
KrumTy: this function already takes a parameter, add another one - instanceId and you can remove it's definition from the component body

redisinsight/ui/src/pages/vector-search/components/query-card-header/QueryCardHeader.tsx
```diff
@@ -0,0 +1,577 @@
...
+import { RiSelect } from 'uiSrc/components/base/forms/select/RiSelect'
+import QueryCardTooltip from 'uiSrc/components/query/query-card/QueryCardTooltip'
+
+import styles from './styles.module.scss'
+
+/**
+ * Caution: This is a copy of the QueryCardHeader component from the Workbench module.
+ *
+ * We're really sorry to do that at this point, but this family fo components should be refactored to be reusable.
+ * At this point, it requires a lot of effort to make it reusable, so we decided to copy the component as is, and apply the necessary changes inline.
```
KrumTy: this adds to the tech debt. What are the changes the original component would require to make it reusable?

Diff checking both of them I currently see a telemetry param change and that's all, so what else is going to be changed?

redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
```diff
@@ -51,6 +53,22 @@ export const VectorSearchQuery = () => {
...
+    onSubmit()
+    collectTelemetryQueryRun()
+  }
+
+  const collectTelemetryQueryRun = () => {
+    sendEventTelemetry({
+      event: TelemetryEvent.SEARCH_COMMAND_SUBMITTED,
+      eventData: {
+        databaseId: instanceId,
+        commands: [query],
```
KrumTy: again, absolutely not a fan of declaring such functions inside the component, especially since they only require 1-2 params. This adds unnecessary unnecessary complexity - you open the component body and half the code is related to a telemetry callbacks

seeing how much space these telemetry functions take I'd even suggest extracting them in a separate telemetry.ts file or something like that

redisinsight/ui/src/pages/vector-search/components/query-card-header/QueryCardHeader.tsx
valkirilov: Yes, I agree. It was discussed with @pawelangelow to proceed this way for now. Check the Notes section in the description of the pull request for more information.

pawelangelow: Yeah, let's copy it for now. We either slow down to do it properly or introduce a tech debt - no "good" scenario :(

redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
pawelangelow: Well, they won't disappear completely as the handlers should live here:
```
const handleClearResults = () => {
  onAllQueriesDelete()
  sendEventTelemetrySomething()
}
...
and the remaining 3
```

redisinsight/ui/src/pages/vector-search/components/commands-view/CommandsView/CommandsView.tsx
valkirilov: Done. The telemetry collection is exported as util, outside of the component.

redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
valkirilov: Both of you are correct. We still have the base handlers as part of the component, because we need them.
But I managed to export the telemetry collection logic into a helper util, outside of the component.

redisinsight/ui/src/pages/vector-search/components/query-card-header/QueryCardHeader.tsx
valkirilov: Ok, I went for a simplified approach, and I think it worked:
- removed the copy/paste `QueryCardHeader` component https://github.com/redis/RedisInsight/pull/4811/commits/d6587d2dbced0ee661905d57a471f11760899dd3
- introduced a simple context that for now will keep information where the component is rendered - Workbench page or Vector Search page https://github.com/redis/RedisInsight/pull/4811/commits/49f5beeca12779fbf5af264d264969ee241eb0f0
- applied changes to emit telemetry conditionally, based on the state from the context https://github.com/redis/RedisInsight/pull/4811/commits/09b8e86e4c4d6aeb82eb5960c804f152eb85df46

redisinsight/ui/src/components/query/context/view-mode.context.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// Provider component
+export const ViewModeContextProvider: React.FC<
+  ViewModeContextProviderProps
+> = ({
+  children,
+  initialViewMode = ViewMode.Workbench, // Default to Workbench if not provided
+}) => {
+  const [viewMode, setViewMode] = useState<ViewMode>(initialViewMode)
```
KrumTy: is the setViewMode used anywhere (other than tests)? If not, you might as well drop the local state and use the enum value passed to the provider

redisinsight/ui/src/components/query/query-card/QueryCardHeader/QueryCardHeader.tsx
```diff
@@ -214,12 +227,16 @@ const QueryCardHeader = (props: Props) => {
...
+      const telemetryEvent =
+        viewMode === ViewMode.Workbench
+          ? isOpen
+            ? TelemetryEvent.WORKBENCH_RESULTS_COLLAPSED
+            : TelemetryEvent.WORKBENCH_RESULTS_EXPANDED
+          : isOpen
+            ? TelemetryEvent.SEARCH_RESULTS_COLLAPSED
+            : TelemetryEvent.SEARCH_RESULTS_EXPANDED
+
+      sendEvent(telemetryEvent, query)
```
pd-redis, line 239: this starts to look ugly

redisinsight/ui/src/components/query/context/view-mode.context.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const context = useContext(ViewModeContext)
+
+  if (!context) {
+    throw new Error(
+      'useViewModeContext must be used within a ViewModeContextProvider',
+    )
+  }
+
+  return context
+}
```
KrumTy:
```suggestion
export const useViewModeContext = () => useContext(ViewModeContext)
```

I think that not throwing an error here and leave it to the hook's consumer to decide what to do would be better. Using a default state that implements the interface instead of undefined would help too.

redisinsight/ui/src/components/query/context/view-mode.context.tsx
```diff
@@ -0,0 +1,51 @@
...
+  children: ReactNode
+  initialViewMode?: ViewMode // Optional prop to set the initial view mode
+}
+
+// Provider component
+export const ViewModeContextProvider: React.FC<
+  ViewModeContextProviderProps
+> = ({
+  children,
+  initialViewMode = ViewMode.Workbench, // Default to Workbench if not provided
```
KrumTy: explicit over default is better in this case imo

redisinsight/ui/src/components/query/context/view-mode.context.tsx
pd-redis, line 1: So basically, this is a context that shares single value across components, right?
Is there any reason not to just add this to Redux store?

redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
pd-redis: And obviously, all the logic can be external to the component by incorporating a hook :)
But since this is mostly personal preference, I am saying this just for completenes and nothing else :)

redisinsight/ui/src/components/query/context/view-mode.context.tsx
KrumTy: it's used in different views now. It would not make sense to dispatch redux actions to update the store as you switch between Search and Workbench tabs

This context solution is the least path of resistance to reusing the QueryCard component. The other being prop drilling, which @valkirilov said was too hard (4-5 levels deep).

The pre-context solution was to copy the whole component and it's test and adjust 3-4 telemetry lines.

redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
valkirilov: Initially, I went for a hook, but then @KrumTy shared some concerns on the topic and advised going for a simpler helper function, and keeping it independent of the React lifecycle. I think it works for now.

redisinsight/ui/src/components/query/context/view-mode.context.tsx
valkirilov: I prefer to keep this default value, for backward compatibility and safety.

redisinsight/ui/src/components/query/context/view-mode.context.tsx
valkirilov: Agree, changes applied, and the state is removed. We'l rely simply on the provided enum.

redisinsight/ui/src/components/query/context/view-mode.context.tsx
KrumTy: this is a new provider, no backwards compatibility is needed as you are defining the interface atm
also, from what I've seen, you are setting it explicitly everywhere (as should be) already

redisinsight/ui/src/components/query/query-card/QueryCardHeader/QueryCardHeader.tsx
valkirilov: Yes, we all agree on that. In general, the component should be refactored so it can be reused properly on multiple places without having such tight business logic inside it. Unfortunately, it's not that simple, so that’s why we went for this quick and dirty fix with the conditional telemetry.

Regarding this piece of code, the best thing I can do is to split it over two conditional operators, and it will still be ugly. So after all, I rely on general refactoring to fix this.

redisinsight/ui/src/components/query/context/view-mode.context.tsx
valkirilov: Done. I'm used to this pattern, but if it's not adopted in this project, I'm ok. removed.

redisinsight/ui/src/components/query/context/view-mode.context.tsx
pd-redis: Its already merged, but I still fail to see how context "dispatch" makes more sense than redux dispatch.
If anything this adds more techincal overhead

PR #4835, opened 2025-08-11
RI-7303: Store create index command in Search results

pawelangelow:
When you create a new index, currently, we:
- insert data
- create the index

So far, the command that created the index was stored in `Workbench` results as it reused `Workbench` action, but it's better for the UX if the user sees the command in the `Search` commands history. 

A.k.a. this:

Screenshot 2025-08-11 at 17 56 19


redisinsight/ui/src/services/hooks/useExecuteQuery.ts
```diff
@@ -0,0 +1,36 @@
...
+        options.afterAll?.()
+        return result
+      } catch (e) {
+        options.onFail?.(e)
+        throw e
+      }
+    },
+    [],
+  )
+}
```
KrumTy: it doesn't make sense for this to be a hook. The callback it returns has no dependencies. Just have a method (`executeQuery`) instead.

PR #4841, opened 2025-08-12
Fe/feature/ri 7231 replace eui unify components style

pd-redis:
# Rearrange and Rename Base Components (RI-7231)

## Summary
This PR reorganizes and renames custom components in the base directory to improve consistency and maintainability. The implementation standardizes naming conventions and export patterns across the codebase.

## Changes
- Renamed components with consistent 'Ri' prefix (e.g., RiText, RiPopover, RiTooltip)
- Reorganized directory structure for better component categorization
- Updated exports from base components (layout, text, tooltip, shared, popover, link)
- Added GlobalStyle theme component for consistent theming and CSS variables
- Standardized styling approach using styled-components
- Updated component imports across the application to use the new structure
- Fixed tests to accommodate the renamed components

## Why
This change improves code organization and maintainability by:
- Establishing consistent naming conventions for custom components
- Creating a more intuitive directory structure
- Centralizing theme variables through GlobalStyle
- Making component imports more consistent throughout the application

## Testing
- Updated and fixed affected tests to work with the renamed components
- Verified that all components render correctly with the new structure


redisinsight/ui/src/components/base/theme/index.ts
```diff
@@ -1,37 +1,75 @@
...
-        space800: '6.4rem',
-      },
-    },
-  },
-}
+export const GlobalStyle = createGlobalStyle<{ theme: Theme }>`
+  :root {
+    // spacing
+    //2.5px/0.2rem
+    --size-xxs: ${({ theme }) => theme.core.space.space025};
```
dantovska, line 8: Nice! I think that in a separate effort we can remove all these "--size-xxs" variables and just replace the in the components we use theme with the theme ones. Components are not that much I think (list, horizontal rule, horizontal spacer, loading content, spacer and few more occurrences), and since you've mapped the values here it would be easy to just replace.

redisinsight/ui/src/components/inline-item-editor/InlineItemEditor.tsx
```diff
@@ -3,14 +3,16 @@ import cx from 'classnames'
 
 import { useTheme } from '@redis-ui/styles'
 
+import { RiPopover, RiTooltip } from 'uiBase/index'
```
dantovska, line 6: I see this uiBase/index is used here and few more places. Is there a reason why not just use uiBase/display instead?

dantovska:
I know this is draft and not ready for merge, but few small things I noticed breaking:
image
image

redisinsight/ui/src/pages/browser/BrowserPage.tsx
```diff
@@ -335,17 +335,20 @@ const BrowserPage = () => {
               borderRadius: `8px`,
             }}
           >
+            <UploadModal>
+              <div>test</div>
+            </UploadModal>
```
dantovska, line 340: and not to forget to remove this 😄

redisinsight/ui/src/components/base/theme/index.ts
pd-redis: Well, purely for brevity using `var(--size-s)` is far shorter to type than `${({ theme }) => theme.core.space.space025}` so I am not sure which one to leave in.

This is all open for discussion.

redisinsight/ui/src/components/inline-item-editor/InlineItemEditor.tsx
pd-redis: No reason, just that EVERYTHING is exported in `index` and this is one thing that we should decide on

redisinsight/desktop/vite.renderer.config.ts
```diff
@@ -27,6 +27,7 @@ export default defineConfig({
     alias: {
       uiSrc: path.resolve(__dirname, '../ui/src'),
       apiSrc: path.resolve(__dirname, '../api/src'),
+      uiBase: path.resolve(__dirname, '../ui/src/components/base'),
```
pawelangelow, line 30: Uh, nice! ❤️

redisinsight/ui/src/components/base/display/call-out/RiCallOut.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { Banner } from '@redis-ui/components'
+
+export type CallOutProps = Omit<React.ComponentProps<typeof Banner>, 'show'> & {
```
pawelangelow, line 4: Should we keep the props with the same name convention as components? A.k.a. `RiCallOutProps`?

redisinsight/ui/src/components/base/theme/index.ts
```diff
@@ -1,37 +1,75 @@
...
-        space400: '3.2rem',
-        space500: '4rem',
-        space550: '4.4rem',
-        space600: '4.8rem',
-        space800: '6.4rem',
-      },
-    },
-  },
-}
+export const GlobalStyle = createGlobalStyle<{ theme: Theme }>`
```
pawelangelow, line 4: Do we need something that resets all the browser defaults for margins, paddings, etc.. Something like https://www.npmjs.com/package/reset-css

redisinsight/ui/src/components/base/index.ts
```diff
@@ -1,9 +1,9 @@
...
 
-export * from './tooltip'
-export * from './popover'
-
-export { RiFilePicker } from './forms/file-picker/RiFilePicker'
+export * from './layout'
+export * from './display'
+export * from './forms'
+export * from './text'
+export * from './shared/WindowControlGroup'
```
KrumTy, line 9: so is this barrel file going to be used for accessing the base components?

redisinsight/ui/src/components/auto-refresh/AutoRefresh.tsx
```diff
@@ -1,6 +1,10 @@
 import React, { useEffect, useState } from 'react'
 import cx from 'classnames'
-import { ChevronDownIcon, RefreshIcon } from 'uiSrc/components/base/icons'
+import { ChevronDownIcon, RefreshIcon, RiIcon } from 'uiBase/icons'
+import { RiIconButton } from 'uiBase/forms'
+import { RiColorText } from 'uiBase/text'
+import { RiSwitchInput } from 'uiBase/inputs'
+import { RiPopover, RiTooltip } from 'uiBase/index'
```
KrumTy, line 7: barrel or no barrel imports; mix is a bit confusing imo

redisinsight/ui/src/components/base/display/image/RiImage.tsx
```diff
@@ -1,8 +1,6 @@
 import React from 'react'
 import { RiImageProps, StyledImage } from './image.styles'
 
-const RiImage = ({ $size, src, alt, ...rest }: RiImageProps) => (
+export const RiImage = ({ $size, src, alt, ...rest }: RiImageProps) => (
   <StyledImage src={src} alt={alt} $size={$size} {...rest} />
 )
-
-export default RiImage
```
KrumTy, line 8: default export + a single barrel file that does `export { default as RiImage } from ...` could resolve the suggested auto import issue

redisinsight/ui/src/components/base/display/link/UserProfileLink.tsx
```diff
@@ -12,7 +12,7 @@ export const UserProfileLink = styled(Link)`
   &:not(:last-child) {
     border-bottom: 1px solid
       ${({ theme }: { theme: ReturnType<typeof useTheme> }) =>
-    theme.color.gray400};
+        theme.color.gray400};
```
KrumTy, line 15: looks like you need to run format on the branch

redisinsight/ui/src/components/base/forms/RiFormField.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { ComponentProps } from 'react'
+import {
+  FormField as RedisFormField,
+  TooltipProvider,
+} from '@redis-ui/components'
+
+export type RedisFormFieldProps = ComponentProps<typeof RedisFormField> & {
+  infoIconProps?: any
```
KrumTy, line 8: can this have a type?

redisinsight/ui/src/pages/browser/components/bulk-actions/BulkActionSummary/BulkActionSummary.tsx
```diff
@@ -16,10 +16,10 @@ export interface Props {
   'data-testid': string
 }
 
-const SummaryContainer = styled(Row)`
+const SummaryContainer = styled(RiRow)`
   padding-top: 18px;
 `
```
pawelangelow, line 21: I'm just brainstorming here, but what if we introduce https://www.npmjs.com/package/styled-system ?

What is the problem that bugs me: We'll introduce (and we already did) a lot of components with small styling fixes, causing a lot of boilerplate:

```
const RiRow = ... // this is OK

// This is necessary because we can't use <RiRow pt="18px" />
const SummaryContainer = styled(RiRow)`
  padding-top: 18px;
`

// Our end goal would be to remove the hardcoded pixels
<RiRow pt="small" /> or <RiRow pt={3} />
```

So what do you think?

Ofc, this is dead project, so as styled-components, but I guess it will help us if we migrate to something similar to tailwindcss in the future.

redisinsight/ui/src/components/base/forms/buttons/RiPrimaryButton.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+import { BaseButton } from './Button'
+import { ButtonProps } from './button.styles'
+
+export const RiPrimaryButton = (props: ButtonProps) => (
+  <BaseButton {...props} variant="primary" />
+)
```
KrumTy, line 7: see my comment in RiDestructiveButton.tsx

redisinsight/ui/src/components/base/forms/buttons/RiDestructiveButton.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+import { ButtonProps } from './button.styles'
+import { BaseButton } from './Button'
+
+export const RiDestructiveButton = (props: ButtonProps) => (
+  <BaseButton {...props} variant="destructive" />
+)
```
KrumTy, line 7: what is the reason behind these button wrappers? Is it to prevent the consumer from passing variant prop?
If so, it might be better to have a single RiButton wrapper on our end with allowed variants and their implementations

redisinsight/ui/src/components/base/display/call-out/RiCallOut.tsx
pd-redis: It is probably not a bad idea!

redisinsight/ui/src/components/base/theme/index.ts
pd-redis: I think redis ui styles have something like that - https://github.com/redis/RedisInsight/blob/75b67991068dad0def700fc6d6eb12865ce65c61/redisinsight/ui/src/contexts/themeContext.tsx#L8

pd-redis: But never actually looked at the css, so ti can be something entirely different

redisinsight/ui/src/pages/browser/components/bulk-actions/BulkActionSummary/BulkActionSummary.tsx
pd-redis: The issue I have with this is where we stop with the props.
Do we do margins, font styles, colors etc too?
There will always be edge cases. I would love for us to have a lego style design system, where you have just predefined components and props and only ever use that ( i think this is @ArtemHoruzhenko's dream) 
But the design and dev effort required makes me think it is not going to happen.

But I am very much open to suggestions. Abstracting the components the way we did means, that changing them in one place will be reflected everywhere so that is already a small win

redisinsight/ui/src/components/base/index.ts
pd-redis: It is possible to access them all from here, unfortunatelly  it has to be like `from "@uiBase/index"` and honestly I still like it more like `import {RiRow, RiTable} from "@uiBase/layout"` it makes it grouped more contextual.
BUT - if most of the team prefers base barrel file, then we'll use it

redisinsight/ui/src/components/auto-refresh/AutoRefresh.tsx
pd-redis: All imports here are barrel imports, the only confising thing for me is why inputs are not part of the forms subtree, but the reallity is that I don't care that much where are things stored

redisinsight/ui/src/components/base/display/image/RiImage.tsx
pd-redis: Or just `export const ...` 
why introduce extra files that don't do anything meaningful? Single file per single export ....

redisinsight/ui/src/components/base/display/link/UserProfileLink.tsx
pd-redis: format runs on safe, but will do it again

redisinsight/ui/src/components/base/forms/RiFormField.tsx
pd-redis: I am not sure, I don't think I have implemented this one

redisinsight/ui/src/components/base/forms/buttons/RiDestructiveButton.tsx
pd-redis: Nope, it is to avoid variants.
To me this is more semantic than `<RIButton variant="danger"`

redisinsight/ui/src/components/base/forms/buttons/RiPrimaryButton.tsx
pd-redis: Again, a design system is the one that has defined looks for primary, secondary, success, failure, danger, info states and typing RiPri will give you this button instead a button where you have to then add variant, then select the variant.
This is both more semantic and more ergonomic and there are 3-4 button variants, I don't think you will sturggle that much remembering them.
And finally, there is export for the BaseButton and if you hate that much semantic buttons feel free to use that one.

redisinsight/ui/src/components/base/display/call-out/RiCallOut.tsx
valkirilov: +1

redisinsight/ui/src/components/base/display/link/RiLink.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+import { LinkProps } from '@redis-ui/components'
+import { StyledLink } from './link.styles'
+
+export const RiLink = ({ color, ...props }: LinkProps) => (
```
valkirilov, line 5: Follow up on the following suggestion https://github.com/redis/RedisInsight/pull/4841/files#r2292865447

Looking at this example, here we don't alter the base interface in any way, but since we're implementing RiComponents as a middle layer between RedisInsight and the outer world, does it make sense to define our own interfaces for each component we wrap? 

```
export interface RiLinkProps extends LinkProps {}

export const RiLink = ({ color, ...props }: RiLinkProps) => (
```

This way, our codebase will be bound to `RiLinkProps` and in case we need to switch the components library (once again), we'll do it in a single place (in theory).

Yes, it will look a bit meaningless, but I'm thinking out loud.

redisinsight/ui/src/components/base/display/loader/RiLoader.tsx
```diff
@@ -0,0 +1,30 @@
...
+      return space.space300
+    default:
+      return space.space100
+  }
+}
+
+export const RiLoader = ({ size, ...rest }: RedisLoaderProps) => {
+  const theme = useTheme()
+  const { space } = theme.core
+  const sizeInPx = size ? convertSizeToPx(size, space) : space.space100
```
valkirilov, line 28: nit: It's not a big deal in this case, but we break the interface here. Redis UI `Loader` component can accept size in pixels ([docs](https://redislabsdev.github.io/redis-ui/?path=/docs/components-loader--docs)), but we do limit it to a set of predefined sizes. 

I'm okay with it that it should be the default behavior, and after all, it's a simple loader, and most probably we'll never need it to be 123.5px big, but just leaving this as a note, the original component does support it.

redisinsight/ui/src/components/base/display/tooltip/HoverContent.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { RiTitle } from 'uiBase/text'
+
+interface RiTooltipContentProps {
+  title?: React.ReactNode
+  content: React.ReactNode
+}
+
+export const HoverContent = ({ title, content }: RiTooltipContentProps) => (
+  <RiCol>
+    {title && <RiTitle size="S">{title}</RiTitle>}
```
valkirilov, line 13: Note: I see we prefer using mostly string literal union types (`size: "S" | "M" | "L"`), and I see it to be inconsistent across the codebase and the different components. 

Here, in this component, the size is with a capital "S". In other components, we rely on lowercase strings (without a union type). In other components, we do lowercase everything and then apply some custom handling to transform the size into rem/px.

And while it's not a big deal, I feel it will look way more consistent if we use enums. Yes, it's my own personal preference, maybe, but we'll no longer care whether it's a capital or lowercase "S", and even if it's type checked or we entered something invalid. We simply put `Size.M` and that's all.

_PS: This is not related in any way to the changes introduced in this pull request, it's a general note._

redisinsight/ui/src/components/base/forms/fieldset/RiFormFieldset.styles.ts
```diff
@@ -0,0 +1,24 @@
...
+  padding: 0;
+  min-width: 0;
+`
+
+export interface StyledLegendProps extends HTMLAttributes<HTMLLegendElement> {
+  display?: 'visible' | 'hidden'
+}
+
+export const StyledLegend = styled.legend<StyledLegendProps>`
+  ${({ theme }: { theme: Theme } & StyledLegendProps) => css`
```
valkirilov, line 20: Just curious, does this nested `css` block bring any additional value, compared the the following construction?
 
```suggestion
margin-bottom: ${({ theme }) => theme.core.space.space100};
```

redisinsight/ui/src/components/base/theme/types.ts
```diff
@@ -7,3 +7,11 @@ export type CommonProps = {
 }
 
 export type Theme = ReturnType<typeof useTheme>
+
+const ASC = 'asc'
+const DESC = 'desc'
+export type Direction = typeof ASC | typeof DESC
```
valkirilov, line 13: nit: I've already mentioned this in other [comment](https://github.com/redis/RedisInsight/pull/4841#discussion_r2297271411), but this is a great example in which using enum will make it simpler. 

And yes, it's a subjective statement, I'm aware of that "more readable" is not a strong argument 😀

```
export enum Direction {
  Asc = 'asc',
  Desc = 'desc'
}
```

valkirilov (commented):
Tremendous effort, nice job! I really like how things are getting more organized now!

Looking at this changeset, I think it could be helpful if we wrote down some of these internal conventions so we can follow them consistently in the future. Maybe something simple like an inline README in markdown that acts as a mini style guide.

It doesn’t need to be complex, just a quick reference with things like:
- prefixing components with `RiComponent`
- how to structure and export files
- how to handle interfaces
- and maybe generic patterns we want to follow

That way, it's easy to get started, or to check later on how we prefer to do things.

Additionally, I know we've raised this a few times, but I think it would really help to have a place where we can see all the components we have and how to use them. Having an internal Storybook might be too much effort to maintain (though it would be a great addition), but even just bringing back the playground you prepared would be a solid starting point.

This is, of course, an additional task, complementary to this refactoring. I'm just mentioning it as a reminder so we can decide which path we want to follow and plan accordingly.

redisinsight/ui/src/components/base/display/loader/RiLoader.tsx
pd-redis: Well, this was Didi's decisision, and tend to agree. If you have system that allows for any size and any color, you don't have a system at all

redisinsight/ui/src/components/base/display/tooltip/HoverContent.tsx
pd-redis: the inconsistency in casing is because EUI uses lowercase, redis-ui upper case - hence the mix
And with or without enums, you will have to transofrm the logic to px, em or whatever - if you want to use theme values

redisinsight/ui/src/components/base/forms/fieldset/RiFormFieldset.styles.ts
pd-redis: No, it does not, since it is already in the context of the theme, but this is part of the implementation of the component, which I did not change with this PR

redisinsight/ui/src/components/base/display/loader/RiLoader.tsx
valkirilov: Yes, that's correct, all features we implement should follow the design system (and rely on this set of sizes), but here, I don't say that we should leave it without any predefined defaults. My point was that we limit the behavior of the 3rd party component, because it already supports this out of the box. 

Anyway, this is the Loader component, so we might never end up in such a situation.

pd-redis:
not relevant

pd-redis:
not relevant

PR #4844, opened 2025-08-12
RI-7242: Change the tooltip title size from S to XS

dantovska:
Tooltip size S seems a little too big for the content, so changing it to XS.

Before:
image-20250812-083718

After:
image

### DIFF
--- a/redisinsight/ui/src/components/base/tooltip/HoverContent.tsx
+++ b/redisinsight/ui/src/components/base/tooltip/HoverContent.tsx
@@ -9,7 +9,7 @@ interface RiTooltipContentProps {
 }
 
 export const HoverContent = ({ title, content }: RiTooltipContentProps) => (
-  <Col>
+  <Col gap="s">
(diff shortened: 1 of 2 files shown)


redisinsight/ui/src/components/base/tooltip/HoverContent.tsx
```diff
@@ -10,7 +10,7 @@ interface RiTooltipContentProps {
 
 export const HoverContent = ({ title, content }: RiTooltipContentProps) => (
   <Col>
```
pd-redis: I would add some `gap` to the Col, it seems a bit too tight in the tooltip

dantovska: Done. Attached the new screenshot. I intentionally added a small gap, because if I add a medium sized one (or bigger), then the content items also seem too tight to each other and when it is one item, it seems far from the title 😄

PR #4850, opened 2025-08-13
Fix e2e tests trigger

KrumTy:
## Description

Currently, when a PR is approved for the first time, it is labeled as e2e-approved. And when a e2e-approved label is added, it should trigger e2e tests. 

In reality what happens is when a PR is approved, the e2e-approved label is added, but the workflow for tests does not trigger. However if the PR is labeled manually via Github ui as e2e-approved, the tests do trigger.

~~Potential solution could be adding github secret to the label action to mimic authorized labeling~~

## Solution summary
This PR reworks how E2E tests are triggered:
- Replace label-based triggering with approval-based triggering and deduplication
- Add a reusable workflow for “approve + dedupe by PR head SHA” logic
- Update E2E workflow to use the reusable workflow and only run heavy jobs on Approved reviews of unique commit sha

E2E tests run only after a PR is approved (or when manually dispatched), per on commit sha
- No runs on commit/push
- Duplicate approvals/reviews for the same commit do not trigger e2e runs: the new run waits for or mirrors the previous run’s result. It also logs the link to the run it copied. 
- When a new commit is pushed, the head SHA changes, so the next approval will trigger a fresh E2E run


pawelangelow (dismissed):

pawelangelow (approved):

pd-redis (approved):

.github/workflows/tests-e2e.yml
```diff
@@ -11,71 +11,80 @@ on:
...
     secrets: inherit
     with:
+      target: build_linux_appimage_x64
       debug: ${{ inputs.debug || false }}
 
-  tests-e2e-playwright:
-    needs: build-docker
-    uses: ./.github/workflows/tests-e2e-playwright.yml
+  builds-complete:
+    needs: [build-docker, build-appimage]
```
KrumTy, line 55: intermediate job to wait for both builds to complete (~5mins) to avoid running tests when the other build has failed and the whole workflow will be marked as failed in the end anyway

KrumTy: test

PR #4855, opened 2025-08-13
RI-6927: Rename some provider and telemetry event names

dantovska:
- RE_CLOUD -> REDIS_CLOUD
- RE_CLUSTER -> REDIS_SOFTWARE


redisinsight/api/src/constants/telemetry-events.ts
```diff
@@ -21,12 +21,12 @@ export enum TelemetryEvents {
   DatabaseImportPartiallySucceeded = 'CONFIG_DATABASES_REDIS_IMPORT_PARTIALLY_SUCCEEDED',
 
   // Events for autodiscovery flows
-  REClusterDiscoverySucceed = 'CONFIG_DATABASES_RE_CLUSTER_AUTODISCOVERY_SUCCEEDED',
-  REClusterDiscoveryFailed = 'CONFIG_DATABASES_RE_CLUSTER_AUTODISCOVERY_FAILED',
-  RECloudSubscriptionsDiscoverySucceed = 'CONFIG_DATABASES_RE_CLOUD_AUTODISCOVERY_SUBSCRIPTIONS_SUCCEEDED',
-  RECloudSubscriptionsDiscoveryFailed = 'CONFIG_DATABASES_RE_CLOUD_AUTODISCOVERY_SUBSCRIPTIONS_FAILED',
-  RECloudDatabasesDiscoverySucceed = 'CONFIG_DATABASES_RE_CLOUD_AUTODISCOVERY_DATABASES_SUCCEEDED',
-  RECloudDatabasesDiscoveryFailed = 'CONFIG_DATABASES_RE_CLOUD_AUTODISCOVERY_DATABASES_FAILED',
+  RedisSoftwareDiscoverySucceed = 'CONFIG_DATABASES_REDIS_SOFTWARE_AUTODISCOVERY_SUCCEEDED',
```
valkirilov, line 24: Question: I might be missing something here, but why change both the constant names and their values? Usually, we use constants to hide the magic strings so we don’t need to update the whole codebase later. Or in this case, were the names actually wrong/misleading and needed to be updated too?

PS: Since you already made the effort to migrate everything, don't make any further changes, I'm just curious 🙂

valkirilov (dismissed):

redisinsight/api/src/constants/telemetry-events.ts
dantovska: I get your point and it makes sense for constants name to stay in general, but maybe if the constant name was descriptive and general enough, but this "RECloud" part I think it is both not descriptive and will look like a leftover we wouldn't have cleaned, at least to me ...

redisinsight/api/migration/1755086732238-update-provider-names.ts
```diff
@@ -0,0 +1,29 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+            UPDATE database_instance
+            SET provider = CASE provider
+                WHEN 'RE_CLOUD' THEN 'REDIS_CLOUD'
+                WHEN 'RE_CLUSTER' THEN 'REDIS_SOFTWARE'
+                WHEN 'REDIS_ENTERPRISE' THEN 'OTHER_REDIS_MANAGED'
+                ELSE provider
+            END
+            WHERE provider IN ('RE_CLOUD', 'RE_CLUSTER');
```
ArtemHoruzhenko: `REDIS_ENTERPRISE` won't change to `OTHER_REDIS_MANAGED` since you filtered it out in `WHERE` section below

redisinsight/api/migration/1755086732238-update-provider-names.ts
```diff
@@ -0,0 +1,29 @@
...
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+            UPDATE database_instance
+            SET provider = CASE provider
+                WHEN 'REDIS_CLOUD' THEN 'RE_CLOUD'
+                WHEN 'REDIS_SOFTWARE' THEN 'RE_CLUSTER'
+                WHEN 'OTHER_REDIS_MANAGED' THEN 'REDIS_ENTERPRISE'
+                ELSE provider
+            END
+            WHERE provider IN ('REDIS_CLOUD', 'REDIS_SOFTWARE');
```
ArtemHoruzhenko: the same

ArtemHoruzhenko: I would ask you to verify migrations locally to make sure it works before merging to main

redisinsight/api/src/__mocks__/cloud-autodiscovery.ts
```diff
@@ -68,8 +68,8 @@ export const mockCloudAutodiscoveryService = jest.fn(() => ({
...
 
 export const mockCloudAutodiscoveryAnalytics = jest.fn(() => ({
-  sendGetRECloudSubsSucceedEvent: jest.fn(),
-  sendGetRECloudSubsFailedEvent: jest.fn(),
-  sendGetRECloudDbsSucceedEvent: jest.fn(),
-  sendGetRECloudDbsFailedEvent: jest.fn(),
+  sendGetRedisCloudSubsSucceedEvent: jest.fn(),
+  sendGetRedisCloudSubsFailedEvent: jest.fn(),
+  sendGetRedisCloudDbsSucceedEvent: jest.fn(),
+  sendGetRedisCloudDbsFailedEvent: jest.fn(),
```
ArtemHoruzhenko, line 74: why do we need to change method names and constants? Why not change values only?

ArtemHoruzhenko: ignore this. found discussion regarding this below.

dantovska: @valkirilov @ArtemHoruzhenko If you both agree and think this is better, I'm fine reverting it.

redisinsight/api/migration/1755086732238-update-provider-names.ts
dantovska: I verified it for the other two types, missed that one... tnx!

redisinsight/api/src/__mocks__/cloud-autodiscovery.ts
valkirilov: Since you already did the effort to rename them, I don't think you should revert the names (at least for the constants - the names of the telemetry events)

ArtemHoruzhenko (approved):

PR #4868, opened 2025-08-15
RI-7382: Fix non editable username and database alias fields

dantovska:
The database connection settings fields are not editable due to being tied to onChange event instead of onChangeCapture. 
Added unit tests to avoid such issues in the future.

### DIFF
--- a/redisinsight/ui/src/pages/home/components/form/DatabaseForm.spec.tsx
+++ b/redisinsight/ui/src/pages/home/components/form/DatabaseForm.spec.tsx
@@ -1,9 +1,9 @@
 import React from 'react'
 import { instance, mock } from 'ts-mockito'
 import { useFormik } from 'formik'
-import { render, screen, fireEvent, act } from 'uiSrc/utils/test-utils'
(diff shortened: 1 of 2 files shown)


redisinsight/ui/src/pages/home/components/form/DatabaseForm.spec.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+jest.mock('uiSrc/slices/app/info', () => ({
+  appInfoSelector: jest.fn().mockReturnValue({
+    server: {
+      buildType: 'DOCKER_ON_PREMISE',
+    },
+  }),
+}))
+
+const mockFormikValues: DbConnectionInfo = {
```
valkirilov: Note: This is a perfect candidate for a reusable factory:
- it will make the test suite simpler, hiding the unnecessary business logic related to the DB Connection Info object structure
- it will provide a simple and reusable way to apply the same concept to various unit tests.
- in case we change the interface of the object, it will be automatically inherited when we update the factory, without the need to manually update it here in the test suite.

Examples:

PR #4870, opened 2025-08-15
RI-7365: Show no indexes message

pawelangelow:
# When no indexes

Screenshot 2025-08-15 at 16 21 43

# When there are indexes

Screenshot 2025-08-15 at 16 21 56

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/hooks/useStartWizard.spec.ts
+++ b/redisinsight/ui/src/pages/vector-search/hooks/useStartWizard.spec.ts
@@ -68,18 +68,4 @@ describe('useStartWizard', () => {
       Pages.vectorSearchCreateIndex(undefined as any),
     )
   })
-
-  it('should call history.push with correct path format', () => {
-    const instanceId = 'test-instance-456'
-    useParamsMock.mockImplementation(() => ({ instanceId }))
-
-    const { result } = renderHook(() => useStartWizard())
-
-    act(() => {
-      result.current()
-    })
-
-    const expectedPath = `/${instanceId}/vector-search/create-index`
-    expect(mockPush).toHaveBeenCalledWith(expectedPath)
-  })
 })


redisinsight/ui/src/pages/vector-search/manage-indexes/ManageIndexesList.tsx
```diff
@@ -5,14 +5,18 @@ import { bufferToString } from 'uiSrc/utils'
...
 
 export const ManageIndexesList = () => {
   const { data, loading } = useRedisearchListData()
+  const hasIndexes = !!data?.length
 
   return (
     <StyledManageIndexesListAction data-testid="manage-indexes-list">
       {loading && <Loader data-testid="manage-indexes-list--loader" />}
 
+      {!loading && !hasIndexes && <NoIndexesMessage />}
```
KrumTy, line 18: maybe worth adding a test for the empty list

redisinsight/ui/src/pages/vector-search/query/StartWizardButton.tsx
```diff
@@ -1,23 +1,17 @@
-import React, { useCallback } from 'react'
-import { useHistory, useParams } from 'react-router-dom'
+import React from 'react'
 import { CallOut } from 'uiSrc/components/base/display/call-out/CallOut'
-import { Pages } from 'uiSrc/constants'
+import useStartWizard from '../hooks/useStartWizard'
```
KrumTy, line 3: you can probably simplify this component's tests by mocking '../hooks/useStartWizard' as opposed to it's internals ('react-router-dom') since you've already tested them; not mandatory ofc

redisinsight/ui/src/pages/vector-search/hooks/useStartWizard.ts
```diff
@@ -0,0 +1,18 @@
...
+  const history = useHistory()
+  const { instanceId } = useParams<{ instanceId: string }>()
+
+  const start = useCallback(() => {
+    history.push(Pages.vectorSearchCreateIndex(instanceId))
+  }, [history, instanceId])
+
+  return {
+    start,
+  }
```
KrumTy: if there's no plan on adding more attributes to the return object, it might be better to return just the function

valkirilov: Is there actually any performance difference between returning a function vs. an object here? To me it looks identical.

The only difference I really see is API style. I know it's a matter of personal preference, but even though this hook is simple enough that it may never be extended, I usually prefer writing hooks in a way that leaves room for easy extension later without needing to touch the consumers.

redisinsight/ui/src/pages/vector-search/hooks/useStartWizard.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    const instanceId = 'test-instance-456'
+    useParamsMock.mockImplementation(() => ({ instanceId }))
+
+    const { result } = renderHook(() => useStartWizard())
+
+    act(() => {
+      result.current.start()
+    })
+
+    const expectedPath = `/${instanceId}/vector-search/create-index`
```
valkirilov: nit: I would put this assertion into the first test case. To me, it looks like we're testing the same thing, so it doesn’t break the atomic behavior of the tests and their simplicity if we simply combine multiple assertions under the same test.

PR #4872, opened 2025-08-19
RI-7328: Fix capabilities alignment by removing additional component and styles

dantovska:
Before:
image

After:
image


redisinsight/ui/src/pages/home/components/form/DbInfo.tsx
```diff
@@ -75,7 +75,7 @@ const DbInfo = (props: Props) => {
       {!isFromCloud && (
         <ListGroupItem
           label={
-            <Text color="subdued" size="s">
+            <Text color="default" size="s">
```
valkirilov, line 78: Question: Looking at this change, the first thing that comes to mind is - should we pass it explicitly, since it's `default`? 

But when I checked the definition of the `Text` component, I didn't see any defaults for the different input props. So I'm wondering if you prefer everything to be explicitly provided, as a pattern?

CC: @pd-redis

KrumTy (approved):

redisinsight/ui/src/pages/home/components/form/DbInfo.tsx
pd-redis: @valkirilov well the text component does not have any specific color. In this case, I would remove the `color`

pd-redis (commented):
@dantovska if you remove `color` prop where it is `default` would it work?

dantovska:
> @dantovska if you remove `color` prop where it is `default` would it work?

This was the first thing I tried, and it does not work when the theme is dark - the default value for color becomes inherit and when the theme is dark, the text is also dark and not readable. This is why I went with "default". cc: @pd-redis

redisinsight/ui/src/pages/home/components/form/DbInfo.tsx
dantovska: This was the first thing I tried, and it does not work when the theme is dark - the default value for color becomes inherit and when the theme is dark, the text is also dark and not readable. This is why I went with "default". cc: @pd-redis

PR #4875, opened 2025-08-19
RI-7347: Tips section slightly broken: links, voting...

dantovska:
| Before | After | 
|----------|----------|
| image   | image   | 
| image   | image  | 
| image   | image   |
| image   | image  |

### How to test:
1. Open an instance (go to Browser Page).
2. Click on the Redis Insights bulb icon to view Tutorials and Tips.

image

3. Click on the Tips.
4. Open one of them. If none exist, click on the Run Database Analysis below. 

image

5. Expand one of the accordions.

### DIFF
--- a/redisinsight/ui/src/components/recommendation/recommendation-voting/components/vote-option/VoteOption.tsx
+++ b/redisinsight/ui/src/components/recommendation/recommendation-voting/components/vote-option/VoteOption.tsx
@@ -14,7 +14,7 @@ import { Nullable } from 'uiSrc/utils'
 
 import { Col, FlexItem, Row } from 'uiSrc/components/base/layout/flex'
 import { Text } from 'uiSrc/components/base/text'
-import { CancelSlimIcon } from 'uiSrc/components/base/icons'
+import { CancelSlimIcon, Icon } from 'uiSrc/components/base/icons'
 import { IconButton } from 'uiSrc/components/base/forms/buttons'
 import { Link } from 'uiSrc/components/base/link/Link'
 import { RiPopover, RiTooltip } from 'uiSrc/components/base'
@@ -47,6 +47,12 @@ const GitHubLink = styled(Link)`
     gap: 4px;
     align-items: center;
   }
+`
+
+const VotingIconButton = styled(IconButton)`
+  width: 28px !important;
+  height: 28px !important;
+  border-radius: 50%;
 `
 
 export interface Props {
@@ -130,7 +136,7 @@ const VoteOption = (props: Props) => {
           position="bottom"
           data-testid={`${voteOption}-vote-tooltip`}
         >
-          <IconButton
+          <VotingIconButton
             disabled={!isAnalyticsEnable}
             icon={iconType[voteOption] ?? 'LikeIcon'}
             className={cx('vote__btn', { selected: vote === voteOption })}
--- a/redisinsight/ui/src/components/recommendation/recommendation-voting/styles.module.scss
+++ b/redisinsight/ui/src/components/recommendation/recommendation-voting/styles.module.scss
@@ -3,19 +3,6 @@
     .voteContent {
       margin-left: 10px;
 
-      .vote__btn {
-        width: 28px !important;
-        height: 28px !important;
-        border-radius: 50%;
-
-        &:hover,
-        &:focus,
-        &.selected {
-          transform: none;
-          background-color: var(--separatorColor);
-        }
-      }
-
       svg {
         width: 34px;
         height: 34px;
--- a/redisinsight/ui/src/components/side-panels/panels/live-time-recommendations/components/recommendation/styles.module.scss
+++ b/redisinsight/ui/src/components/side-panels/panels/live-time-recommendations/components/recommendation/styles.module.scss
@@ -88,21 +88,6 @@
       font-size: inherit !important;
     }
   }
-
-  .link {
-    color: var(--externalLinkColor) !important;
-  }
-
-  :global {
-    .vote__btn {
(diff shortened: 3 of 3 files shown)


redisinsight/ui/src/styles/themes/light_theme/_theme_color.scss
```diff
@@ -156,7 +156,7 @@ $recommendationBorderColor: #3953c3;
 
 
 // Recommendations
-$recommendationsBgColor: #FFFFFF;
+$recommendationsBgColor: #F0F0F0;
```
valkirilov: I noticed that this color is being used in multiple places in the codebase. Is it safe to change it here, globally?

pd-redis:
Are these changes - colors, fonts - because the design requires them?
Also, we are trying to move away from scss and custom colors. I would look for matching or similar color in the theme and use it.

redisinsight/ui/src/styles/themes/light_theme/_theme_color.scss
dantovska: It is used on three places, two of them are in the recommendation component and I tested them - for which is the intention of changing this value. The third occurrence however, is in a class and this class is not used anywhere, so we're safe ^_^

dantovska:
> Are these changes - colors, fonts - because the design requires them? Also, we are trying to move away from scss and custom colors. I would look for matching or similar color in the theme and use it.

@pd-redis 
If you referring to the recommendationsBgColor, I'm changing it in one place as where it's used I want to set the same color - greyish and not entirely white. This color is actually fixing the background for "Is this useful?" wrapper. I think this is the best best solution for this problem. 
I'm not sure if we have discussed how to handle background styles for components though... maybe you are thinking of creating a styled wrapper component where we can just set a theme color instead of experimenting with the palette?

pd-redis:
> > Are these changes - colors, fonts - because the design requires them? Also, we are trying to move away from scss and custom colors. I would look for matching or similar color in the theme and use it.
> 
> @pd-redis If you referring to the recommendationsBgColor, I'm changing it in one place as where it's used I want to set the same color - greyish and not entirely white. This color is actually fixing the background for "Is this useful?" wrapper. I think this is the best best solution for this problem. I'm not sure if we have discussed how to handle background styles for components though... maybe you are thinking of creating a styled wrapper component where we can just set a theme color instead of experimenting with the palette?

We Should use styled wrapper and not scss for sure.
And use a theme color which is alredy in the theme and is relatively close to the one currently used

valkirilov (dismissed):

redisinsight/ui/src/components/recommendation/recommendation-copy-component/RecommendationCopyComponent.tsx
```diff
@@ -1,14 +1,46 @@
...
+  border-radius: 4px;
+  background-color: ${(props) =>
+    props.$isDbAnalysis
+      ? props.theme.semantic.color.background.neutral100
+      : props.theme.semantic.color.background.neutral400};
+`
+
+const StyledKeyName = styled(Text)`
+  padding: 4px;
+  font:
```
pd-redis, line 30: size XS for the text is 12.5px, size S is 14px, prefer using these styles over explicit css

redisinsight/ui/src/components/recommendation/recommendation-copy-component/RecommendationCopyComponent.tsx
```diff
@@ -1,14 +1,46 @@
...
+  border: 1px dashed ${({ theme }) => theme.semantic.color.border.neutral500};
+  border-radius: 4px;
+  background-color: ${(props) =>
+    props.$isDbAnalysis
+      ? props.theme.semantic.color.background.neutral100
+      : props.theme.semantic.color.background.neutral400};
+`
+
+const StyledKeyName = styled(Text)`
+  padding: 4px;
```
pd-redis: `${({ theme }) => theme.core.space.space050}` is probably better

redisinsight/ui/src/components/recommendation/recommendation-copy-component/RecommendationCopyComponent.tsx
```diff
@@ -1,14 +1,46 @@
...
+`
+
+const StyledText = styled(Text)`
+  font:
+    normal normal normal 13px/16px Graphik,
+    sans-serif !important;
+`
+
+const StyledIconButton = styled(IconButton)`
+  margin-right: 5px;
```
pd-redis: ${({ theme }) => theme.core.space.space050} is probably better

redisinsight/ui/src/components/recommendation/recommendation-voting/styles.module.scss
```diff
@@ -4,8 +4,8 @@
       margin-left: 10px;
 
       .vote__btn {
-        width: 34px;
-        height: 34px;
+        width: 28px !important;
```
pd-redis: If extracting component is an option, prefer that

valkirilov (approved):

redisinsight/ui/src/components/recommendation/recommendation-voting/components/vote-option/VoteOption.tsx
```diff
@@ -49,6 +49,12 @@ const GitHubLink = styled(Link)`
   }
 `
 
+const VotingIconButton = styled(IconButton)`
+  width: 28px !important;
```
pd-redis, line 53: It is quite likely, that this should work without !important, no need to change it, but have in mind

PR #4877, opened 2025-08-19
RI-7372: make "Manage indexes" a panel

KrumTy:
### Description

- changes Manage indexes from a flyout to a panel 
- Saved Queries / Manage indexes panels are mutually exclusive 
- some refactoring to the way events are logged for the two panels

| Before | After |
| --- | --- |
| Screenshot 2025-08-19 at 17 25 46 | Screenshot 2025-08-19 at 17 25 23 |


redisinsight/ui/src/pages/vector-search/hooks/useTelemetryMountEvent.ts
```diff
@@ -0,0 +1,34 @@
...
+      onUnmountEvent &&
+        sendEventTelemetry({
+          event: onUnmountEvent,
+          eventData: {
+            databaseId: instanceId,
+            ...eventData,
+          },
+        })
+    }
+  }, [])
```
KrumTy, line 33: no dependencies as it's explicitly meant to run once per mount as per usage

redisinsight/ui/src/pages/vector-search/hooks/useTelemetryMountEvent.ts
```diff
@@ -0,0 +1,34 @@
...
+import { sendEventTelemetry, TelemetryEvent } from 'uiSrc/telemetry'
+import { EventData } from 'uiSrc/telemetry/interfaces'
+
+export const useTelemetryMountEvent = (
+  onMountEvent: TelemetryEvent,
+  onUnmountEvent?: TelemetryEvent,
+  eventData?: EventData,
+) => {
+  const { instanceId } = useParams<{ instanceId: string }>()
+
```
KrumTy, line 13: this encapsulates some event tracking logic that was previously linked to various click events and split across components. 
The use case - when expanding/collapsing the right panel, an event should be logged

redisinsight/ui/src/pages/vector-search/manage-indexes/ManageIndexesScreen.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <VectorSearchScreenHeader>
+        <Title size="M" data-testid="title">
+          Manage indexes
+        </Title>
+      </VectorSearchScreenHeader>
+      <VectorSearchScreenFooter
+        data-testid="manage-indexes-screen-body"
+        grow={1}
+      >
+        <ManageIndexesList />
```
valkirilov, line 34: Now, when we moved the **Index Details** sections outside of the drawer, the table with the attributes details is not scaling properly when the **Manage Indexes** panel is wider.

It's caused by the way we made the table horizontally scrollable. According to the [docs of Redis UI table](https://github.com/redislabsdev/redis-ui/blob/master/docs/stories/Table/Table.stories.tsx#L111) component, we should wrap it inside a div with a fixed width, in order to make it scrollable.

That's why we have the following [dirty hack](https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/vector-search/manage-indexes/IndexAttributesList.styles.ts#L11) (_ignore the fact the width of the drawer is hardcoded as magic number, because there is no such variable_), but now we should come up with a different solution.

Image

redisinsight/ui/src/pages/vector-search/manage-indexes/ManageIndexesScreen.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { ManageIndexesList } from './ManageIndexesList'
+
+export const ManageIndexesScreen = () => {
+  useTelemetryMountEvent(
+    TelemetryEvent.SEARCH_MANAGE_INDEXES_DRAWER_OPENED,
+    TelemetryEvent.SEARCH_MANAGE_INDEXES_DRAWER_CLOSED,
+  )
+
+  return (
+    <VectorSearchScreenWrapper
```
valkirilov, line 21: Note: I'm not sure whether it's linked with your changes, or more to Redis UI components styling, but I've noticed that the colors of the borders are slightly off (and it's more noticeable on dark mode, but it's applicable to the light theme as well):
- **Query Results** panel (on the left) use one color, dark grey
- **Manage Indexes** panel uses a second color, a little bit brighter
- inside it, the **Section** component and the table inside it are using again the darker variant

Image


The same can be noticed for the "**Saved Queries**" panel as well.

Image


Maybe this should go for a discussion with the design team, in a separate ticket?

valkirilov (approved):

PR #4879, opened 2025-08-20
Ri-7333 infinite notifications

pd-redis:
Changed size of titles
Chnaged type of the toast from notice to informative
Clear all current Infinite messages when new ones arrive

### DIFF
--- a/redisinsight/ui/src/components/notifications/Notifications.tsx
+++ b/redisinsight/ui/src/components/notifications/Notifications.tsx
@@ -165,7 +165,7 @@ const Notifications = () => {
             dispatch(removeInfiniteNotification(id))
           },
         },
-        { variant: riToast.Variant.Notice, autoClose: ONE_HOUR },
+        { variant: riToast.Variant.Informative, autoClose: ONE_HOUR },
       )
       toastIdsRef.current.set(id, toastId)
     })
--- a/redisinsight/ui/src/components/notifications/Notifications.tsx
+++ b/redisinsight/ui/src/components/notifications/Notifications.tsx
@@ -123,15 +123,17 @@ const Notifications = () => {
         toastIdsRef.current.set(id, toastId)
       },
     )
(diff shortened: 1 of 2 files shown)


redisinsight/ui/src/components/notifications/Notifications.tsx
```diff
@@ -165,16 +167,22 @@ const Notifications = () => {
             dispatch(removeInfiniteNotification(id))
           },
         },
-        { variant: riToast.Variant.Notice, autoClose: ONE_HOUR },
+        { variant: riToast.Variant.Informative, autoClose: ONE_HOUR },
       )
+      infiniteToastIdsRef.current.add(toastId)
       toastIdsRef.current.set(id, toastId)
```
KrumTy, line 173: why use a second storage? Isn't toastIdsRef enough for the operation?

redisinsight/ui/src/components/notifications/Notifications.tsx
```diff
@@ -123,15 +123,17 @@ const Notifications = () => {
         toastIdsRef.current.set(id, toastId)
       },
     )
-
-  const showInfiniteToasts = (data: InfiniteMessage[]) =>
+  const infiniteToastIdsRef = useRef(new Set<number | string>())
+
+  const showInfiniteToasts = (data: InfiniteMessage[]) => {
+    infiniteToastIdsRef.current.forEach((toastId) => {
+      setTimeout(() => {
```
KrumTy, line 130: this timeout is not cleared; 
it's possible that showInfiniteToasts can be called twice in a quick succession due to the useEffect trigger; 
not sure if that would cause bugs

pd-redis: Yeah, I want this to be cleared every time, but don't want it to look too jumpy, that is why the timeout is used at all

PR #4880, opened 2025-08-20
RI-7359 Added support for dark theme for the "Data content" section in Create Index wizard

valkirilov:
# Description

Make the icons in the "**Data content**" section on the first step of the **Create Index** flow on the **Vector Search** page to support dark theme:
- added new icons with light gray color, to be used for the dark theme
- slightly reworked the way we pass the icons to the Create Index flow, so we can toggle their appearance based on the theme 

image

# How it was tested

1. Open a connection to a new or existing Database
2. Go to the **Search** page from the top navbar
- If you don't have any indexes yet, you'll see the "**Create Index**" wizard
- Otherwise, you should trigger the creation process by clicking "**Get Started**" from the alert below the top navbar
- Or simply, go to the following URL manually: http://localhost:8080/:databaseId/vector-search/create-index

3. On the first step of the creation process, observe the icons of the boxes in the "**Data content**" section

_PS: You can toggle the appearance from **Settings** -> **General** -> **Color Theme**_

| Before | After |
| - | - |
image|image
image| image


KrumTy:
@valkirilov  I would suggest a much simpler approach to what you are trying to achieve

Switch to main, open the [SVG](https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/assets/img/icons/bike.svg?short_path=ca52e56) and change `fill="black"` to `fill="currentColor"` and  the icon should accept the theme color out of the box

valkirilov:
> @valkirilov I would suggest a much simpler approach to what you are trying to achieve
> 
> Switch to main, open the [SVG](https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/assets/img/icons/bike.svg?short_path=ca52e56) and change `fill="black"` to `fill="currentColor"` and the icon should accept the theme color out of the box

Thanks! Definitely, it's much simpler to use this approach. I was confused to follow the pattern we already have for the rest of the icons, where we have two variants for each, but this way is much better. Thank you.

PR #4881, opened 2025-08-20
[RI-7374] [RI-7373]: update Saved Queries layout to match new design

KrumTy:
## Description 

This PR addresses two issues related to the same sections 
- RI-7374 - adjusts the Saved Queries wrapper to match the new design
- RI-7373 - adds close button to Saved Queries & Manage indexes
- adjusts paddings for Saved Queries & Manage indexes to match the new design

| Before | After |
| --- | --- |
| Screenshot 2025-08-20 at 16 47 48 | Screenshot 2025-08-20 at 16 47 20 |

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/saved-queries/styles.ts
+++ b/redisinsight/ui/src/pages/vector-search/saved-queries/styles.ts
@@ -14,18 +14,6 @@ export const VectorSearchSavedQueriesSelectWrapper = styled.div`
   align-items: center;
 `
 
-export const VectorSearchSavedQueryCardWrapper = styled.div`
-  display: flex;
-  flex-direction: row;
-  justify-content: space-between;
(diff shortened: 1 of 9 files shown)


redisinsight/ui/src/pages/vector-search/saved-queries/styles.ts
```diff
@@ -14,6 +14,18 @@ export const VectorSearchSavedQueriesSelectWrapper = styled.div`
   align-items: center;
 `
 
+export const VectorSearchSavedQueryCardWrapper = styled.div`
```
valkirilov: I'm fine with a simple div, but here maybe @pd-redis can advise on the correct component, because most probably we have something in our layout components suitable for that purpose, like `FlexGroup`.

valkirilov (approved):

PR #4887, opened 2025-08-21
RI-7366: Saved Queries - show empty message when no indexes are available

KrumTy:
### Description

- After creating a new index, automatically selects that index in saved queries panel and expands it
- When no indexes are available, Saved Queries shows a placeholder component

| Before | After |
| --- | --- |
| always showing the preset queries, even when no index has been inserted | when no index is available, show the same component used as in Manage Indexes |
| Screenshot 2025-08-21 at 15 38 30 | Screenshot 2025-08-21 at 14 27 04 |

There is a new design for the "No indexes" component which will be addressed in another ticket

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.tsx
+++ b/redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.tsx
@@ -1,23 +1,15 @@
 import React, { useState, useMemo, useEffect } from 'react'
 
-import { Title, Text } from 'uiSrc/components/base/text'
+import { Title } from 'uiSrc/components/base/text'
 
-import { RiSelect } from 'uiSrc/components/base/forms/select/RiSelect'
-import { EmptyButton, IconButton } from 'uiSrc/components/base/forms/buttons'
-import { FieldTag } from 'uiSrc/components/new-index/create-index-step/field-box/FieldTag'
+import { IconButton } from 'uiSrc/components/base/forms/buttons'
 import { FieldTypes } from 'uiSrc/pages/browser/components/create-redisearch-index/constants'
 import { Loader } from 'uiSrc/components/base/display'
 
-import { CancelSlimIcon, PlayFilledIcon } from 'uiSrc/components/base/icons'
-import {
-  RightAlignedWrapper,
-  TagsWrapper,
-  VectorSearchSavedQueriesContentWrapper,
-  VectorSearchSavedQueriesSelectWrapper,
-} from './styles'
+import { CancelSlimIcon } from 'uiSrc/components/base/icons'
+import { VectorSearchSavedQueriesContentWrapper } from './styles'
 import { SavedIndex } from './types'
 import {
-  VectorSearchScreenBlockWrapper,
   VectorSearchScreenFooter,
   VectorSearchScreenHeader,
   VectorSearchScreenWrapper,
@@ -28,6 +20,8 @@ import { useRedisearchListData } from '../useRedisearchListData'
 import { collectChangedSavedQueryIndexTelemetry } from '../telemetry'
 import { PresetDataType } from '../create-index/types'
 import NoIndexesMessage from '../manage-indexes/NoIndexesMessage'
+import { QueryCard } from './QueryCard'
+import { IndexSelect } from './IndexSelect'
 
 const mockSavedIndexes: SavedIndex[] = [
   {
@@ -82,6 +76,7 @@ export const SavedQueriesScreen = ({
   const selectedIndexItem = savedIndexes.find(
     (index) => index.value === selectedIndex,
   )
+  const hasIndexes = savedIndexes.length > 0
 
   useEffect(() => {
     if (selectedIndex) return
@@ -98,10 +93,6 @@ export const SavedQueriesScreen = ({
     collectChangedSavedQueryIndexTelemetry({
       instanceId,
     })
-  }
-
-  if (loading) {
-    return <Loader data-testid="manage-indexes-list--loader" />
   }
 
   return (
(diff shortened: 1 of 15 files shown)


redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.tsx
```diff
@@ -62,35 +122,35 @@ export const SavedQueriesScreen = ({
...
+                data-testid="select-saved-index"
+                onChange={onIndexChange}
+                valueRender={({ option, isOptionValue }) =>
+                  isOptionValue ? (
+                    option.value
+                  ) : (
+                    <TagsWrapper>
+                      {option.value}
+                      {option.tags.map((tag) => (
+                        <FieldTag key={tag} tag={tag as any} />
```
valkirilov: nit: Can't we use `as FieldTypes` instead of `as any`, or do we need to break the type checking for some reason?

redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.tsx
```diff
@@ -62,35 +122,35 @@ export const SavedQueriesScreen = ({
...
+                    </TagsWrapper>
+                  )
+                }
+              />
+            </VectorSearchSavedQueriesSelectWrapper>
+          ) : (
+            <NoIndexesMessage />
+          )}
+          {selectedIndexItem?.queries.map((query) => (
+            <VectorSearchScreenBlockWrapper key={query.value} padding={6}>
```
pawelangelow: Nit: 😢 It's a personal preference, but I prefer if we duplicate the wrapper logic with early exit for <NoIndexesMessage />:

```
const hasIndexes = savedIndexes.length > 0
if (!hasIndexes) {
  return (
    <VectorSearchScreenWrapper direction="column" data-testid="saved-queries-screen">
      <VectorSearchScreenHeader padding={6}>
        <Title size="S" data-testid="title">Saved queries</Title>
        <IconButton
          size="XS"
          icon={CancelSlimIcon}
          aria-label="Close"
          data-testid="close-saved-queries-btn"
          onClick={onClose}
        />
      </VectorSearchScreenHeader>
      <VectorSearchScreenFooter grow={1} padding={6}>
        <VectorSearchSavedQueriesContentWrapper>
          <NoIndexesMessage />
        </VectorSearchSavedQueriesContentWrapper>
      </VectorSearchScreenFooter>
    </VectorSearchScreenWrapper>
  )
}
```

and after it we can still extract few things:
```
const renderSelectValue = ({
  option,
  isOptionValue,
}: {
  option: SavedIndex
  isOptionValue: boolean
}) => {
  if (isOptionValue) return option.value
  return (
    <TagsWrapper>
      {option.value}
      {option.tags.map((tag) => (
        <FieldTag key={tag} tag={tag as any} />
      ))}
    </TagsWrapper>
  )
}

const QueryCard = ({
  label,
  value,
  onInsert,
}: {
  label: string
  value: string
  onInsert: (v: string) => void
}) => (
  <VectorSearchScreenBlockWrapper key={value} padding={6}>
    <Text>{label}</Text>
    <RightAlignedWrapper>
      <EmptyButton
        icon={PlayFilledIcon}
        onClick={() => onInsert(value)}
        data-testid="btn-insert-query"
      >
        Insert
      </EmptyButton>
    </RightAlignedWrapper>
  </VectorSearchScreenBlockWrapper>
)
```

redisinsight/ui/src/pages/vector-search/useRedisearchListData.ts
```diff
@@ -27,5 +31,6 @@ export const useRedisearchListData = () => {
   return {
     loading,
     data,
+    stringData,
```
valkirilov, line 34: Note: To be honest, I'm not quite sure whether I like the idea to return the same data in two different variants (buffer and a string), so I would propose to return only the string representation if that’s what we need. 

However, I've heard a few times that buffer to string conversion is not always safe (something due to size limitation and so on), so I'm not sure whether it's applicable to this use case.

Consider this a note, in case you're not sure either.

valkirilov (approved):

redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.tsx
```diff
@@ -22,26 +22,86 @@ import {
...
+  )
+  const selectedIndexItem = savedIndexes.find(
+    (index) => index.value === selectedIndex,
+  )
+
+  useEffect(() => {
+    if (selectedIndex) return
+
+    // Select the first index by default if none is selected yet
+    const firstIndex = savedIndexes[0]?.value
```
pawelangelow, line 85: Probably this is going to be changed, but I don't know how exactly we can select the new index here. Do we have a timestamp for them?

redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.tsx
```diff
@@ -22,26 +22,86 @@ import {
 } from '../styles'
 import { useTelemetryMountEvent } from '../hooks/useTelemetryMountEvent'
 import { TelemetryEvent } from 'uiSrc/telemetry'
+import { useRedisearchListData } from '../useRedisearchListData'
+import { collectChangedSavedQueryIndexTelemetry } from '../telemetry'
+import { PresetDataType } from '../create-index/types'
+import { Loader } from '@redis-ui/components'
+import NoIndexesMessage from '../manage-indexes/NoIndexesMessage'
+
+const mockSavedIndexes: SavedIndex[] = [
```
pawelangelow, line 26: Actually, these are the `presetSavedIndexes`

redisinsight/ui/src/pages/vector-search/saved-queries/SavedQueriesScreen.tsx
KrumTy: currently I have configured it to pass it via state param after create

the line you are referring to is when no new index was created so the default order is used, I don't know what it is. Maybe we can do alphabetical in the 2nd phase when more indexes can be inserted from here

pawelangelow: We can revisit it later. It's fine for now.

pawelangelow (approved):

PR #4889, opened 2025-08-22
RI-7363: Add Redis instance requirements check for vector search

pawelangelow:
# Description

Added fallback banners to let the users know when they can't use the new Vector Search page

## Redis Query Engine (RQE) is not available

When **Redis Query Engine** module is not available, the whole page is disabled.

image

# Vector Sets are not available

When they use an older version of Redis (below 7.2) and Vector Sets are not available, they see an info banner.

image

_Note: We have similar interfaces in other parts of the app, and I saw that the button leading to the "Cloud Login" is protected behind a feature flag. Because of this, I used the same behavior for the new screens, so in case we're in a build with the feature flags disabled, the button won't show._

# How it was tested

## Redis Query Engine (RQE) is not available

1. First, you need a Redis instance without the Redis Query Engine module enabled

_Note: You can use the following command to create a Docker container with such an instance_
```bash
docker run -d --name redis-no-redisearch -p 6401:6379 redis:7.4-alpine
```

2. Go to **Databases** page and connect to the instance
3. Go to **Search** page (from the main navbar)

You'll see a message that you can't use this feature.

| Light | Dark |
| - | - |
image|image

### Note

Depending on whether you're already logged in to Redis Cloud and have Cloud Databases imported in Redis Insight, the wording of the message may differ, so the users will know they can either create a cloud database or use an existing one.

| Not logged into Redis Cloud | Logged into Redis Cloud |
| - | - |
image|image

#### Import Cloud Databases

1. Click on the "**Cloud Sign In**" button the top right corner and make sure to login with Google
2. Click on the "**Cloud Account**" button in the top right corner, and select "**Import Cloud databases**" from the dropdown
3. Make sure to finish the import, so you have a cloud database saved in Redis Insight

| No Cloud Databases | Available Cloud Databases |
| - | - |
image|image

## Vector Seta are not available

1. First, you need an older Redis instance without support for Vector Sets (older than 7.2)

_Note: You can use the following command to create a Docker container with such an instance_
```bash
docker run -d --name redis6-redismod -p 6404:6379 redislabs/redismod:latest
```

4. Go to **Databases** page and connect to the instance
5. Go to **Search** page (from the main navbar)

You'll see the search page, but with a banner to let you know that you can't use the full functionalities.

| Light | Dark |
| - | - |
image|image
image|image


redisinsight/ui/src/pages/vector-search/create-index/hooks/useRedisInstanceCompatibility.ts
```diff
@@ -0,0 +1,47 @@
...
+  loading: boolean | undefined
+  hasRedisearch: boolean | null
+  hasSupportedVersion: boolean | null
+}
+
+const MIN_SUPPORTED_REDIS_VERSION = '7.2.0'
+const REDISEARCH_MODULE_SET = new Set(REDISEARCH_MODULES)
+
+export const isVersionSupported = (raw?: string | null) => {
+  if (!raw) return null
```
KrumTy:
```suggestion
export const isVersionSupported = (raw?: string | null): boolean => {
  if (!raw) return false
```

redisinsight/ui/src/pages/vector-search/create-index/hooks/useRedisInstanceCompatibility.ts
```diff
@@ -0,0 +1,47 @@
...
+  (): UseRedisInstanceCompatibilityReturn => {
+    const {
+      loading,
+      modules = [],
+      version,
+    } = useSelector(connectedInstanceSelector)
+
+    const hasRedisearch = modules
+      ? modules.some((m) => REDISEARCH_MODULE_SET.has(m.name))
+      : null
```
KrumTy:
```suggestion
    const hasRedisearch = modules.some((m) => REDISEARCH_MODULE_SET.has(m.name))
```

pawelangelow: Keep in mind, this data is calculated based on what we get from `connectedInstanceSelector` and as such, it might return `loading: true` and `modules` to be some default value as well as `version` a.k.a. not yet determined. That's why I wanted to support the third state. If you have a suggestion for it, please go ahead and share it 🐱

KrumTy: "has" prefix implies boolean
you can use enums if there are more than two values

the current hook's implementation only looks at the state - e.g. it doesn't trigger anything that would cause loading to be changed, so this is a bit odd on it's own

an alternative solution would be to use some sort of async method

but then again, this is just a hook without actual usage so I can't provide best solution

pawelangelow: The real usage is something like
```
const isVersionOk = ...

if(isVersionOk) {
  <Nice />
} else {
  <NotNice />
}
```
Well, this is a hook, if reactivity is your concern. useSelector would update when the state changes.

About the shape - IMO its standard data fetching hook - loading, data, error. But with changed API for more intuitive use. If loading is true - other two values are null/undefined. React Query or SWR have similar concepts - data, loading, error.

redisinsight/ui/src/pages/vector-search/create-index/hooks/useRedisInstanceCompatibility.ts
valkirilov: Done 👍

redisinsight/ui/src/pages/vector-search/create-index/hooks/useRedisInstanceCompatibility.ts
valkirilov: I think that the main concern of @KrumTy here was related to the fact that the variable is called `hasSomething` (which implies its value to be a boolean), but with the existing implementation we might end up with `true`, `false`, and `null/undefined`. But at the end of the day, I think we treat all of the falsy values the same way (at least for now). 

Anyway, I made a minor change to simplify it. The hook still does the main job it was built for.

dantovska:
RqeNotAvailableCard looks a lot like the ModuleNotFound modal... have you considered reusing this?

redisinsight/ui/src/pages/vector-search/components/rqe-not-available/RqeNotAvailableCard.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { StyledCard, StyledCardBody } from './ReqNotAvailableCard.styles'
+import { ModuleNotLoaded } from 'uiSrc/components'
+import { RedisDefaultModules } from 'uiSrc/slices/interfaces'
+
+export const RqeNotAvailableCard = () => {
+  return (
+    <StyledCard data-testid="rqe-not-available-card">
+      <StyledCardBody direction="column" gap="xxl">
+        <ModuleNotLoaded
```
valkirilov, line 10: > RqeNotAvailableCard looks a lot like the ModuleNotFound modal... have you considered reusing this?

Very good observation, thanks for the note @dantovska. I think I have managed to reuse the existing visuals instead of defining a new interface, similar to what we already had.

PR #4890, opened 2025-08-25
RI-7354: fix the control + enter button to run the query in on Search Page

dantovska:
image

On the Vector Search page, clicking **Control + Enter** should do the same as the **Run** button. This did not work as the value was not passed to to the onQuerySubmit method. Also, changed the onClick event for the button in QueryLiteActions, as it was passing an event which is not desired scenario.

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
+++ b/redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
@@ -64,8 +64,8 @@ export const VectorSearchQuery = ({
   )
   const isSavedQueriesOpen = rightPanel === RightPanelType.SAVED_QUERIES
 
-  const onQuerySubmit = () => {
-    onSubmit()
+  const onQuerySubmit = (value?: string) => {
+    onSubmit(value)
     collectTelemetryQueryRun({
       instanceId,
       query,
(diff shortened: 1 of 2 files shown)


redisinsight/ui/src/pages/vector-search/query/VectorSearchQuery.tsx
```diff
@@ -64,8 +64,8 @@ export const VectorSearchQuery = ({
   )
   const isSavedQueriesOpen = rightPanel === RightPanelType.SAVED_QUERIES
 
-  const onQuerySubmit = () => {
-    onSubmit()
+  const onQuerySubmit = (value?: string) => {
+    onSubmit(value)
```
KrumTy, line 68: what is this value? Is it the query?
From what I see onSubmit has 3 params and I think it's meant to use the local state query by default as first param (set via setQuery)

dantovska: The value corresponds to the commandInit / query argument. If you go to the end where onSubmit is used, or better - start from the event handling the keyboard event calling onSubmit, the query value is passed there to the onSubmit and left to be handled here. 

```ui/src/pages/workbench/components/query/Query.tsx:editorDidMount:265-271```

I'm open for refactoring ideas, if you have something else in mind

KrumTy (approved):

PR #4893, opened 2025-08-25
RI-7353: fix database info tooltip

dantovska:
| Before | After |
| ----- | ----- |
| image | image |

Changes to the base UI components:
- extend text with variant property
- extend image with xs size for the size

### DIFF
--- a/redisinsight/ui/src/components/base/text/Text.tsx
+++ b/redisinsight/ui/src/components/base/text/Text.tsx
@@ -8,7 +8,6 @@ export const Text = ({
   color,
   size,
   textAlign,
-  weight,
   ...rest
 }: TextProps) => {
   const sizeMap = {
@@ -27,7 +26,6 @@ export const Text = ({
       className={cn(className, 'RI-text')}
       $color={color}
       $align={textAlign}
-      $weight={weight}
       size={sizeMap.size as BodySizesType}
     />
   )
--- a/redisinsight/ui/src/components/base/text/Text.tsx
+++ b/redisinsight/ui/src/components/base/text/Text.tsx
@@ -8,6 +8,7 @@ export const Text = ({
   color,
   size,
   textAlign,
+  variant,
   ...rest
 }: TextProps) => {
   const sizeMap = {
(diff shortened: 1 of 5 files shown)


valkirilov (dismissed):

redisinsight/ui/src/components/base/text/Text.tsx
```diff
@@ -26,6 +27,7 @@ export const Text = ({
       className={cn(className, 'RI-text')}
       $color={color}
       $align={textAlign}
+      $weight={weight}
```
pd-redis: https://redislabsdev.github.io/redis-ui/?path=/story/components-typography-body--variants

Although the name is not the best, the text component already has this

KrumTy (approved):

redisinsight/ui/src/components/base/text/Text.tsx
```diff
@@ -8,6 +8,7 @@ export const Text = ({
   color,
   size,
   textAlign,
+  variant,
```
pd-redis: I am not sure what is the point of explicitly passing this to the parent?
Are you doing something with this prop?

redisinsight/ui/src/components/base/text/text.styles.ts
```diff
@@ -31,6 +31,7 @@ export type TextProps = Omit<BodyProps, 'color' | 'size'> &
     color?: ColorType
     size?: BodyProps['size'] | 'm' | 's' | 'xs'
     textAlign?: 'left' | 'center' | 'right'
+    variant?: BodyProps['variant'] | 'regular' | 'semiBold' | 'italic'
```
pd-redis: What am I missing? Why do we need to explicitly declare this?

redisinsight/ui/src/components/base/text/Text.tsx
dantovska: I was thinking of extending this functionality by adding bold as well, but decided not to and missed to remove this... tnx!

redisinsight/ui/src/components/base/text/text.styles.ts
dantovska: I was thinking of extending this functionality by adding bold as well, but decided not to and missed to remove this... tnx!

KrumTy (approved):

valkirilov (approved):

PR #4894, opened 2025-08-25
[RI-7366] Vector Search - Saved Queries not opening for electron builds

KrumTy:
## Description

After creating an index, the user should be navigated to the root Vector Search page and have his Saved Queries for this index opened.

It seems that using react router's state param does not work for electron in this case. I have replaced it with query params solution and seems to be working fine across builds

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/pages/VectorSearchPage.tsx
+++ b/redisinsight/ui/src/pages/vector-search/pages/VectorSearchPage.tsx
@@ -10,13 +10,13 @@ import { VectorSearchPageWrapper } from './../styles'
 type Params = {
   instanceId: string
 }
-type LocationState = {
-  defaultSavedQueriesIndex?: string
-}
 
 const VectorSearchPage = () => {
   const { instanceId } = useParams<Params>()
-  const { state } = useLocation<LocationState | undefined>()
+  const { search } = useLocation()
+  const defaultSavedQueriesIndex =
+    new URLSearchParams(search).get('defaultSavedQueriesIndex') || undefined
+
   usePageViewTelemetry({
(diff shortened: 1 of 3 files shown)


redisinsight/ui/src/pages/vector-search/pages/VectorSearchPage.tsx
```diff
@@ -10,13 +10,13 @@ import { VectorSearchPageWrapper } from './../styles'
...
-type LocationState = {
-  defaultSavedQueriesIndex?: string
-}
 
 const VectorSearchPage = () => {
   const { instanceId } = useParams<Params>()
-  const { state } = useLocation<LocationState | undefined>()
+  const { search } = useLocation()
+  const defaultSavedQueriesIndex =
+    new URLSearchParams(search).get('defaultSavedQueriesIndex') || undefined
```
valkirilov, line 18: tip: You can put this param behind a constant, so we can easily change it in the future, without the need to apply the same change across the codebase. Also, it will reduce the chance of typos.

```js
export const DEFAULT_SAVED_QUERIES_INDEX_PARAM = 'defaultSavedQueriesIndex'

const defaultSavedQueriesIndex =
    new URLSearchParams(search).get(DEFAULT_SAVED_QUERIES_INDEX_PARAM) || undefined

history.push({
    pathname: Pages.vectorSearch(instanceId),
    search: `?${DEFAULT_SAVED_QUERIES_INDEX_PARAM}=${encodeURIComponent(...)
})`
```

KrumTy: yes, however I didn't want to create a new constants.ts file just for that
I also expect this behavior to possibly change with next version

PR #4899, opened 2025-08-27
RI-7384: fix plugins imports to avoid riConfig or redux state initialization errors

ArtemHoruzhenko:
The main problem was related to how components imported to plugins.
e.g. `import { RiTooltip } from '../../../../../components'` will fail because bootstrap file in the components folder has more components which might require riConfig or redux state to be initialized
when change to `import { RiTooltip } from '../../../../../components/base/tooltip/RITooltip'` everything works since RiTooltip is stateless itself


pd-redis (approved):
especially for plugins, we should avoid import from top level

PR #4900, opened 2025-08-27
RI-7365 Redesign "No indexes" messages in "Manage Indexes" and "Saved Queries" panels

valkirilov:
# Description

Redesign the fallback messages that we use to inform the users that they don't have any data to be displayed in the "**Manage Indexes**" and "**Saved Queries**" panels on the **Vector Search** page:
- Make the `NoIndexesMessage` component reusable and accept an input variant that toggles the displayed information based on where it is used
- Slightly redesign the message, add an image icon, and update the text
- Rename the component and apply some general refactoring, since it's now reusable and shared


redisinsight/ui/src/pages/vector-search/components/no-data-message/data.ts
dantovska, line 1: Have you considered naming this constants?

dantovska (approved):
Looks awesome!

redisinsight/ui/src/pages/vector-search/components/no-data-message/data.ts
valkirilov: I'm not sure that I understand your idea. Currently, you can use these constants like that:

```
NO_DATA_MESSAGES[NoDataMessageKeys.ManageIndexes].title
NO_DATA_MESSAGES[NoDataMessageKeys.ManageIndexes].description
NO_DATA_MESSAGES[NoDataMessageKeys.ManageIndexes].icon
```

dantovska: Sorry, I meant the file name. Having the name data usually leads me to think that is mocked data or something, but I "googled" it and maybe your name is better.

valkirilov: Ahaaa, got it. Do you have an internal convention for that? 
It's not a problem to rename it to `constants.ts` if that's the case 🙂

PR #4901, opened 2025-08-27
RI-6962: add support to switch between time units for TS plugin visualization

ArtemHoruzhenko:
In scope of this task new switcher for time units was added below the chart
Based on data we will determine default time unit: seconds or milliseconds
Anyway user will be able to switch manually

Screenshot 2025-08-27 at 15 56 58

### DIFF
--- a/redisinsight/ui/src/packages/redistimeseries-app/src/components/Chart/ChartResultView.tsx
+++ b/redisinsight/ui/src/packages/redistimeseries-app/src/components/Chart/ChartResultView.tsx
@@ -1,13 +1,17 @@
 import React, { useState } from 'react'
 import {
+  AxisScale,
+  ChartConfig,
+  GraphMode,
   TimeSeries,
   YAxisConfig,
-  ChartConfig,
-  AxisScale,
-  GraphMode,
 } from './interfaces'
 import ChartConfigForm from './ChartConfigForm'
 import Chart from './Chart'
+import {
+  determineDefaultTimeUnits,
+  normalizeDatapointUnits,
+} from 'uiSrc/packages/redistimeseries-app/src/components/Chart/utils'
 
 enum LAYOUT_STATE {
   INITIAL_STATE,
@@ -30,6 +34,7 @@ export default function ChartResultView(props: ChartResultViewProps) {
 
   const [chartConfig, setChartConfig] = useState<ChartConfig>({
     mode: GraphMode.line,
+    timeUnit: determineDefaultTimeUnits(props.data),
     title: '',
     xlabel: '',
     staircase: false,
@@ -77,7 +82,7 @@ export default function ChartResultView(props: ChartResultViewProps) {
       </div>
       <Chart
         chartConfig={chartConfig}
-        data={props.data}
+        data={normalizeDatapointUnits(props.data, chartConfig.timeUnit)}
         onRelayout={onRelayout}
         onDoubleClick={onDoubleClick}
       />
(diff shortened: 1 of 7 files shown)


redisinsight/ui/src/packages/redistimeseries-app/src/components/Chart/ChartResultView.tsx
```diff
@@ -77,7 +82,7 @@ export default function ChartResultView(props: ChartResultViewProps) {
       </div>
       <Chart
         chartConfig={chartConfig}
-        data={props.data}
+        data={normalizeDatapointUnits(props.data, chartConfig.timeUnit)}
```
KrumTy: is it worth memoizing this result? I'm not sure how often ChartResultView rerenders and how often the props.data changes

ArtemHoruzhenko: @KrumTy As I can see data is `ts.range` command result, which is permanent for plugin visualization.
however we need to recalculate it each time we change time units.
let me know if it does make sense to memoize this and I'll do if needed

KrumTy: hm I see these setChartState/setChartConfig state change functions that can be called in various callbacks

any time they are called, the data will be recalculated, so I think it makes sense to have it in a useMemo

redisinsight/ui/src/packages/redistimeseries-app/src/components/Chart/utils.ts
```diff
@@ -134,3 +139,32 @@ export function hexToRGBA(hex: string, alpha: number): string {
...
+      })
+    }
+  } catch (e) {
+    // ignore an error to return original data
+  }
+
+  return timeSeries
+}
+
+export function determineDefaultTimeUnits(timeSeries: TimeSeries[]): TimeUnit {
```
valkirilov, line 166: I remember that you’ve already mentioned that it's a challenging task to determine the "time frame" unit precisely because this number can vary widely. And I think you're right.

Here’s a snapshot of the chart for the default `bike_sales` time series, which was generated using one of the official tutorials.

Picking seconds results in that (see the labels on the x-axis):
Image

At least, it looks fine for milliseconds, although I'm not sure whether it's correct.

Image

I guess it's the best we can do so far, so this is simply a note, for my observations. I guess, we're ok with this end result.

valkirilov (dismissed):

redisinsight/ui/src/packages/redistimeseries-app/src/components/Chart/ChartResultView.tsx
ArtemHoruzhenko: done, thank you for suggestion

redisinsight/ui/src/packages/redistimeseries-app/src/components/Chart/utils.ts
ArtemHoruzhenko: @valkirilov correct, but by default did you see milliseconds result?

valkirilov: Yes, by default, it was milliseconds, no worries.

ArtemHoruzhenko: So I hope that we covered 99% of cases with such approach. Let's see if/how we can improve in the future

valkirilov (approved):

KrumTy (approved):

PR #4903, opened 2025-08-28
RI-7341: Fix big icons on Browser Page

dantovska:
- Bulk Actions button: replaced it with an icon from Redis UI Components (Subscriptions icon), which has the right size and does not need additional sizing and alignment
- Refactor the search mode switch button using styled components to have normal sized buttons and normal behavior, compared to the previous release implementation

| Before | After |
| ----- | ---- | 
| image | image |


valkirilov (dismissed):
Thanks! Finally, this icon is fixed 😀🙌

redisinsight/ui/src/components/base/forms/buttons/Button.tsx
```diff
@@ -20,29 +20,36 @@ export const BaseButton = ({
   iconSide = 'left',
   loading,
   size = 'medium',
+  iconSize,
```
pd-redis: I don't think this is the correct approach. 
If the icon looks too big for the button, it just might be, why not look into adjusting the icon?
Or we're going to have to add this prop to each button with an icon?
If that is the case, then better approach will be to change the sizes in size map to better match what we need

redisinsight/ui/src/pages/browser/components/browser-search-panel/styles.module.scss
```diff
@@ -105,6 +82,7 @@
     border-radius: 4px !important;
 
     svg {
+      width: 26px;
```
pd-redis: Is this a magic number? Where did you get it from?

redisinsight/ui/src/components/base/forms/buttons/Button.tsx
dantovska: Well, I didn't see other issues similar to this one, so I don't know if I should apply something that will affect others. 
What do you mean by "into adjusting the icon"? 
"Or we're going to have to add this prop to each button with an icon?" - no, I left the implementation as it was - the icon size was following the button size, and will still follow it, unless set specifically.

redisinsight/ui/src/pages/browser/components/browser-search-panel/styles.module.scss
dantovska: few rows above, from the viewTypeBtn style's height.

redisinsight/ui/src/components/base/forms/buttons/Button.tsx
pd-redis: If all other icons look fine, using current approach, then obviously this icon is somehow off :)
So designer should fix it - add a little more space for the SVG

dantovska: why was the size introduced to the ButtonIcon in the first place?

valkirilov (approved):

redisinsight/ui/src/pages/browser/components/browser-search-panel/BrowserSearchPanel.tsx
```diff
@@ -45,6 +45,30 @@ export interface Props {
...
+  height: 26px;
+  border-radius: ${({ theme }) => theme.core.space.space050};
+
+  padding: ${({ theme }) => theme.core.space.space100};
+  margin: ${({ theme }) => theme.core.space.space050};
+
+  svg {
+    color: ${({ theme }) => theme.semantic.color.text.neutral700};
+
+    g > path {
```
pd-redis, line 59: wow

pd-redis (approved):

redisinsight/ui/src/pages/browser/components/browser-search-panel/BrowserSearchPanel.tsx
dantovska: meaning? 😄

PR #4904, opened 2025-08-28
RI-7376: Remove launch database icon

dantovska:
- remove button icon;
- increase button size to match the components around it

| Before | After |
| ---- | ---- |
| image | image-20250827-214746 |


pd-redis (dismissed):

valkirilov (dismissed):

redisinsight/ui/src/components/messages/module-not-loaded/ModuleNotLoaded.tsx
```diff
@@ -161,24 +164,27 @@ const ModuleNotLoaded = ({
...
-                  </>
-                ) : (
-                  item
-                ),
-              )}
-            </Text>
+            <>
+              <Spacer size="l" />
+              <Text
+                className={cx(
```
pd-redis, line 170: Since you're refactoring it anyway, why not use styled instead of classes?

dantovska: Sounds good. However, turned out we have one more component like this one and one new one, so we're to sync them all in one (in a different task). We can discuss more offline.

valkirilov (approved):

PR #4906, opened 2025-08-28
RI-7329: Align TestConnection to the left side

dantovska:
### Align TestConnection to the left side as previously implemented

| Before | After |
| ---- | ---- |
| image | image |


pd-redis:
Please discard this PR.
Similar approach is needed in at least one more place and I added it in the keys PR

PR #4909, opened 2025-08-29
RI-7364 Connection to server lost

valkirilov:
# Description

Fix the "Connection Status" indicator in the Database Overview section to not alert that the connection is lost when it actually is active. It was happening when you try to establish a connection with another (not-running) database, wrongly affecting the status indicator for the current database connection.

_PS: Additionally, I found a weird behavior that the manual Refresh of the connection status was not working consistently. You can learn more about that in the comments below._ 

image

# How it was tested

1. Open Redis Insight and connect to a (_running_) database instance, you should be able to do so
2. Try to switch the database connection to another **not-running** database instance, a toast notification will inform you, and the status icon in the top navbar shouldn't turn red
3. If you turn off the Redis instance of the active connection, a toast notification will inform you about the lost connection, and the status icon in the top navbar will turn red

## Before 

As you can see, when we're connected to a database instance and try to establish a connection with other (not-running) database, it also affects the status indicator of the current database.

https://github.com/user-attachments/assets/7b8b031c-eb97-403f-b86a-6d8d374f25fb

## After

Now, we no longer update the status indicator when we receive a "connection lost" error if it's not related to the active connection.

https://github.com/user-attachments/assets/f63e934b-7257-485a-9f0e-aa13a00f36ea

Of course, if we receive a "connection lost" error related to the active connection, we do update the status indicator, as expected.

https://github.com/user-attachments/assets/12caef3b-8001-463d-8d9a-98839ee1d79f

### DIFF
--- a/redisinsight/ui/src/components/database-overview/hooks/useDatabaseOverview.ts
+++ b/redisinsight/ui/src/components/database-overview/hooks/useDatabaseOverview.ts
@@ -48,8 +48,8 @@ export const useDatabaseOverview = () => {
     } = {},
   } = overview
 
-  const loadData = () => {
-    if (connectedInstanceId && !connectivityError) {
+  const loadData = (forceDataReload = false) => {
+    if (connectedInstanceId && (!connectivityError || forceDataReload)) {
       dispatch(getDatabaseConfigInfoAction(connectedInstanceId))
       setLastRefreshTime(Date.now())
     }
@@ -75,8 +75,8 @@ export const useDatabaseOverview = () => {
       },
     })
 
-  const handleRefresh = () => {
-    loadData()
+  const handleRefresh = (forceRefresh?: boolean) => {
+    loadData(forceRefresh)
   }
   const handleRefreshClick = () => {
     // clear error, if connectivity is broken, the interceptor will set it again
--- a/redisinsight/ui/src/services/apiService.ts
+++ b/redisinsight/ui/src/services/apiService.ts
@@ -104,6 +104,7 @@ export const isConnectivityError = (
 
 export const connectivityErrorsInterceptor = (error: AxiosError) => {
   const { response } = error
+  const responseUrl = response?.request?.responseURL || ''
   const responseData = response?.data as {
     message?: string
     code?: string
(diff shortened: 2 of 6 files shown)


redisinsight/ui/src/components/database-overview/hooks/useDatabaseOverview.ts
```diff
@@ -48,8 +48,8 @@ export const useDatabaseOverview = () => {
     } = {},
   } = overview
 
-  const loadData = () => {
-    if (connectedInstanceId && !connectivityError) {
+  const loadData = (forceDataReload = false) => {
+    if (connectedInstanceId && (!connectivityError || forceDataReload)) {
```
valkirilov, line 52: This change is necessary, because I found a weird behavior when you manually run the "Refresh" database connection status check. On every click, it was toggling between "connection lost" and "connection established" states.

https://github.com/user-attachments/assets/72c5fa4b-b75f-4c60-ba41-061430ed13b8

This logic, [implemented](https://github.com/redis/RedisInsight/pull/4518/commits/8e38af43f44ab3e83830d834413d1aa38e452fc7) a few months ago, was designed to safeguard us from infinite requests initiated by the auto-refresh functionality. This prevents us from reloading data when there’s no active connection. 

Unfortunately, this logic also affected manual refreshes, so I had to extend the behavior to ensure that data is always reloaded when triggered by user interaction. Here is how it works now, after the fix.

https://github.com/user-attachments/assets/ef081b21-3250-4a04-b8d5-7d4da07100bf

redisinsight/ui/src/services/apiService.ts
```diff
@@ -118,7 +119,16 @@ export const connectivityErrorsInterceptor = (error: AxiosError) => {
...
 
-    store?.dispatch<any>(setConnectivityError(message || ApiErrors.ConnectionLost))
+    const state = store.getState()
+    const isConnectedToDatabase =
+      !!state.connections.instances.connectedInstance?.id
+    const isErrorTargetsConnectedDatabase = responseUrl.includes(
+      state.connections.instances.connectedInstance?.id,
+    )
+
+    if (isConnectedToDatabase && isErrorTargetsConnectedDatabase) {
```
valkirilov, line 129: This is the action that should show the "connection lost" indicator, but we want to do it only if the error is related to the active connection that the user sees at the moment. 

If we're on the Databases page with the list of all saved connections, or we try to connect to another (not-running) database from the dropdown list of options in the top navbar, we shouldn't interfere with the status indicator for the current connection.

dantovska (approved):

PR #4918, opened 2025-09-02
Ri-7334: icons titles

pd-redis:
update redis-ui packages, update their setup, did not notice any major changes to the ui


valkirilov (dismissed):

dantovska:
I am getting a blank page when I try to open Redis Copilot when running locally using electron - `yarn dev:desktop`:
```
[2] [81605:0903/152406.029603:INFO:CONSOLE:5716] "Uncaught TypeError: Cannot read properties of undefined (reading 'sizes')", source: http://localhost:8080/node_modules/.vite/deps/chunk-IUQK4RL4.js?v=6db85dd7 (5716)
[2] [81605:0903/152406.036965:INFO:CONSOLE:14064] "The above error occurred in the <Linkstyle__StyledLink> component:
```

dantovska:
Also, variant no longer applies to Link:
```        
    <Link
          variant="small-inline"
          target="_blank"
          href={getUtmExternalLink(EXTERNAL_LINKS.pubSub, {
            medium: UTM_MEDIUMS.Main,
            campaign: UTM_CAMPAINGS.PubSub,
          })}
        >
          here.
   </Link>
```

.nvmrc
```diff
@@ -1 +1 @@
-22.12.0
\ No newline at end of file
+22
```
KrumTy, line 1: why drop the specific version?

dantovska:
The command helper links do not look very clear IMO:
| Dark | Light |
| --- | --- |
| image | image |


+ the pointer is turned to disabled sign on hover

redisinsight/ui/src/components/base/external-link/ExternalLink.tsx
```diff
@@ -1,15 +1,14 @@
...
+import { Link, type LinkProps } from 'uiSrc/components/base/link/Link'
 
-export type Props = EuiLinkProps & {
+export type Props = LinkProps & {
   href: string
   iconPosition?: 'left' | 'right'
   iconSize?: IconProps['size']
-  variant?: LinkButtonVariants
+  variant?: LinkProps['variant']
+  size?: LinkProps['size']
```
dantovska, line 11: I see that in the library there is no default size set, but I think we are using S in most of the places. Do you think it makes sense to set it as default in Redis Insight Link wrapper?

package.json
```diff
@@ -241,10 +241,10 @@
   "dependencies": {
     "@elastic/datemath": "^5.0.3",
     "@elastic/eui": "34.6.0",
-    "@redis-ui/components": "^38.1.4",
-    "@redis-ui/icons": "^4.16.1",
-    "@redis-ui/styles": "^11.4.0",
-    "@redis-ui/table": "^2.4.0",
+    "@redis-ui/components": "^39.18.0",
+    "@redis-ui/icons": "^6.1.1",
+    "@redis-ui/styles": "^12.9.0",
```
valkirilov, line 246: There is something wrong with the colors on our end, using dark theme and a [notice banner](https://redislabsdev.github.io/redis-ui/?path=/story/components-banner--semantic-banners&globals=theme:dark), in the Redis UI Storybook, it's purple, but in Redis Insight, it still looks pinkish.

The same thing is valid for the [toast notifications](https://redislabsdev.github.io/redis-ui/?path=/story/components-toast--semantic-toasts&globals=theme:dark) that use the same variant.

Image

Image

redisinsight/ui/src/components/command-helper/components/command-helper-search-output/CHSearchOutput.tsx
```diff
@@ -90,7 +90,7 @@ const CHSearchOutput = ({ searchedCommands }: Props) => {
                     handleClickCommand(e, command)
                   }}
                 >
-                  <UnderlineReverseLink color="text" variant="regular">
+                  <UnderlineReverseLink color="text" variant="button">
```
dantovska, line 93: Adding href (and switching back to inline variant) seems to fix it, which is crazy ... I don't know if we should use link here at all...

```suggestion
                  <UnderlineReverseLink href="#" color="text" variant="inline">
```

package.json
dantovska: image

works fine on my end

valkirilov: Great news! Interesting why it's failing on my end, though. 
Anyway, marking it as resolved then. @dantovska, thanks for the input.

pd-redis: Also, I don't think this PR is the place to fix it, if it is a problem

.nvmrc
pd-redis: because it had issues after update

redisinsight/ui/src/components/base/external-link/ExternalLink.tsx
pd-redis: I don't mind

redisinsight/ui/index.html
pd-redis, line 1: Modified this because on desktop, #root is not full height and that breaks connectivity error layout

PR #4920, opened 2025-09-03
RI-7259: Fix Toast message display by highlighting the title

dantovska:
#### Add styling to redis ui toast messages
- Make the title same size as the text, but bold
- Make the added / deleted or whatever contents needs to be highlighted - to be bold
- Change Ok to OK for success messages to match error messages 

| Before | After |
| ---- | ---- |
| Screenshot 2025-07-25 at 15 32 05 | image image |

### DIFF
--- a/redisinsight/ui/src/components/base/display/toast/RiToast.tsx
+++ b/redisinsight/ui/src/components/base/display/toast/RiToast.tsx
@@ -34,7 +34,7 @@ export const riToast = (
     }
     toastContent.message = (
       <ColorText color={color}>
-        <Title size="M">{message}</Title>
+        <Title size="XS">{message}</Title>
         <Spacer size="m" />
       </ColorText>
     )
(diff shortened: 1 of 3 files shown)


pawelangelow (approved):

redisinsight/ui/src/components/base/display/toast/RiToast.tsx
```diff
@@ -37,7 +34,8 @@ export const riToast = (
     }
     toastContent.message = (
       <ColorText color={color}>
-        <StyledMessage>{message}</StyledMessage>
+        <Title size="XS">{message}</Title>
```
pd-redis: this puts H5 within a span, not good

pd-redis: Why is the need of a h tag and spacer?
Why not change StyledMessage to style a tag and set appropriate space?

dantovska: Ok, changing it. Tnx!

dantovska: as we discussed, in redis ui spaces are removed from h tags... so I need to add additional. Other than that, since there are two components that can do a bold title and a space, for me there is no difference which approach we use, but I think if we don't need additional style and it is used only in one place, we don't need a styled component ...

valkirilov (approved):

PR #4921, opened 2025-09-04
RI-7275 Create e2e tests for "Vector Search" page

valkirilov:
# Description

Created e2e tests to verify the availability of the "**Vector Search**" page:
- it should open the page when all requirements are met
- it should open a fallback message when the Redis Query Engine module is not enabled
- it should show a fallback message when Redis version doesn't support vector sets

## How the tests work

- In each test case, we have a setup step that initializes the environment and prepares a Redis instance suitable for it. Then, we open the correct page (navigate through the UI) until we reach it.

# How to run the tests

```
# From the root directory
yarn dev:api

yarn dev:ui

yarn test:chromium:local-web vector-search/version-compatibility

yarn playwright show-report
``` 

image


ArtemHoruzhenko (approved):
lgtm

PR #4924, opened 2025-09-05
RI-7131: fix failing e2e tests

ArtemHoruzhenko:


.github/workflows/tests-e2e-docker-critical-path.yml
```diff
@@ -0,0 +1,137 @@
...
+          up --abort-on-container-exit --force-recreate
+
+      - name: Upload Test Report
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: ${{ env.REPORT_NAME }}-${{ matrix.parallel }}
+          path: /usr/src/app/report
+
+      - name: Send report to Slack
```
valkirilov, line 121: I know you copied/pasted this when you split the workflow by categories, but are you aware of any Slack channel where such messages were sent? Or is it a legacy that was deprecated long ago?

.github/workflows/tests-e2e.yml
valkirilov, line 1: Note: Not related to your changes, just posting this as something we should think about in the future. 

Currently, we have [tests.yml](https://github.com/redis/RedisInsight/blob/main/.github/workflows/tests.yml) that is used to run the different types of tests (frontend, backend, integration), but doesn't reference the e2e tests in any way. I think it might be useful if we link up this workflow as well, as a separate step, so we end up with one centralized place where all the types of tests are configured, and run a specific type only when necessary (as it's done now). 

But again, I'm bringing up this for a discussion only, we don't have to apply any changes now.

tests/e2e/helpers/conf.ts
```diff
@@ -50,22 +50,6 @@ export const ossStandaloneV7Config = {
     databasePassword: process.env.OSS_STANDALONE_V7_PASSWORD
 };
 
-export const ossStandaloneV6Config = {
```
valkirilov, line 53: Why did you remove this config? 

_PS: I think we use it in the playwright tests, so did you remove the respective Docker container, or was the only cleanup you performed on this config file?_

tests/e2e/pageObjects/components/common/toast.ts
```diff
@@ -1,11 +1,12 @@
...
-    toastCloseButton = Selector('[data-test-subj=toastCloseButton]');
-    toastSubmitBtn = Selector('[data-testid=submit-tooltip-btn]');
-    toastCancelBtn = Selector('[data-testid=toast-cancel-btn]');
+    toastHeader = Selector('[data-testid=redisui-toast-message]', { timeout: 30000 });
+    toastBody = Selector('[data-testid=redisui-toast-description]');
+    toastSuccess = Selector('[data-testid=redisui-toast]');
+    toastError = Selector('[data-testid=toast-error]', { timeout: 30000 });
+    toastCloseButton = Selector('[data-testid=redisui-toast-action-button]');
+    toastSubmitBtn = Selector('[data-testid=redisui-toast-action-button]');
+    // todo: investigate. there is no cancel button in toast
```
valkirilov, line 10: Note: It depends on the props we pass to Redis UI, but in theory, we can have both an action button (OK) and a close icon.

Example:
- https://redislabsdev.github.io/redis-ui/?path=/story/components-toast--actions

PS: Leave it as it is, no need to apply any other changes now.

.github/workflows/tests-e2e-docker-critical-path.yml
ArtemHoruzhenko: We had it for sure but I assume we don't have it now. Let see, if it is needed - we will add channel, if not - we can remove it

tests/e2e/helpers/conf.ts
ArtemHoruzhenko: it was pointing to Redis v8 (weird), so I removed it since by default we use redis v8 in oss-standalone.

tests/e2e/pageObjects/components/common/toast.ts
ArtemHoruzhenko: Correct, was just fixing with selectors I had. We will rework all of these while moving to playwright. Didn't want to waste time on it. Ideally all such selectors must have testid

tests/e2e/helpers/conf.ts
valkirilov: Yeah, I've noticed the mismatch between the versions. So maybe I should apply the same change for playwright then? 

But are you sure it points to Redis v8, because what I see for playwright seems to be wrong - it's pointing to Redis instance on port 8100, which uses Redis v6. (maybe it's wrong only for playwright, but ok for test cafe)

- https://github.com/redis/RedisInsight/blob/4de818057c3f18cfcf728823666eb6ab1208937f/tests/playwright/env/.local-web.env

ArtemHoruzhenko: I tihnk yes. Let's use only redis v8 for now. later on when we will move tests to playwright we will add only necessary RTEs. Also we will review tests that are using not v8 Redis and decide if we can combine it with existing RTE instead of having tons of them

PR #4831, opened 2025-08-08
RI-7298: Rework delete confirmation

pawelangelow:
https://github.com/user-attachments/assets/79be1a2e-d2c1-4fad-9e0c-20ba21487e08

PR #4925, opened 2025-09-05
RI-7275 Create e2e tests for "Manage Indexes" panel on the "Vector Search" page

valkirilov:
# Description

Created e2e tests to verify the "**Manage Indexes**" panel on the "**Vector Search**" page:
- it should show a fallback message when there are no indexes
- it should show details about the indexes (both HASH and JSON)
- it should delete an index

## How the tests work

- We have a setup step that initializes the environment and prepares a Redis instance suitable for our tests. Then, we open the correct page (navigate through the UI until we reach it) and prepare the necessary steps required by the tests to run (create an index via the API).

# How to run the tests

```
# From the root directory
yarn dev:api

yarn dev:ui

yarn test:chromium:local-web vector-search/manage-indexes

yarn playwright show-report
``` 

image

### DIFF
--- a/tests/playwright/pageObjects/pages/vector-search/vector-search-page.ts
+++ b/tests/playwright/pageObjects/pages/vector-search/vector-search-page.ts
@@ -224,14 +224,9 @@ export class VectorSearchPage extends BasePage {
         expectedMessage: string,
         timeout = 2000,
     ): Promise<void> {
-        try {
-            await this.waitForLocatorVisible(this.toast.toastSuccess, timeout)
-            await expect(this.toast.toastBody).toContainText(expectedMessage)
-            await this.toast.closeToast()
-        } catch {
-            // No toast appeared - this is acceptable for some actions
-            // Success is typically verified by other means (navigation, etc.)
-        }
+        await this.waitForLocatorVisible(this.toast.toastContainer, timeout)
+        await expect(this.toast.toastMessage).toContainText(expectedMessage)
+        await this.toast.closeToast()
     }
 
     async expandIndexDetails(indexName: string): Promise<void> {
--- a/tests/playwright/pageObjects/components/common/toast.ts
+++ b/tests/playwright/pageObjects/components/common/toast.ts
@@ -3,6 +3,8 @@ import { BasePage } from '../../base-page'
 import { ToastSelectors } from '../../../selectors'
 
 export class Toast extends BasePage {
+    // Deprecated - use new toast selectors below
+    // TODO: Remove deprecated selectors and usages after migrating all toasts
     public readonly toastHeader: Locator
 
     public readonly toastBody: Locator
@@ -17,6 +19,15 @@ export class Toast extends BasePage {
 
     public readonly toastCancelBtn: Locator
 
+    // New toast selectors
+    public readonly toastContainer: Locator
+
+    public readonly toastMessage: Locator
+
+    public readonly toastDescription: Locator
+
+    public readonly toastActionButton: Locator
+
     constructor(page: Page) {
         super(page)
         this.toastHeader = page.locator(ToastSelectors.toastHeader)
@@ -26,6 +37,12 @@ export class Toast extends BasePage {
         this.toastCloseButton = page.locator(ToastSelectors.toastCloseButton)
         this.toastSubmitBtn = page.getByTestId(ToastSelectors.toastSubmitBtn)
         this.toastCancelBtn = page.getByTestId(ToastSelectors.toastCancelBtn)
+
+        // New toast selectors
+        this.toastContainer = page.locator(ToastSelectors.toastContainer)
+        this.toastMessage = page.locator(ToastSelectors.toastMessage)
+        this.toastDescription = page.locator(ToastSelectors.toastDescription)
+        this.toastActionButton = page.locator(ToastSelectors.toastActionButton)
     }
 
     async isCloseButtonVisible(): Promise<boolean> {
@@ -33,6 +50,6 @@ export class Toast extends BasePage {
     }
 
     async closeToast(): Promise<void> {
-        await this.toastCloseButton.click()
+        await this.toastActionButton.click()
     }
 }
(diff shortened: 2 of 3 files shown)


tests/playwright/helpers/conf.ts
```diff
@@ -20,61 +21,62 @@ export const workingDirectory =
...
     databaseUsername: process.env.OSS_STANDALONE_V7_USERNAME,
     databasePassword: process.env.OSS_STANDALONE_V7_PASSWORD,
 }
 
-export const ossStandaloneV6Config = {
+// TODO: Rename this, please
+export const ossStandaloneV6Config: AddNewDatabaseParameters = {
     host: process.env.OSS_STANDALONE_V8_HOST,
     port: process.env.OSS_STANDALONE_V8_PORT,
     databaseName: `${process.env.OSS_STANDALONE_V8_DATABASE_NAME || 'test_standalone-v6'}-${uniqueId}`,
```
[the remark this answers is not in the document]
valkirilov: Thanks, but I don't want to make this change as part of this pull request. Updating this name in many places will simply flood the PR with changes, distracting the developers during the code review. We'll resolve it in a separate PR.

tests/playwright/factories/redisearch-index.factory.ts
ArtemHoruzhenko, line 1: I think I saw this in another PR, why it is included here too?

valkirilov: I have cherry-picked the commit, so I can reuse it in two separate PRs. Once the first is merged, I'll rebase the branch.

tests/playwright/pageObjects/pages/vector-search/vector-search-page.ts
```diff
@@ -113,4 +148,90 @@ export class VectorSearchPage extends BasePage {
...
+    ): Promise<void> {
+        try {
+            await this.waitForLocatorVisible(this.toast.toastSuccess, timeout)
+            await expect(this.toast.toastBody).toContainText(expectedMessage)
+            await this.toast.closeToast()
+        } catch {
+            // No toast appeared - this is acceptable for some actions
+            // Success is typically verified by other means (navigation, etc.)Expand commentComment on lines R179 to R181ResolvedCode has comments. Press enter to view.
+        }
+    }
```
ArtemHoruzhenko: not sure I understand properly. If there is no toast by design why should we still wait for 2s (by default)? 
It is not critical but in general 2s here, 2s there and we observe that our tests are running for hours...

tests/playwright/pageObjects/pages/vector-search/vector-search-page.ts
```diff
@@ -113,4 +148,90 @@ export class VectorSearchPage extends BasePage {
...
+        )
+        await this.waitForLocatorNotVisible(
+            this.manageIndexesContainer.getByText('Type'),
+        )
+        await this.waitForLocatorNotVisible(
+            this.manageIndexesContainer.getByText('Weight'),
+        )
+        await this.waitForLocatorNotVisible(
+            this.manageIndexesContainer.getByText('Noindex'),
+        )
```
ArtemHoruzhenko: not sure why we need this? 
It looks like 
```        
await this.waitForLocatorNotVisible(this.manageIndexesIndexDetails)
await this.waitForLocatorVisible(this.manageIndexesIndexCollapsedInfo)
```
is enough, no?

redisinsight/ui/src/pages/vector-search/manage-indexes/IndexSection.tsx
```diff
@@ -71,22 +71,21 @@ export const IndexSection = ({ index, ...rest }: IndexSectionProps) => {
   return (
     <Section.Compose
       collapsible
-      collapsedInfo={<CategoryValueList categoryValueList={indexSummaryInfo} />}
-      content={<IndexAttributesList indexInfo={indexInfo} />}
```
ArtemHoruzhenko, line 75: are you sure you wanted to delete this?

redisinsight/ui/src/pages/vector-search/manage-indexes/IndexSection.tsx
```diff
@@ -71,22 +71,21 @@ export const IndexSection = ({ index, ...rest }: IndexSectionProps) => {
...
-      // TODO: Add FieldTag component to list the types of the different fields
       defaultOpen={false}
       onOpenChange={handleOpenChange}
-      onAction={() => setIsPopoverOpen(true)}
```
ArtemHoruzhenko, line 79: and this

redisinsight/ui/src/pages/vector-search/manage-indexes/IndexSection.tsx
valkirilov: Yes, it was intentional. We moved to a composable component in https://github.com/redis/RedisInsight/pull/4831, so these props are no longer used, and we can remove them to avoid future confusion.

tests/playwright/pageObjects/pages/vector-search/vector-search-page.ts
valkirilov: Actually, this is a nice catch. But not because of the timeouts,I have actually dug deeper into why we needed the try-catch block initially, and I found that these helper methods weren’t validating anything. 

Since the migration of EUI, behind the scenes these checks rely on the EUI class names, that are no longer available. So I went to refactor this helper method, and now it should behave properly. Thank you!

tests/playwright/pageObjects/pages/vector-search/vector-search-page.ts
valkirilov: You’re right, I agree with you. In the first step, I want to ensure that the table is visible. So, we go and verify the columns, and that’s fine. 

However, here, we want to check that the information is collapsed and gone, so there’s no point in doing these checks. I've removed it, thanks for pointing it out.

Issue #4927, opened 2025-09-05
[Bug]: Database list search not working for not connected databases, database still visible after search


Matiszak:
**Preconditions** (Any important steps we need to know)

Have a database which is in "NOT CONNECTED" state (e.g. I'm importing it from a file)

**Steps to reproduce** (How to reproduce what you found step by step)

1. Import database from a file
2. On main page in "Database List Search" search for a string that should show **not** show database above (e.g. database with name "redis_abc" and put in searchbox "test")

**Actual behavior** (A short description of what you found)

Database is still visible on the databases list

**Expected behavior** (A short description of what you expected to find)

Database is not visible on the databases list

**Screenshots** (Paste or drag-and-drop a screenshot or a link to a recording)

<img width="2335" height="1471" alt="Image" src="https://github.com/user-attachments/assets/14d29609-9f89-4012-bee9-4e916394a77b" />

**Additional context** (Operating system, version of Redis Insight, Redis database version, Redis module version, database type, connection type, logs, or any other information)

From my debugging it's caused by this line: https://github.com/redis/RedisInsight/blob/93a7404da53726c1c5266b478bdf9c5f256609ac/redisinsight/ui/src/pages/home/components/search-databases-list/SearchDatabasesList.tsx#L47-L50

During my debugging variables were:
```
item.connectionType = "NOT CONNECTED"
CONNECTION_TYPE_DISPLAY = {
    "STANDALONE": "Standalone",
    "CLUSTER": "OSS Cluster",
    "SENTINEL": "Sentinel"
}
```
This causes
`CONNECTION_TYPE_DISPLAY[item.connectionType]` evaluation to undefined
And overall
```
item.connectionType &&
              CONNECTION_TYPE_DISPLAY[item.connectionType]
                ?.toLowerCase()
                ?.indexOf(value) !== -1
```
to be true

Which causes database to be shown on the list despite fact that it does not contain any value containing searched string

valkirilov:
@Matiszak, thank you for your contribution. Now your changes are merged to the `main` branch, and the fix will be included in the next official release of Redis Insight.

valkirilov:
@Matiszak, the issue should be resolved now, and you can verify it as part of the latest release of Redis Insight [v3.0.2](https://github.com/redis/RedisInsight/releases/tag/3.0.2). 

I'll close the issue for now, but feel free to reopen it in case you encounter any other issues related to that. Thanks for your feedback!

[...]

PR #4928, opened 2025-09-06
#4927 Fix search database on list not filtering out databases that ar…

Matiszak:
Fix search database on list not filtering out databases that are not connected. Refactor tests for better understanding.


valkirilov:
@Matiszak, thank you for taking the time to contribute ✌️

PR #4930, opened 2025-09-08
RI-7401 Fix broken key name layout

ArtemHoruzhenko:
Fix small layout issue for key details header

**On hover**
| Before | After |
| - | - |
Screenshot 2025-09-09 at 09 04 33|Screenshot 2025-09-09 at 09 06 44

Size, length, TTL on hover
| Before | After |
| - | - |
Screenshot 2025-09-09 at 09 04 39|Screenshot 2025-09-09 at 09 07 01

### DIFF
--- a/redisinsight/ui/src/pages/browser/modules/key-details-header/KeyDetailsHeader.tsx
+++ b/redisinsight/ui/src/pages/browser/modules/key-details-header/KeyDetailsHeader.tsx
@@ -166,7 +166,12 @@ const KeyDetailsHeader = ({
                   )}
                 </FlexItem>
               </Row>
-              <Row centered className={styles.groupSecondLine} gap="m">
+              <Row
+                centered
+                className={styles.groupSecondLine}
+                gap="m"
+                align="start"
+              >
                 <KeyDetailsHeaderSizeLength width={width} />
                 <KeyDetailsHeaderTTL onEditTTL={handleEditTTL} />
                 <FlexItem grow>
--- a/redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-name/KeyDetailsHeaderName.tsx
+++ b/redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-name/KeyDetailsHeaderName.tsx
@@ -26,7 +26,7 @@ import {
   stringToBuffer,
 } from 'uiSrc/utils'
 
-import { FlexItem } from 'uiSrc/components/base/layout/flex'
+import {FlexItem, Row} from 'uiSrc/components/base/layout/flex'
 import { IconButton } from 'uiSrc/components/base/forms/buttons'
 import { CopyIcon } from 'uiSrc/components/base/icons'
 import { Text } from 'uiSrc/components/base/text'
@@ -34,6 +34,7 @@ import { RiIcon } from 'uiSrc/components/base/icons/RiIcon'
 import { RiTooltip } from 'uiSrc/components'
 import { TextInput } from 'uiSrc/components/base/inputs'
 import styles from './styles.module.scss'
+import styled from 'styled-components'
 
 export interface Props {
   onEditKey: (
@@ -140,6 +141,10 @@ const KeyDetailsHeaderName = ({ onEditKey }: Props) => {
     })
   }
 
+  const StyledInputWrapper = styled(Row)`
+    min-width: 150px;
+  `
+
   return (
     <FlexItem
       grow
@@ -169,7 +174,9 @@ const KeyDetailsHeaderName = ({ onEditKey }: Props) => {
             isLoading={loading}
             declineOnUnmount={false}
           >
-            <div className={styles.inputWithIcon}>
+            <StyledInputWrapper
+              align="center"
+            >
               <TextInput
                 name="key"
                 id="key"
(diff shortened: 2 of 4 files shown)


redisinsight/ui/src/pages/browser/modules/key-details-header/styles.module.scss
```diff
@@ -62,6 +62,7 @@
 
 .groupSecondLine {
   margin-top: 4px !important;
+  align-items: flex-start !important;
```
dantovska: Instead of adding this style, you can set align="start" to the Row tag where this class is added.  Now that I'm looking at it, you can entirely remove this class and instead of margin-top, you can adda spacer with size S or XS

redisinsight/ui/src/components/base/tooltip/RITooltip.tsx
```diff
@@ -11,6 +12,10 @@ export interface RiTooltipProps
   anchorClassName?: string
 }
 
+const StyledTooltip = styled(Tooltip)`
+  word-break: break-all;
+`
+
```
dantovska, line 18: Are we sure we want this behaviour applied to all tooltips? This will change all tooltips behavior and may break some use cases... I know we discussed not to use classes, but maybe this would be the best approach here - the className could be applied to the same Tag you are replacing.

redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-name/KeyDetailsHeaderName.tsx
```diff
@@ -155,84 +153,80 @@ const KeyDetailsHeaderName = ({ onEditKey }: Props) => {
...
+          <InlineItemEditor
+            onApply={() => applyEditKey()}
+            isDisabled={!keyIsEditable}
+            disabledTooltipText={TEXT_UNPRINTABLE_CHARACTERS}
+            onDecline={(event) => cancelEditKey(event)}
+            viewChildrenMode={!keyIsEditing}
+            isLoading={loading}
+            declineOnUnmount={false}
+          >
+            <div className={styles.inputWithIcon}>
```
dantovska: you can change this div and class with Row, and apply the width only (if needed?)

redisinsight/ui/src/components/base/tooltip/RITooltip.tsx
ArtemHoruzhenko: in my opinion - yes, we want. I don't see any case where we don't want to move text to the next line in a tooltip

redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-name/KeyDetailsHeaderName.tsx
ArtemHoruzhenko: @dantovska I've added several styled components but at the same time I had to add more css rules in existing scss files. In general I do think that moving all into styled components is not part of current task. It is more about refactoring in the future

redisinsight/ui/src/components/base/tooltip/RITooltip.tsx
ArtemHoruzhenko: Here is the fix. Now we will always truncate name and just show icons after on hover. 
Screenshot 2025-09-10 at 09 41 12

redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-name/styles.module.scss
```diff
@@ -8,21 +8,19 @@
...
-  align-items: center;
-  min-width: 150px;
-}
-
 .keyInput {
   height: 31px !important;
   font-size: 14px !important;
   font-weight: 500 !important;
   flex: 1;
+  max-width: 400px;
```
dantovska, line 20: I see you have hardcoded 450px and 400 in most places, because we do not want to overflow the container. I'm wondering if there is another way to solve this? For example, when it is in full screen mode, the text is still truncated but there is a space for it to be displayed inside the container.

redisinsight/ui/src/pages/browser/modules/key-details-header/KeyDetailsHeader.tsx
```diff
@@ -166,7 +166,12 @@ const KeyDetailsHeader = ({
                   )}
                 </FlexItem>
               </Row>
-              <Row centered className={styles.groupSecondLine} gap="m">
+              <Row
+                centered
+                className={styles.groupSecondLine}
```
dantovska, line 171: maybe we can remove this class which does nothing but margin and add spacer instead?

ArtemHoruzhenko: May be we can. I didn't check and think it must be part of another task.

redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-name/styles.module.scss
ArtemHoruzhenko: Yes, we discussed this and confirmed that it is good for now

dantovska (approved):

PR #4931, opened 2025-09-08
RI-7395: fix state issue causing rdi delete popover wasn't showing

ArtemHoruzhenko:


dantovska (approved):
Looks good! 
@valkirilov introduced a new way of mocking data using faker which I think looks good and whenever there is a new test file created and mocked data needed, this is a good candidate. I've done it here - https://github.com/redis/RedisInsight/pull/4868/files#diff-d18705a00f4888ff0f7e7c9219cd30b877eb09dfd6ec7899cc7f6bc7133548b1 
Just an idea, if you want to consider it here.

redisinsight/ui/src/pages/rdi/home/instance-list/RdiInstancesListWrapper.real.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+import RdiInstancesListWrapper from './RdiInstancesListWrapper'
+
+jest.mock('uiSrc/slices/rdi/instances', () => ({
+  ...jest.requireActual('uiSrc/slices/rdi/instances'),
+  instancesSelector: jest.fn(),
+}))
+
+const mockInstances: RdiInstance[] = [
+  {
+    id: '1',
```
valkirilov: > Looks good! @valkirilov introduced a new way of mocking data using faker which I think looks good and whenever there is a new test file created and mocked data needed, this is a good candidate. I've done it here - https://github.com/redis/RedisInsight/pull/4868/files#diff-d18705a00f4888ff0f7e7c9219cd30b877eb09dfd6ec7899cc7f6bc7133548b1 Just an idea, if you want to consider it here.

Agree, this looks like a perfect candidate for a reusable factory:
- it will make the test suite simpler, hiding the unnecessary business logic related to the DB Connection Info object structure
- it will provide a simple and reusable way to apply the same concept to various unit tests.
- in case we change the interface of the object, it will be automatically inherited when we update the factory, without the need to manually update it here in the test suite.

More Examples:

dantovska (approved):

PR #4937, opened 2025-09-09
RI-7275 Create e2e tests for "Saved Queries" panel on the "Vector Search" page

valkirilov:
# Description

Created e2e tests to verify the "**Saved Queries**" panel on the "**Vector Search**" page:
- it should show a fallback message when there are no indexes
- it should show details about the saved queries of the preset index
- it should insert saved queries in the query editor

## How the tests work

- We have a setup step that initializes the environment and prepares a Redis instance suitable for our tests. Then, we open the correct page (navigate through the UI until we reach it) and prepare the necessary steps required by the tests to run (create an index via the API).

# How to run the tests

```
# From the root directory
yarn dev:api

yarn dev:ui

yarn test:chromium:local-web vector-search/saved-queries

yarn playwright show-report
``` 

image

### DIFF
--- a/tests/playwright/tests/vector-search/saved-queries.spec.ts
+++ b/tests/playwright/tests/vector-search/saved-queries.spec.ts
@@ -1,4 +1,3 @@
-import { faker } from '@faker-js/faker'
 import { PresetDataType } from 'uiSrc/pages/vector-search/create-index/types'
 import { VectorSearchPage } from '../../pageObjects/pages/vector-search/vector-search-page'
 import { test, expect } from '../../fixtures/test'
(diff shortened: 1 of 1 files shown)


tests/playwright/helpers/api/api-indexes.ts
```diff
@@ -0,0 +1,94 @@
...
+    async createRedisearchIndexApi(
+        indexParameters: CreateRedisearchIndexParameters,
+        databaseParameters: AddNewDatabaseParameters,
+    ): Promise<void> {
+        const databaseId = await this.databaseAPIRequests.getDatabaseIdByName(
+            databaseParameters.databaseName,
+        )
+
+        const requestBody = {
+            index: stringToBuffer(indexParameters.indexName),
```
ArtemHoruzhenko, line 29: why do we need to convert string to buffer?

valkirilov: This is the pattern used so far in the other helper utils that communicate with the api, so I just followed it:

https://github.com/redis/RedisInsight/blob/896cf80b51a1cb11d327964b823622a65a184af6/tests/playwright/helpers/api/api-keys.ts#L27

To be honest, I'm not sure what the actual reason behind this is, but my understanding was that we convert everything to buffers in case there are big chunks of data, so we can stream them in a memory-efficient way.

ArtemHoruzhenko: Nope, this functionality is to send binary data outside of utf8 range. I think if we are not going to use it - we don't need it here. But no need to change anything. Let's not add it for next helpers

PR #4946, opened 2025-09-11
RI-7397: Update Create Index screen

dantovska:
Making the following changes to the Create Index Page:
1. Change title Vector Index -> Create Index
2. Remove the Button Group selection Build new index / Use Preset Index. 
3. Add a banner when data is pre-set explaining why the index editing is disabled. Banner text: “Editing the index is not available for pre-set data. Click Create Index to continue.
4. Remove the inner borders around Create Index

Before:
image

After:
image

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.spec.tsx
+++ b/redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.spec.tsx
@@ -199,4 +199,18 @@ describe('VectorSearchCreateIndex', () => {
       })
     })
   })
+
+  it('should show disabled data editing banner when on final step with preset data', () => {
+    renderVectorSearchCreateIndexComponent({ initialStep: 2 })
+
+    const banner = screen.getByTestId('disabled-data-editing-banner')
+    expect(banner).toBeInTheDocument()
+  })
+
+  it('should not show disabled data editing banner when not on final step', () => {
+    renderVectorSearchCreateIndexComponent({ initialStep: 1 })
+
+    const banner = screen.queryByTestId('disabled-data-editing-banner')
+    expect(banner).not.toBeInTheDocument()
+  })
 })
--- a/redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.tsx
+++ b/redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.tsx
@@ -56,7 +56,7 @@ export type VectorSearchCreateIndexProps = {
   initialStep?: number
 }
 
-const DisabledBannerText = () => (
+const DisabledDataEditingBannerText = () => (
   <Text component="div">
     Editing the index is not available for pre-set data. Click{' '}
     <Text variant="semiBold" component="span">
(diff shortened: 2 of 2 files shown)


redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.spec.tsx
```diff
@@ -183,11 +183,11 @@ describe('VectorSearchCreateIndex', () => {
       renderVectorSearchCreateIndexComponent()
 
       // Simulate going to the index info step
-      const buttonNext = screen.getByText('Proceed to index')
+      const buttonNext = screen.getByTestId('proceed-to-index-button')
```
valkirilov, line 186: nit: I'm not sure whether you have a convention over that in the project, to always use test IDs, but actually React Testing Library recommends relying on them as the last resort. 

According to their guidelines, they suggest writing the tests to resemble the way the users interact with the UI as closely as possible. And in this case, it's easier to select these items via their text representation. 

> Based on [the Guiding Principles](https://testing-library.com/docs/guiding-principles), your test should resemble how users interact with your code (component, page, etc.) as much as possible. With this in mind, we recommend this order of priority.

> getByTestId: The user cannot see (or hear) these, so this is only recommended for cases where you can't match by role or text or it doesn't make sense (e.g. the text is dynamic).

Reference: https://testing-library.com/docs/queries/about/#priority

_PS: Yes, I know sometimes it gets annoying when you have to update the test every time you update the labels._

redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.tsx
```diff
@@ -34,24 +35,45 @@ import {
...
+    testId: 'create-index-button',
+  },
 ]
 
 export type VectorSearchCreateIndexProps = {
   initialStep?: number
 }
 
+const DisabledBannerText = () => (
+  <Text component="div">
```
valkirilov, line 60: Note: Maybe we can add a test to verify the presence of this banner. Now it may seem meaningless, since this is the only flow and it's always visible, but in the next iterations, we’re going to add different types of data (not only preset) and we might show this conditionally, so maybe it won't harm us if we have a test to verify that.

valkirilov (commented):
Nice! You can also check whether the e2e tests for this page still pass successfully after the UI changes you made.

```
# From the root directory
yarn dev:api

yarn dev:ui

yarn test:chromium:local-web vector-search/create-index
```

redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.spec.tsx
dantovska: I didn't mind using the text at all, just it made sense to change it in this test because I changed the create-index-button and I kind of wanted to be consistent. The reason I changed the create-index-button was because changing the title to Create index made the getByText to return two "Create index" elements and I didn't want to introduce some kind of additional logic, when I can use ID. 
Didn't know about the priorities though, tnx!

valkirilov: No worries, it's not a problem to select it via test IDs as well. Some teams rely only on test IDs, so if you have such a convention internally, it's fine. It's minor, after all 🙂

redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.tsx
dantovska: Added the test. Since we don't have a form to choose preset / custom data and sampleDataType has a hardcoded value, I only considered the final step for the test. I suppose when we introduce this functionality, this test will fail and we will be forced to enhance it with the sampleDataType setup.

PR #4957, opened 2025-09-15
RI-6855 Added title for the new vector search page

valkirilov:
# Description

Added a title for the new **Vector Search** page, following the pattern described by the rest of the available pages (Browse, Workbench, etc.)

image


redisinsight/ui/src/pages/vector-search/query/HeaderActions.spec.tsx
```diff
@@ -3,13 +3,6 @@ import { cleanup, render, screen, userEvent } from 'uiSrc/utils/test-utils'
 import { HeaderActions, HeaderActionsProps } from './HeaderActions'
 import useRedisInstanceCompatibility from '../create-index/hooks/useRedisInstanceCompatibility'
 
-// Workaround for @redis-ui/components Title component issue with react-children-utilities
```
valkirilov, line 6: Note: This is not related to the change of the title, but removing this, since it was reported that the issue was resolved by removing a custom mock, made on a global level a few years ago.

PR #4959, opened 2025-09-16
Fe/feature/ri 7431 keys header columns popover

ArtemHoruzhenko:
|Before|After|
|-|-|
|Screenshot 2025-09-16 at 15 17 08|Screenshot 2025-09-16 at 15 16 49|
|Screenshot 2025-09-16 at 15 17 02|Screenshot 2025-09-16 at 15 16 39|

### DIFF
--- a/redisinsight/ui/src/components/base/forms/buttons/ToggleButton.tsx
+++ b/redisinsight/ui/src/components/base/forms/buttons/ToggleButton.tsx
@@ -1,8 +1,1 @@
-import React from 'react'
-import { ToggleButton as RedisUiToggleButton } from '@redis-ui/components'
-
-export type ButtonProps = React.ComponentProps<typeof RedisUiToggleButton>
-
-export const ToggleButton = ({ ...props }: ButtonProps) => {
-  return <RedisUiToggleButton {...props} />
-}
+export { ToggleButton } from '@redis-ui/components'


redisinsight/ui/src/components/base/forms/combo-box/AutoTag.tsx
```diff
@@ -210,7 +218,7 @@ const StyledWrapper = styled(Row)`
   position: relative;
   border: 1px solid ${({ theme }) => theme.semantic.color.border.neutral600};
   border-radius: 0.4rem;
-  padding: 0.15rem 0.5rem;
+  padding: 0 0.5rem;
```
dantovska: please use the theme spacing values

redisinsight/ui/src/components/base/forms/combo-box/AutoTag.tsx
```diff
@@ -128,6 +128,13 @@ export const AutoTag = ({
     }
   }
 
+  const handleBlur: React.FocusEventHandler<HTMLInputElement> = (e) => {
```
dantovska, line 131: this method looks like and does the same thing as handleEnter. Can we unify them?

redisinsight/ui/src/pages/browser/components/keys-header/KeysHeader.tsx
```diff
@@ -50,15 +44,15 @@ import { RiPopover, RiTooltip } from 'uiSrc/components/base'
...
+import { Col, FlexItem, Row } from 'uiSrc/components/base/layout/flex'
 import { setConnectivityError } from 'uiSrc/slices/app/connectivity'
 import { Checkbox } from 'uiSrc/components/base/forms/checkbox/Checkbox'
 import { RiIcon } from 'uiSrc/components/base/icons/RiIcon'
 import styles from './styles.module.scss'
 import { ButtonGroup } from 'uiSrc/components/base/forms/button-group/ButtonGroup'
 import styled from 'styled-components'
-import {TextButton} from '@redis-ui/components'
-import {Text} from 'uiSrc/components/base/text'
+import { ToggleButton } from '@redis-ui/components'
```
dantovska: You should not import directly from `@redis-ui/components`.
If you want to use some component, you need to import it through a wrapper file. The file can extend the functionality or do nothing at all in some cases, but still, should be used as a proxy. Look for `ui/src/components/base/display/banner/index.ts` as a reference.

redisinsight/ui/src/components/base/forms/combo-box/AutoTag.tsx
ArtemHoruzhenko: not, actually. one is for key enter where we are checking if "enter" was pressed when this one happens on blur

redisinsight/ui/src/components/base/forms/combo-box/AutoTag.tsx
ArtemHoruzhenko: Sorry, didn't get this

dantovska: instead of 0.5rem, we should use the following:

`${({ theme }) => theme.core.space.space050}`

space050 is actually 0.4rem, but I guess it should work. 

It could be something like:
```
padding-block: 0;
padding-inline: ${({ theme }) => theme.core.space.space050};
```

redisinsight/ui/src/components/base/forms/buttons/ToggleButton.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+import { ToggleButton as RedisUiToggleButton } from '@redis-ui/components'
+
+export type ButtonProps = React.ComponentProps<typeof RedisUiToggleButton>
+
+export const ToggleButton = ({ ...props }: ButtonProps) => {
+  return <RedisUiToggleButton {...props} />
+}
```
dantovska:
```
import { ToggleButton } from '@redis-ui/components'

export { ToggleButton }
```

and then export it from the buttons index.ts

ArtemHoruzhenko: I did it as it was done for another component that doesn't add anything new itself and thought that it is the pattern. But I'm fine to do it this way or even shorter

PR #4960, opened 2025-09-16
RI-7431: Change add key button view

ArtemHoruzhenko:
| Before | After |
| - | - |
|Screenshot 2025-09-16 at 17 24 53|Screenshot 2025-09-16 at 17 24 38|

### DIFF
--- a/redisinsight/ui/src/pages/browser/components/actions/Actions.tsx
+++ b/redisinsight/ui/src/pages/browser/components/actions/Actions.tsx
@@ -6,7 +6,6 @@ import {
   TelemetryEvent,
 } from 'uiSrc/telemetry'
 import {
-  Button,
   SecondaryButton,
 } from 'uiSrc/components/base/forms/buttons'
 import styles from 'uiSrc/pages/browser/components/browser-search-panel/styles.module.scss'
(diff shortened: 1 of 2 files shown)


pawelangelow (dismissed):

redisinsight/ui/src/pages/browser/components/actions/Actions.tsx
```diff
@@ -6,7 +6,7 @@ import {
   TelemetryEvent,
 } from 'uiSrc/telemetry'
 import {
-  PrimaryButton,
+  Button,
```
valkirilov: nit: We also have an alias for a `SecondaryButton` that you can use :)

valkirilov (dismissed):

redisinsight/ui/src/pages/browser/components/actions/Actions.tsx
ArtemHoruzhenko: Changed but I didn't like it. In general we added custom props to change the behavior and it became not clear which exactly variant we use.

valkirilov: Usually, the goal of having semantics in the way we build up our design systems is to hide the "implementation details" and make the end result look consistent everywhere. 

Yes, we can always use the base components and customize them however we want, but I believe this is what we wanted to avoid when @pd-redis's introduced these semantic aliases. Anyway, we can talk about this in the future, it's not something critical/important right now :)

valkirilov (approved):

PR #4962, opened 2025-09-17
RI-7431: Delete key popover layout

ArtemHoruzhenko:
Fixed behavior when clicking on trash button results into selecting row and fetching key details
Also fixed popover layout

|Before|After|
|-|-|
Screenshot 2025-09-17 at 15 19 55|Screenshot 2025-09-17 at 15 20 34

### DIFF
--- a/redisinsight/ui/src/pages/browser/components/delete-key-popover/DeleteKeyPopover.tsx
+++ b/redisinsight/ui/src/pages/browser/components/delete-key-popover/DeleteKeyPopover.tsx
@@ -1,17 +1,18 @@
 import React from 'react'
+import styled from 'styled-components'
 
 import cx from 'classnames'
 import { KeyTypes, ModulesKeyTypes } from 'uiSrc/constants'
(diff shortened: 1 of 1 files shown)


redisinsight/ui/src/pages/browser/components/delete-key-popover/DeleteKeyPopover.tsx
```diff
@@ -33,40 +39,46 @@ export const DeleteKeyPopover = ({
...
-        data-testid="submit-delete-key"
-      >
-        Delete
-      </DestructiveButton>
-    </>
-  </RiPopover>
-)
+}: DeleteProps) => {
+  const onClick = (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+    e.stopPropagation()
```
ArtemHoruzhenko, line 44: not sure if this is a proper approach but I've added stop propagation to avoid selecting row when you clicked on an element inside that row (trash icon in our case)

PR #4964, opened 2025-09-17
RI-7456 Hide "getting started" button when Redis version is not compatible

valkirilov:
# Description 

Hide the "**Getting Started**" button (leading to the "**Create Index**" wizard) from the "**No Results**", "**Manage Indexes**", and "**Saved Queries**"  panels on the **Vector Search** page when the Redis version is older than 7.2 and doesn't support Vector Sets.

image

# How it was tested

1. First, you need an older Redis instance without support for Vector Sets (older than 7.2)

_Note: You can use the following command to create a Docker container with such an instance_
```bash
docker run -d --name redis6-redismod -p 6404:6379 redislabs/redismod:latest
```

4. Go to **Databases** page and connect to the instance
5. Go to **Search** page (from the main navbar)

You'll see the search page, but with a banner to let you know that you can't use the full functionalities.

Respectively, the "**Getting started**" button will no longer be visible in the "**No Results**", "**Manage Indexes**" and "**Saved Queries**" panels.

| Before | After |
|-|-|
image|image
image|image

## Note

The "**Getting Started**" button is still there, when you use a newer Redis version that supports Vector Sets.

image

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/components/no-data-message/NoDataMessage.spec.tsx
+++ b/redisinsight/ui/src/pages/vector-search/components/no-data-message/NoDataMessage.spec.tsx
@@ -1,7 +1,12 @@
 import React from 'react'
-import { render, screen } from 'uiSrc/utils/test-utils'
+import { cleanup, render, screen } from 'uiSrc/utils/test-utils'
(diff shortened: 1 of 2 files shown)


dantovska (approved):
looks gr8! verified

PR #4967, opened 2025-09-17
RI-7370: Delimiter settings. Apply button fix

ArtemHoruzhenko:
Reworks fix for behavior when if you didn't press enter during delimiter creation and clicked apply. tag won't be saved.

### DIFF
--- a/redisinsight/ui/src/components/base/forms/combo-box/AutoTag.tsx
+++ b/redisinsight/ui/src/components/base/forms/combo-box/AutoTag.tsx
@@ -7,6 +7,21 @@ import { CommonProps, Theme } from 'uiSrc/components/base/theme/types'
 import { Row } from 'uiSrc/components/base/layout/flex'
 import { IconButton } from 'uiSrc/components/base/forms/buttons'
 
+const StyledWrapper = styled(Row)`
+  position: relative;
+  border: 1px solid ${({ theme }) => theme.semantic.color.border.neutral600};
+  border-radius: 0.4rem;
+  padding: ${({ theme }: { theme: Theme }) =>
+  `${theme.core.space.space000} ${theme.core.space.space050}`};
+  background-color: ${({ theme }) =>
+  theme.semantic.color.background.neutral100};
+`
+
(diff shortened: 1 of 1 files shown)


dantovska:
Do we want the delimiter to be saved when user did not click enter but clicked apply?

dantovska:
image

Also, when I click on 2, I cannot edit the input field, I need to click precisely on 1.

ArtemHoruzhenko:
> Do we want the delimiter to be saved when user did not click enter but clicked apply?
Right. That was the issue. Basically we repeat previous functionality here

> image
> Also, when I click on 2, I cannot edit the input field, I need to click precisely on 1.
Nice catch! Fixed
Screenshot 2025-09-18 at 11 13 58
Screenshot 2025-09-18 at 11 14 15
Screenshot 2025-09-18 at 11 14 26

redisinsight/ui/src/components/base/forms/combo-box/AutoTag.tsx
```diff
@@ -220,3 +221,11 @@ const StyledWrapper = styled(Row)`
...
     theme.semantic.color.background.neutral100};
 `
+
+const FlexibleInputWrapper = styled.div`
+  flex: 1;
+  min-width: 50px;
+  > * {
+    width: 100%;
+  }
+`
```
dantovska: Instead of creating a wrapper, you can add the styles to the input:
```suggestion
const StyledInput = styled(Input)`
  flex: 1;
  min-width: 30px;
`
```

Also, please define it at the top of the follow, not the bottom. Use it like this (almost same as now):
```
           <StyledInput
            variant="underline"
            placeholder={getPlaceholder()}
            onChange={handleInputChange}
            onKeyDown={handleEnter}
            value={tag}
            data-test-subj="autoTagInput"
          />
```

dantovska: Or 27px, whatever works as min-width, as discussed offline.

dantovska (approved):

PR #4970, opened 2025-09-18
RI-7431: rework autorefresh popover layout

ArtemHoruzhenko:
Rework AutoRefresh popover to align to the rest design

|Before|After|
|-|-|
Screenshot 2025-09-18 at 09 08 27|Screenshot 2025-09-18 at 09 07 53
Screenshot 2025-09-18 at 09 08 32|Screenshot 2025-09-18 at 09 07 58


dantovska (commented):
Maybe we should apply this to the AutoRefresh in mini-dash as well? 

image

ArtemHoruzhenko:
> Maybe we should apply this to the AutoRefresh in mini-dash as well?

It is applied

PR #4981, opened 2025-09-23
Fe/feature/ri 7546 get started page ff and show flag

ArtemHoruzhenko:
1. Disable onboarding (set appropriate local storage flag) after index created on vector search multi-step form
2. Reworked ff for vector search to hide onboarding multi-step form and promotion button when ff disabled


redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.tsx
```diff
@@ -124,7 +136,9 @@ export const VectorSearchCreateIndex = ({
   }, [step])
 
   useEffect(() => {
+    console.log('___ success/error changed!!!!', { success, error })
```
valkirilov: I guess you forgot to clean up this after debugging something.

valkirilov (commented):
We have the "**Getting Started**" button on a few more places, so I guess we should add the feature flag there as well:
- **Query Results**, when you don't have any searches yet
- **Manage Indexes** panel, when you don't have any indexes
- **Saved Queries** panel, when you don't have any indexes

Image

I believe this is the single place you can change:
- https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/vector-search/components/no-data-message/NoDataMessage.tsx#L29

redisinsight/ui/src/pages/vector-search/create-index/VectorSearchCreateIndex.tsx
ArtemHoruzhenko: was checking how review is happening. 
Now I know that it is ok 👍

ArtemHoruzhenko:
> Nice job! 🙌
> 
> We have the "**Getting Started**" button on a few more places, so I guess we should add the feature flag there as well:
> 
> * **Query Results**, when you don't have any searches yet
> * **Manage Indexes** panel, when you don't have any indexes
> * **Saved Queries** panel, when you don't have any indexes
> 
> Image
> I believe this is the single place you can change:
> 
> * https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/vector-search/components/no-data-message/NoDataMessage.tsx#L29

Good catch!
Covered few more places
|FF Enabled|FF Disabled|
|-|-|
Screenshot 2025-09-24 at 13 09 08|Screenshot 2025-09-24 at 13 08 55

valkirilov (approved):

PR #4986, opened 2025-09-24
RI-7547 Stop showing tooltip suggestions in the editor when inserting saved queries

valkirilov:
# Description

Disable the "suggestions tooltip" in the code editor on the **Vector Search** page when inserting a saved query, because it was hiding the **Run** button (_on a specific screen size_).

| Before | After |
| - | - |
image|image

# How it was tested

2. Go to the **Search** tab from the main navbar

_Note: Make sure to create a new index via the preset sample datasets, following the Vector Search onboarding wizard._

3. Open the **Saved Queries** panel from the top right corner
4. Click on the "**Insert**" button of any of the provided queries

At this point, the query should be added automatically to the code editor, but the "Run" button should remain visible.

https://github.com/user-attachments/assets/ee63c13e-7078-4f32-b518-0940d6ecf3e4

### DIFF
--- a/redisinsight/ui/src/pages/workbench/components/query/Query/Query.tsx
+++ b/redisinsight/ui/src/pages/workbench/components/query/Query/Query.tsx
@@ -493,6 +493,12 @@ const Query = (props: Props) => {
     const { data, forceHide, forceShow } = getSuggestions(editor, command)
     suggestionsRef.current = data
 
+    // Prevent suggestions if editor is not focused or cursor is not set
+    if (!editor.hasTextFocus() || !editor.getPosition()) {
+      editor.trigger('', 'hideSuggestWidget', null)
+      return
+    }
+
     if (!forceShow) {
       editor.trigger('', 'editor.action.triggerParameterHints', '')
       return
--- a/tests/playwright/pageObjects/pages/vector-search/vector-search-page.ts
+++ b/tests/playwright/pageObjects/pages/vector-search/vector-search-page.ts
@@ -36,6 +36,7 @@ export class VectorSearchPage extends BasePage {
     public readonly editorContainer: Locator
     public readonly editorViewLine: Locator
     public readonly editorTextBox: Locator
+    public readonly editorSuggesstionPopup: Locator
     public readonly editorSubmitButton: Locator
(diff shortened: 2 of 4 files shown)


ArtemHoruzhenko (approved):
lgtm

redisinsight/ui/src/pages/workbench/components/query/Query/Query.tsx
```diff
@@ -493,6 +493,12 @@ const Query = (props: Props) => {
     const { data, forceHide, forceShow } = getSuggestions(editor, command)
     suggestionsRef.current = data
 
+    // Prevent suggestions if editor is not focused or cursor is not set
+    if (!editor.hasTextFocus() || !editor.getPosition()) {
```
KrumTy, line 497: do we want test coverage for handleSuggestions?

valkirilov: There are ~0 tests related to this functionality, but I can try to build up some test coverage.

valkirilov: Bad news, sir. It seems like there was a reason for not having any tests related to the code editor. In the unit tests, currently we have mocked the Monaco editor, so it's not rendered at all. Nevertheless, I tried to see if there is a way to use the current mock (or extend it), but for the purpose of this use case, we rely on internal events (for the Monaco editor) to be in place and trigger the suggestion popup. 

So, I was not able to create any tests to verify this behavior, via unit tests. 

Still, there is a way to cover this with an e2e test with Playwright, since we have the full user experience there, but this is kind of an edge case, so I'm not sure whether we need it. What do you think? Should I invest time to implement such an e2e test case?

valkirilov: Ok, I have added two simple checks to the existing e2e tests for the new Search page, so now we verify whether the suggestions tooltip is visible in the code editor:
- it should be visible when you manually type in commands
- it shouldn't be visible when you insert commands from the "Saved Queries"

pd-redis (approved):

PR #4995, opened 2025-09-25
RI-000: keep manual trigger only for e2e

ArtemHoruzhenko:


.github/workflows/tests-e2e.yml
```diff
@@ -1,13 +1,6 @@
 name: ✅ E2E Tests
-run-name: >
-  E2E • ${{ github.event_name == 'pull_request_review'
```
valkirilov, line 3: You don't like this name? 😀

.github/workflows/tests-e2e.yml
```diff
@@ -21,43 +14,24 @@ permissions:
...
   build-appimage:
-    needs: approval-dedupe
-    if: needs.approval-dedupe.outputs.mode == 'run'
     uses: ./.github/workflows/pipeline-build-linux.yml
     secrets: inherit
     with:
       target: build_linux_appimage_x64
       debug: ${{ inputs.debug || false }}
 
-  builds-complete:
```
valkirilov, line 54: I see your point of removing this, as a dependency for both the `tests-e2e-playwright` and `e2e-appimage-tests` jobs, because each of them depends on a separate build step. 

But in the end, we don't have a way to trigger only one of these steps. As I understand, they will always run together, and if one of them fails, it means the whole pipeline goes red (yes, we can re-run them individually, but overall it means "tests didn't pass"). 

I believe this is why @KrumTy made this `builds-complete` step, to ensure that we have a proper setup for running the tests and save some time and resources, in case the build fails. I'm not against either of the approaches. I'm just pointing this out since it's relatively new code that we’re reverting now.

ArtemHoruzhenko: I reverted a lot of new code :) 
I understand your point but it shouldn't work this way. We shouldn't wait until not real dependency will be build before starting tests. I think we might separate desktop and web tests in the future. For now IMO it is enough to keep as is but no "builds-complete"

.github/workflows/tests-e2e.yml
ArtemHoruzhenko: we don't need this code, basically. there are checks github.event_name == 'pull_request_review' that will never satisfy

valkirilov (approved):

PR #4998, opened 2025-09-26
RI-7508: get db version from another selector since it might be not a…

ArtemHoruzhenko:
get db version from another selector since it might be not returned by another selector.

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/create-index/hooks/useRedisInstanceCompatibility.ts
+++ b/redisinsight/ui/src/pages/vector-search/create-index/hooks/useRedisInstanceCompatibility.ts
@@ -1,6 +1,6 @@
 import { useMemo } from 'react'
 import { useSelector } from 'react-redux'
-import { connectedInstanceSelector } from 'uiSrc/slices/instances/instances'
+import { connectedInstanceInfoSelector, connectedInstanceSelector } from 'uiSrc/slices/instances/instances'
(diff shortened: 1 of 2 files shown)


redisinsight/ui/src/pages/vector-search/create-index/hooks/useRedisInstanceCompatibility.ts
```diff
@@ -15,10 +15,13 @@ const REDISEARCH_MODULE_SET = new Set(REDISEARCH_MODULES)
 
 const useRedisInstanceCompatibility =
   (): UseRedisInstanceCompatibilityReturn => {
+    const {
+      version,
+    } = useSelector(connectedInstanceInfoSelector)
```
valkirilov, line 20: Ouch, this means we have different behavior for what we store in Redux, based on the environment 😔

Do you think that we can get rid of one of the selectors (not now, I'm just asking if you checked what we store in both selectors, and whether it's possible to live only with one of them).

valkirilov (approved):

redisinsight/ui/src/pages/vector-search/create-index/hooks/useRedisInstanceCompatibility.ts
ArtemHoruzhenko: Something to think about. I'm looking on another solution for this currently. Let see if it will work

PR #5011, opened 2025-09-30
RI-7540: update key name text input styles

KrumTy:
| Old UI | Before | After |
| --- | --- | --- |
| Screenshot 2025-09-30 at 13 08 02 | Screenshot 2025-09-30 at 13 08 45 | Screenshot 2025-09-30 at 13 06 47 |

### DIFF
--- a/redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-name/styles.module.scss
+++ b/redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-name/styles.module.scss
@@ -64,3 +64,7 @@
 .hidden {
   display: none;
 }
+
+.keyFlexItem {
+  flex-grow: 5 !important;
+}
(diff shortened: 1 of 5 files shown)


redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-name/styles.module.scss
```diff
@@ -62,3 +64,7 @@
 .hidden {
   display: none;
 }
+
+.keyFlexItem {
+  flex-grow: 5 !important;
```
pd-redis: For `FlexItem`, `grow` prop can take values from 1 to 10, which represent exactly this. Use that instead of custom styles

redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-name/styles.module.scss
pd-redis, line 1: We should try to get rid of scss - move to single css solution - rather than adding to it

pd-redis (changes requested):

redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-name/styles.module.scss
KrumTy: I brought the class name and style that was removed with the previous file commit

now removed in favor of grow prop

pd-redis (approved):

PR #5012, opened 2025-09-30
RI-7560: fix bottom group panel movement when expand / collapse

dantovska:
### Apply changes to the panels on the Browser Page to avoid not user friendly behavior:
1. Fix the moving on the bottom group component (containing the CLI / Command Helper / Profiler badges)

**Before:**

https://github.com/user-attachments/assets/acd3838b-973e-4091-9912-86daa1da80d3

**After:**

https://github.com/user-attachments/assets/4933f9a0-204b-4d51-932a-33c8ae4eb452

2. Change the border radius of the key details panel to match the border radius on the key list panel

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/redisinsight/ui/src/pages/browser/BrowserPage.tsx
+++ b/redisinsight/ui/src/pages/browser/BrowserPage.tsx
@@ -5,6 +5,8 @@ import cx from 'classnames'
 
 import { isNumber } from 'lodash'
 import { useTheme } from '@redis-ui/styles'
+import styled from 'styled-components'
+
 import {
   formatLongName,
   getDbIndex,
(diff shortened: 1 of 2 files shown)


valkirilov (approved):

redisinsight/ui/src/pages/browser/BrowserPage.tsx
```diff
@@ -361,7 +361,7 @@ const BrowserPage = () => {
             })}
             style={{
               border: `1px solid ${theme.semantic.color.border.neutral500}`,
-              borderRadius: `5px`,
+              borderRadius: `8px`,
```
pd-redis: is there a way to move this to the resizable panel component and avoid the need of custom style here?

PR #5013, opened 2025-10-01
RI-7573 Rename "Saved queries" to "Sample queries"

valkirilov:
# Description

Rename "Saved queries" to "Sample queries" everywhere in the new Search page
- button in the top right corner used to open the respective panel
- title of the panel
- fallback message when there is no data

image

# How it was tested

## Manually

2. Go to the **Search** tab from the main navbar
3. Open the **Sample Queries** panel from the top right corner

## Automatically via e2e tests

```
# From the root directory
yarn dev:api

yarn dev:ui

yarn test:chromium:local-web vector-search/saved-queries

yarn playwright show-report
``` 

image

### DIFF
--- a/tests/playwright/tests/vector-search/saved-queries.spec.ts
+++ b/tests/playwright/tests/vector-search/saved-queries.spec.ts
@@ -85,11 +85,11 @@ test.describe('Vector Search - Saved Queries', () => {
         // )
 
         await expect(searchPage.savedQueriesContainer).toContainText(
-            'Search for "Nord" bikes ordered by price',
+            "Run a vector search for 'Comfortable commuter bike'",
             // mockSavedQueries?.queries[0].label!,
         )
         await expect(searchPage.savedQueriesContainer).toContainText(
-            'Find road alloy bikes under 20kg',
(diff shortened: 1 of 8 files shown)


tests/playwright/tests/vector-search/saved-queries.spec.ts
```diff
@@ -109,7 +109,7 @@ test.describe('Vector Search - Saved Queries', () => {
 
         // Ensure the queries are displayed
         await expect(searchPage.savedQueriesContainer).toContainText(
-            'Search for "Nord" bikes ordered by price', // TODO: Replace this with actual query, once we reimplement them soon
+            "Run a vector search for 'Comfortable commuter bike'", // TODO: Replace this with actual query, once we reimplement them soon
```
pawelangelow, line 112: So, can we drop the comment?

valkirilov: I prefer to keep it for now, because in the second iteration, we plan to add CRUD for these saved queries, and ideally, we should check not for the hardcoded queries, but for the actual queries.

PR #5018, opened 2025-10-01
RI-4764: Fix insights bulb highlight style

dantovska:
### Fix insights bulb highlight style
- make the highlighting styles to appear on the bulb 
- background when insights open
- refactor the code to use styled components and only essential styling

#### Where to find
**Option 1 [DEV]:** InsightsTrigger, replace the isHighlighted declaration and initialization with hard coded value instead from the store:
``` 
  const { isHighlighted } = { isHighlighted: true }
```

**Option 2:** 
1. Create string key with random name and value some JSON, e.g. {"a": "b"}. 
2. New tip should appear for using JSONs in this case instead of strings 

**Before:**
The lightbulb did not show at all - the style was broken

https://github.com/user-attachments/assets/3f06a553-d4e9-47d8-b96e-ad3ec699d517

**After:**

https://github.com/user-attachments/assets/f771d879-48d7-4a18-b1a3-9ea0eaba7005


redisinsight/ui/src/components/triggers/insights-trigger/insights-trigger.styles.ts
```diff
@@ -0,0 +1,40 @@
...
+import { IconButtonProps } from "uiSrc/components/base/forms/buttons/IconButton"
+
+export const BulbWrapper = styled.div`
+  position: relative;
+`
+
+export const BulbHighlighting = styled.span`
+  // TODO: Using the background color from the previous value until there is an appropriate color 
+  // from the pallete to use for both light and dark themes.
+  background-color: #ffaf2b;
```
valkirilov, line 13: We already have such a color defined in the palette, so you can use it.

```suggestion
  background-color: ${({ theme }) => theme.color.orange500};
```

dantovska: test it on both themes 😄 yes, it's not supposed to change like semantic colors, but it does 😄

pd-redis (approved):

PR #5020, opened 2025-10-02
Ri-7586: issues pointed by Yarden

pd-redis:


redisinsight/ui/src/components/command-helper/components/command-helper-search-output/CHSearchOutput.tsx
```diff
@@ -90,7 +90,7 @@ const CHSearchOutput = ({ searchedCommands }: Props) => {
                     handleClickCommand(e, command)
                   }}
                 >
-                  <UnderlineReverseLink color="text" variant="regular">
+                  <UnderlineReverseLink color="text" variant="button">
```
dantovska: Adding href (and switching back to inline variant) seems to fix it, which is crazy ... I don't know if we should use link here at all...

```suggestion
                  <UnderlineReverseLink href="#" color="text" variant="inline">
```

dantovska (dismissed):

valkirilov (approved):

PR #5031, opened 2025-10-03
RI-7574: allow setting default theme from query param

KrumTy:


redisinsight/ui/src/components/config/Config.tsx
```diff
@@ -211,7 +209,7 @@ const Config = () => {
   const checkAndSetTheme = () => {
     const theme = config?.theme
     if (theme && localStorageService.get(BrowserStorageItem.theme) !== theme)
-      themeService.applyTheme(theme as Theme)
+      changeTheme(theme)
```
KrumTy, line 212: changing theme via changeTheme is the correct approach as it updates the theme provider and calls themeService.applyTheme internally

redisinsight/ui/src/contexts/themeContext.tsx
```diff
@@ -26,6 +26,14 @@ interface Props {
 
 const THEME_NAMES = THEMES.map(({ value }) => value)
 
+const getQueryTheme = () => {
+  const queryThemeParam = new URLSearchParams(window.location.search)
```
pd-redis, line 30: Can't we use something from react-router for this?

redisinsight/ui/src/contexts/themeContext.tsx
```diff
@@ -26,6 +26,14 @@ interface Props {
...
+    .get('theme')
```
valkirilov, line 31: Note: It's not directly related to your changes, but I would suggest putting this "magic string" behind an enum. 

We have similar logic in about 15 other places in the codebase where we use the same approach. I believe having a centralized location where we can view a list of all the available query parameters that Redis Insight accepts and handles will be beneficial. It will be both cleaner and also better in terms of self-explanatory "code documentation".
 
If you find this advice useful, we can log a separate tech debt ticket with low priority and reference it outside the scope of this change.

redisinsight/ui/src/contexts/themeContext.tsx
```diff
@@ -41,11 +49,14 @@ export class ThemeProvider extends React.Component<Props> {
   constructor(props: any) {
     super(props)
 
+    // theme query param can be used to override the local/default theme
+    const queryTheme = getQueryTheme()
     const storedThemeValue = localStorageService.get(BrowserStorageItem.theme)
     const theme =
-      !storedThemeValue || !THEME_NAMES.includes(storedThemeValue)
+      queryTheme ||
```
pd-redis: This starts to look too complex, maybe refactor it?

KrumTy: it is complex and I don't think there's a way to make it not look ugly w/o major refactoring (out of scope)

redisinsight/ui/src/contexts/themeContext.tsx
KrumTy: there's useLocation hook that returns the query string but we are in the context of a class component here, so I made a simple method for accessing the param

redisinsight/ui/src/contexts/themeContext.tsx
pd-redis: I don't think this is too major

``` typeScript
    const storedThemeValue = localStorageService.get(BrowserStorageItem.theme)
    // theme query param can be used to override the local/default theme
    const queryTheme = getQueryTheme()

    let theme = defaultState.theme
    if (storedThemeValue && THEME_NAMES.includes(storedThemeValue)) {
      theme = storedThemeValue
    }
    if (queryTheme) {
      theme = queryTheme
    }
```

redisinsight/ui/src/contexts/themeContext.tsx
pd-redis: OK

redisinsight/ui/src/contexts/themeContext.tsx
KrumTy: I was referring to the underlying fact that we rely on env var, constant, local storage, db entry and now a query string for the theme value;
refactored the bit

PR #5037, opened 2025-10-06
fix(ui): add spacing betwen the key details in the header

valkirilov:
# Description

Added spacing between the key details in the header.

| Before | After |
| - | - |
image|Screenshot 2025-10-06 at 15 35 17

# How it was tested

2. Open the **Browser** tab from the top navigation
3. Open existing key or add a new one (regardless of its type)

Observe the header with the details about the key (size, ttl, etc.)

| Old | Before | After |
| - | - | - |
Screenshot 2025-10-06 at 15 36 45|image|Screenshot 2025-10-06 at 15 35 17

# Side Quest

Additionally, there was a very odd behavior when you hover over the TTL field and trigger the inline editor. Now, I made some very ugly CSS fixes to make it look like the way before the EUI replacement.

_Note: There is a very weird `letter-spacing: -0.12px;` and `-webkit-font-smoothing: antialiased;` applied across the codebase that causes some odd rendering of the font. I didn't touch that, because it looks like a general issue, but we should think about it. Maybe it was a fix related to an old font, but now it makes the current one look blurred._

### Old

https://github.com/user-attachments/assets/bd6d1860-7cb2-464b-85d7-7cd310f871c1

### Before 
https://github.com/user-attachments/assets/600d7ecd-7e50-46c5-9145-7d5a8b2423be

### After
https://github.com/user-attachments/assets/f70a57d1-6514-4523-8ea8-181358d5cf48

### DIFF
--- a/redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-ttl/KeyDetailsHeaderTTL.tsx
+++ b/redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-ttl/KeyDetailsHeaderTTL.tsx
@@ -142,7 +142,7 @@ const KeyDetailsHeaderTTL = ({ onEditTTL }: Props) => {
         )}
         <Text
           size="s"
-          className={cx(styles.subtitleText, styles.subtitleTextTTL, {
+          className={cx(styles.subtitleTextTTL, {
             [styles.hidden]: ttlIsEditing || ttlIsHovering,
           })}
           data-testid="key-ttl-text"


redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/monaco-editor/MonacoEditor.tsx
```diff
@@ -64,8 +66,8 @@ const MonacoEditor = (props: BaseProps) => {
         editorWrapperClassName={styles.editorWrapper}
         onEditorDidMount={onEditorDidMount}
       />
-
-      <EuiFlexItem className={styles.actions}>
+      <Spacer size="m" />
```
valkirilov, line 68: Bonus Points: Not related to the changes of the popover colors, but I fixed this minimal spacing issue as well.

| Before | After |
| - | - |
Image|Image

redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-ttl/KeyDetailsHeaderTTL.tsx
```diff
@@ -139,7 +142,7 @@ const KeyDetailsHeaderTTL = ({ onEditTTL }: Props) => {
         )}
         <Text
           size="s"
-          className={cx(styles.subtitleText, {
+          className={cx(styles.subtitleText, styles.subtitleTextTTL, {
```
dantovska: since subtitleText is adding some paddings and you are removing it in subtitleTextTTL, does it make sense to remove the subtitleText class?

redisinsight/ui/src/pages/browser/modules/key-details/components/rejson-details/monaco-editor/MonacoEditor.tsx
dantovska: Nice, there is a related ticket so it should be closed after this. I will link it to your issue. Tnx!

redisinsight/ui/src/pages/browser/modules/key-details-header/components/key-details-header-ttl/KeyDetailsHeaderTTL.tsx
valkirilov: Thanks for pointing it out. I thought we needed it because it comes with additional styling, but I just checked it and it works fine even without it. So, now it's removed, thanks!

dantovska (approved):

PR #5040, opened 2025-10-07
RI-7519: set text and link sizes to be the same

pd-redis:
Made links same size, made the button a little more like in the figma designs 

Before:
image

After:

image


dantovska:
This seems to be missing a space: 
image

valkirilov (dismissed):

pd-redis:
> This seems to be missing a space: image

I don't think we can fix this, maybe report it to @ViktarStarastsenka

ViktarStarastsenka:
> > This seems to be missing a space: image
> 
> I don't think we can fix this, maybe report it to @ViktarStarastsenka

I already raised a fix. It is resolved by now.

PR #5043, opened 2025-10-07
RI-7614 Filter Query Results based on the type of the page

valkirilov:
# Description

Show the correct command results on the **Search** and **Workbench** pages, by filtering out the queries based on the page from which they were executed.
- make sure we use the feature flag to toggle the data source (Index DB or SQLite) for the **Search** page as well (like it's done for the Workbench page)
- show filtered commands when using IndexDB (based on their type, matching the page)
- show filtered commands when using SQLite (rework the main hook that was responsible to CRUD the history, so it can work for the Search page as well)
- extended the IndexDB indexes to support easier deletion by type (Search/Workbench), when using the "Clear results" button. It should remove only commands related to the page, not all of them

https://github.com/user-attachments/assets/39b733d2-880f-42a9-81ca-e410d77d7733

| Before | After |
| - | - |
image|image
image|image

# How it was tested

We have two available ways to store the history of the executed commands - SQLite and IndexDB inside the browser itself.

## Using Index DB

Create `redisinsight/ui/.env` and make sure to put `RI_FEATURES_ENV_DEPENDENT_DEFAULT_FLAG=false` inside it.

### Using SQLite

It's the default mode, but in case you added `RI_FEATURES_ENV_DEPENDENT_DEFAULT_FLAG` make sure to flip it to  `true`

### Reproduction Steps

1. Go to **Databases** and open a connection to an existing database instance, ot create a new one
2. Open the **Search** page and execute a command inside the code editor (like `SEARCH COMMAND`)
3. Go to the **Workbench** page and execute another command (like `WORKBENCH COMMAND`)

Switch between the pages and make sure you see the correct commands in the list with the results.

| Search | Workbench |
| - | - |
image|image

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/query/useQuery.ts
+++ b/redisinsight/ui/src/pages/vector-search/query/useQuery.ts
@@ -1,193 +1,40 @@
 import { useCallback, useEffect, useRef, useState } from 'react'
 import { useParams } from 'react-router-dom'
-import { chunk, reverse } from 'lodash'
-import { useSelector } from 'react-redux'
-import { RootState } from 'uiSrc/slices/store'
-import {
-  Nullable,
-  getCommandsForExecution,
-  getExecuteParams,
-  isGroupResults,
-  isSilentMode,
-} from 'uiSrc/utils'
+import { useDispatch, useSelector } from 'react-redux'
 import { CodeButtonParams } from 'uiSrc/constants'
 import {
   RunQueryMode,
   ResultsMode,
-  CommandExecutionUI,
-  CommandExecution,
   CommandExecutionType,
 } from 'uiSrc/slices/interfaces'
-import { PIPELINE_COUNT_DEFAULT } from 'uiSrc/constants/api'
 import {
-  addCommands,
-  clearCommands,
-  findCommand,
-  removeCommand,
-} from 'uiSrc/services/workbenchStorage'
-import {
-  createErrorResult,
-  createGroupItem,
-  executeApiCall,
-  generateCommandId,
-  limitHistoryLength,
-  loadHistoryData,
-  prepareNewItems,
-  scrollToElement,
-  sortCommandsByDate,
-} from './utils'
+  fetchWBHistoryAction,
+  workbenchResultsSelector,
+  sendWbQueryAction,
+  deleteWBCommandAction,
+  clearWbResultsAction,
+  fetchWBCommandAction,
+  toggleOpenWBResult,
+} from 'uiSrc/slices/workbench/wb-results'
+import { Nullable } from 'uiSrc/utils'
 
 const useQuery = () => {
   const { instanceId } = useParams<{ instanceId: string }>()
   const scrollDivRef = useRef<HTMLDivElement>(null)
+  const dispatch = useDispatch()
 
   const [query, setQuery] = useState('')
-  const [items, setItems] = useState<CommandExecutionUI[]>([])
-  const [clearing, setClearing] = useState(false)
-  const [processing, setProcessing] = useState(false)
-  const [isLoaded, setIsLoaded] = useState(false)
-  const envDependentFlag = useSelector(
-    (state: RootState) =>
-      state?.app?.features?.featureFlags?.features?.envDependent?.flag,
+  const { items, clearing, processing, isLoaded } = useSelector(
+    workbenchResultsSelector,
   )
 
   const resultsMode = ResultsMode.Default
   const activeRunQueryMode = RunQueryMode.ASCII
+  const executionType = CommandExecutionType.Search
 
   useEffect(() => {
-    const loadHistory = async () => {
-      try {
-        const historyData = await loadHistoryData(instanceId)
-        setItems(historyData)
-      } catch (error) {
-        // Silently handle error
-      } finally {
-        setIsLoaded(true)
-      }
-    }
-
-    loadHistory()
-  }, [instanceId])
-
-  const handleApiSuccess = useCallback(
-    async (
-      data: CommandExecution[],
-      commandId: string,
-      isNewCommand: boolean,
-    ) => {
-      setItems((prevItems) => {
-        const updatedItems = prevItems.map((item) => {
-          const result = data.find((_, i) => item.id === commandId + i)
-          if (result) {
-            return {
-              ...result,
-              loading: false,
-              error: '',
-              isOpen: !isSilentMode(resultsMode),
-            }
-          }
-          return item
-        })
-        return sortCommandsByDate(updatedItems)
-      })
-
-      if (envDependentFlag === false) {
-        await addCommands(reverse(data))
-      }
-
-      if (isNewCommand) {
-        scrollToElement(scrollDivRef.current, 'start')
-      }
-    },
-    [resultsMode],
-  )
-
-  const handleApiError = useCallback((error: unknown) => {
-    const message =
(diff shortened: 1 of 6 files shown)


redisinsight/ui/src/pages/vector-search/create-index/hooks/useCreateIndex.ts
```diff
@@ -57,7 +63,7 @@ export const useCreateIndex = (): UseCreateIndexResult => {
         const data = await executeQuery(instanceId, cmd)
 
         // Step 3: Persist results locally so Vector Search history (CommandsView) shows it
-        if (Array.isArray(data) && data.length) {
+        if (envDependentFlag === false && Array.isArray(data) && data.length) {
```
valkirilov, line 66: We don't want to preserve the commands inside IndexDB always, it should depend on this feature flag, as it's handled in the Workbench.

redisinsight/ui/src/pages/vector-search/query/useQuery.ts
```diff
@@ -84,7 +91,9 @@ const useQuery = () => {
         return sortCommandsByDate(updatedItems)
       })
 
-      await addCommands(reverse(data))
+      if (envDependentFlag === false) {
```
valkirilov: We don't want to preserve the commands inside IndexDB always, it should depend on this feature flag, as it's handled in the Workbench.

redisinsight/ui/src/services/workbenchStorage.ts
```diff
@@ -70,8 +76,21 @@ export class WorkbenchStorage {
...
+            CommandExecutionIndex.DatabaseId,
+            'databaseId',
+            { unique: false },
+          )
+          objectStore.createIndex(CommandExecutionIndex.CommandId, 'id', {
+            unique: true,
+          })
+          objectStore.createIndex(
+            CommandExecutionIndex.DatabaseIdType,
+            ['databaseId', 'type'],
```
valkirilov, line 89: We should only remove the records related to the page (Search/Workbench) when clicking the “Clear Results” button. Instead of adding an additional check in the cursor (which iterates over the records while deleting them), I’m introducing a new index that can easily target the correct type.

However, if you find this approach to be incorrect, I can revert to the iterative deletion method.

redisinsight/ui/src/services/workbenchStorage.ts
```diff
@@ -250,8 +270,8 @@ export class WorkbenchStorage {
           const objectStore = db
             .transaction(storeName, 'readwrite')
             ?.objectStore(storeName)
-          const idbIndex = objectStore?.index('dbId')
-          const indexReq = idbIndex?.openCursor(dbId)
+          const idbIndex = objectStore?.index(indexName)
+          const indexReq = idbIndex?.openCursor(indexSelector)
```
valkirilov, line 274: Here, we simply pass the database instance and the type we target (search or workbench), and it will do it.

```javascript
const idbIndex = objectStore?.index('dbId_type')
const indexReq = idbIndex?.openCursor(['databaseId', 'SEARCH'])
```

Of course, we still support the old behavior, so you can still target a specific database instance and delete all records related to it.

```javascript
const idbIndex = objectStore?.index('dbId')
const indexReq = idbIndex?.openCursor('databaseId')
```

Otherwise, we should delete the records by their type inline, inside the cursor below.

redisinsight/ui/src/services/workbenchStorage.ts
```diff
@@ -278,19 +298,27 @@ export class WorkbenchStorage {
...
+        BrowserStorageItem.wbCommandsHistory,
+        dbId,
+      )) as CommandHistoryType) || []
+
+    if (!commandExecutionType) {
+      return history
+    }
 
-    return history || []
+    return history.filter((item) => item.type === commandExecutionType)
```
valkirilov, line 321: Here we do the magic to apply some filtering, based on the type (Search/Workbench), so we don't show the full history on both pages.

redisinsight/ui/src/pages/vector-search/query/useQuery.ts
```diff
@@ -1,184 +1,40 @@
...
-  const [clearing, setClearing] = useState(false)
-  const [processing, setProcessing] = useState(false)
-  const [isLoaded, setIsLoaded] = useState(false)
+  const { items, clearing, processing, isLoaded } = useSelector(
+    workbenchResultsSelector,
+  )
 
   const resultsMode = ResultsMode.Default
   const activeRunQueryMode = RunQueryMode.ASCII
+  const executionType = CommandExecutionType.Search
```
valkirilov, line 33: Basically, we need to rely on the current implementation for the history manipulation (and automatically switch between SQLite and IndexDB based on the feature flag), so we should go for the existing helpers instead of building them again.

redisinsight/ui/src/pages/vector-search/create-index/hooks/useCreateIndex.ts
```diff
@@ -26,14 +28,18 @@ export const useCreateIndex = (): UseCreateIndexResult => {
   const [success, setSuccess] = useState(false)
   const [loading, setLoading] = useState(false)
   const [error, setError] = useState<Error | null>(null)
+  const envDependentFlag = useSelector(
+    (state: RootState) =>
+      state?.app?.features?.featureFlags?.features?.envDependent?.flag,
+  )
```
ArtemHoruzhenko, line 34: I feel like we must return `envDependent: { flag: boolean}` here and use it below like `envdepenent.flag`

redisinsight/ui/src/pages/vector-search/create-index/hooks/useCreateIndex.ts
ArtemHoruzhenko: I don't think this should be done on this layer. Ideally we must always call addCommands and then have a storage strategy based on envDependent flag

redisinsight/ui/src/pages/vector-search/create-index/hooks/useCreateIndex.ts
valkirilov: I see we have `appFeatureFlagsFeaturesSelector` Redux selector, so we can rely on it, or create a new one for this `envDependant` flag as well. It should be applied to a few more places in the app as well.

redisinsight/ui/src/pages/vector-search/create-index/hooks/useCreateIndex.ts
pd-redis: First, `addCommands` is tightly related to `wbHistoryStorage`, it is not meant to be general solution for command history storage.  Internally it uses `WorkbenchStorage` class, that is writing to IndexedDb. 
Same or similar thing can be applied to search history storage. Since it is new implementation, I agree with Artem, that better approach would be an implementation, that takes a `useBackendPersistanse = false` flag and internally decides how to store the commands, but using `addCommands` will ALWAYS add these commands to workbench history, which is apparently a problem

redisinsight/ui/src/services/workbenchStorage.ts
pd-redis: I thinks this is wrong approach. Do not change WorkbenchStorage class to differentiate between storage types, just create vectorSearchHistoryStorage 
```typescript
export const vectorSearchHistoryStorage = new WorkbenchStorage(
  riConfig.app.vectorSearchName,
  2,
)
```

And if you want to generalize `addCommands`, `getLocalWbHistory` etc, modify them to accept the storage instance

PR #5044, opened 2025-10-07
Bugfix/ri 7625 handle no such index error

ArtemHoruzhenko:
Fixed error handling during index creation
In additional fixed create index layout 
|Before|After|
|-|-|
Screenshot 2025-10-07 at 16 10 15|Screenshot 2025-10-07 at 16 09 15

### DIFF
--- a/redisinsight/ui/src/pages/browser/components/create-redisearch-index/CreateRedisearchIndexWrapper.tsx
+++ b/redisinsight/ui/src/pages/browser/components/create-redisearch-index/CreateRedisearchIndexWrapper.tsx
@@ -11,9 +11,7 @@ import { Link } from 'uiSrc/components/base/link/Link'
 import { RiTooltip } from 'uiSrc/components'
 import CreateRedisearchIndex from './CreateRedisearchIndex'
 
-import { Spacer } from 'uiSrc/components/base/layout'
-import Divider from 'uiSrc/components/divider/Divider'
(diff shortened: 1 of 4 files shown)


redisinsight/ui/src/pages/browser/components/create-redisearch-index/CreateRedisearchIndexWrapper.tsx
```diff
@@ -73,9 +75,8 @@ const CreateRedisearchIndexWrapper = ({
           </Link>
         </Text>
       </FlexItem>
-      <Spacer size="xl" />
     </StyledHeader>
-    <HorizontalRule margin="xs" />
+    <Divider colorVariable="separatorColor" className={styles.divider} />
```
dantovska: why move back to the divider? importing the styles from the browser add-key seems wrong to me.

dantovska: if it is just the color for the horizontal rule, this could have been adjusted ...

ArtemHoruzhenko: @dantovska I moved to the same component we use for delimiter in other places. 
Agree about styles... ctrl+c - ctrl+v... will fix it

PR #5045, opened 2025-10-07
RI-7618: make entry id input looks the same as other inputs on the form

ArtemHoruzhenko:
|Before|After|
|-|-|
Screenshot 2025-10-07 at 16 36 22|Screenshot 2025-10-07 at 16 36 13
Screenshot 2025-10-07 at 16 36 28|Screenshot 2025-10-07 at 16 36 06

### DIFF
--- a/redisinsight/ui/src/pages/browser/modules/key-details/components/stream-details/add-stream-entity/StreamEntryFields/StreamEntryFields.tsx
+++ b/redisinsight/ui/src/pages/browser/modules/key-details/components/stream-details/add-stream-entity/StreamEntryFields/StreamEntryFields.tsx
@@ -149,7 +149,6 @@ const StreamEntryFields = (props: Props) => {
             onChange={handleEntryIdChange}
             onBlur={onEntryIdBlur}
             onFocus={() => setIsEntryIdFocused(true)}
-            valid={!entryIdError}
             autoComplete="off"
             data-testid={config.entryId.id}
           />


redisinsight/ui/src/pages/browser/modules/key-details/components/stream-details/add-stream-entity/StreamEntryFields/StreamEntryFields.tsx
```diff
@@ -149,7 +149,6 @@ const StreamEntryFields = (props: Props) => {
             onChange={handleEntryIdChange}
             onBlur={onEntryIdBlur}
             onFocus={() => setIsEntryIdFocused(true)}
-            valid={!entryIdError}
             autoComplete="off"
             data-testid={config.entryId.id}
```
valkirilov, line 153: Note: This might require bigger changes, so maybe we shouldn't handle it as part of this task, but still, I wanted to point it out that according to Redis UI, we should show the errors inline and highlight the border color, but currently we do this in a different manner in Redis Insight. 

Still, minor and not crucial at this point, just a note.

| Redis UI | Redis Insight |
| - | - |
Image|Image

ArtemHoruzhenko: I know. It seems to me that we should make a decision on this behavior and follow it in all inputs. For now it is much easy to remove "valid"/"error" colors for a single input rather then add it to all the rest

valkirilov: Yes, totally agree with that. Because changing the visuals is not only replacing the components, but also following the concept behind the design system that we adopt. 

But it's a big topic, that's why I pointed it out only as a note.

PR #5046, opened 2025-10-07
RI-7610: remove purple color while hovering on formatters select

ArtemHoruzhenko:
|Before|After|
|-|-|
Screenshot 2025-10-07 at 16 45 39|Screenshot 2025-10-07 at 16 45 27


dantovska (commented):
Maybe we do want to leave some color there, not remove any color on hover?

valkirilov (approved):

KrumTy (approved):

ArtemHoruzhenko:
> Maybe we do want to leave some color there, not remove adding a color on hover? this changes the experience and I don't know if it is better...

I think you are right but currently It was agreed to remove bg color at all. We will back to it when we have designs

PR #5057, opened 2025-10-13
RI-7614 Introduce shared commands history (for Search and Workbench)

valkirilov:
# Description

Introduce a simple abstraction layer for fetching Command History data from any data source.

Currently, we support storing this information in two places:
- SQLite
- IndexedDB

And we don't have a simple way to CRUD the data in a generic way, without having a highly coupled implementation dependent on the data store. With this pull request, we aim to introduce a way to make this more reusable and easier to extend in the future by creating a common interface that should be independent of the data source used behind it.

### Using SQLite

This is the default data store, so when you start the application, go and execute a bunch of queries on both the Search and Workbench pages. The following should work as expected:
- loading history
- executing new commands or re-executing existing cooomands
- expanding a row for additional details
- removing a single history record
- clearing all history records

https://github.com/user-attachments/assets/078adaf3-35d1-40d1-9fae-0fb09d9408d6

### Using IndexedDB

In order the toggle the data source and make the app use IndexedDB, put the following environment variable.

```
RI_FEATURES_ENV_DEPENDENT_DEFAULT_FLAG=false
```

Then, you can follow the same scenarios as for the SQLite variant.

https://github.com/user-attachments/assets/a0c172be-6629-494a-839c-5844978287b2

## More details

At this point, we do use the **Command History** on the **Search** and **Workbench** pages. And so far, its implementation has been inconsistent:
- on **Workbench**, we can store this information in SQLite or Index DB (toggleable via a feature flag), and it goes in Redux
- on **Search** page, we were storing this only in Index DB, we don't have the feature flag toggle, and we store it wrong inside the Workbench index. And lastly, it's using a hook with local React state, it's completely independent from Redux.

With this pull request, we aim to introduce a generic way to handle this, but we'll focus on the implementation for the **Search** page only, which is the problematic one. Later on, we can easily bind the Workbench implementation to the new services, because they're built in a way that can be easily reused across the codebase (or at least, this is the plan)

### Notes

Here is a short breakdown of the concepts introduced by this pull request:
- created an interface to define the behavior for the services that should fetch command history data (no matter of the data source). It's a really simple CRUD, and all services should implement it when interacting with various data sources (SQLite, IndexedDB, etc.)
- created an example implementation for fetching command history data from both SQLite and IndexedDB
- created a service to toggle the data source when fetching commands history data, based on a feature flag and target SQLite or IndexedDB
- extend the IndexedDB database, to support storage for both Search and Workbench pages, using the same schema.
- apply the changes only to the Search page related functionalities. The Workbench page remains untouched and operational, but can be updated in the future to use the same service.

PS: There are more comments bellow, scroll down.

Screenshot 2025-10-10 at 13 59 00

### DIFF
--- a/redisinsight/ui/src/services/commands-history/database/CommandsHistoryIndexedDB.ts
+++ b/redisinsight/ui/src/services/commands-history/database/CommandsHistoryIndexedDB.ts
@@ -23,7 +23,6 @@ import {
   wbHistoryStorage,
   WorkbenchStorage,
 } from 'uiSrc/services/workbenchStorage'
-import { C } from 'msw/lib/glossary-2792c6da'
 
 export class CommandsHistoryIndexedDB implements CommandsHistoryDatabase {
   private readonly dbStorage: WorkbenchStorage


redisinsight/ui/src/pages/vector-search/query/useQuery.ts
```diff
@@ -44,13 +45,17 @@ const useQuery = () => {
...
+  })
+
   const resultsMode = ResultsMode.Default
   const activeRunQueryMode = RunQueryMode.ASCII
 
   useEffect(() => {
     const loadHistory = async () => {
       try {
-        const historyData = await loadHistoryData(instanceId)
+        const historyData = await getCommandsHistory(instanceId)
```
valkirilov: My goal is to preserve the current way of managing command history through the Search page (with this hook) while modifying only the underlying logic responsible for interacting with the data store. 

This approach ensures that the existing components remain unchanged, while simultaneously achieving the goal of having an independent data store that can be easily changed via a feature flag. 

And of course, by design, this hook (and the components that depend on it) remain unaware of this information, and they will not be affected by any future changes of the data source behind it.

redisinsight/ui/src/services/commands-history/database/CommandsHistoryIndexedDB.ts
```diff
@@ -0,0 +1,15 @@
+import { CommandExecutionType } from 'uiSrc/slices/interfaces'
+import { CommandsHistoryDatabase, CommandHistoryResult } from './interface'
+
+export class CommandsHistoryIndexedDB implements CommandsHistoryDatabase {
+  async getCommandsHistory(
+    instanceId: string,
+    commandExecutionType: CommandExecutionType,
+  ): Promise<CommandHistoryResult> {
+    // TODO: Implementation for fetching command history from IndexedDB
```
valkirilov: It will be implemented later (maybe in a separate pull request).

redisinsight/ui/src/services/commands-history/hooks/useCommandsHistory.ts
```diff
@@ -0,0 +1,56 @@
...
+
+interface CommandHistoryProps {
+  commandExecutionType: CommandExecutionType
+}
+
+interface CommandsHistoryHook {
+  getCommandsHistory: (instanceId: string) => Promise<CommandExecutionUI[]>
+}
+
+export const useCommandsHistory = ({
```
valkirilov: This is the main hook that should be used when working with Command History data:
- it defines all the necessary CRUD operations for dealing with this type of data
- and it does the magic to target the correct data source, based on a feature flag, hiding these implementation details from its consumers

PS: For now, it is only getting data, but I'll add the rest later.

redisinsight/ui/src/services/commands-history/database/interface.ts
```diff
@@ -0,0 +1,18 @@
+import { AxiosError } from 'axios'
+import {
+  CommandExecutionType,
+  CommandExecutionUI,
+} from 'uiSrc/slices/interfaces'
+
+export interface CommandsHistoryDatabase {
```
valkirilov, line 9: This is the interface that every service should comply with. It's basic CRUD, simple as that. And it's independent from the data source behind it.

redisinsight/ui/src/services/commands-history/database/CommandsHistorySQLite.ts
```diff
@@ -0,0 +1,46 @@
...
+import {
+  CommandExecution,
+  CommandExecutionType,
+  CommandExecutionUI,
+} from 'uiSrc/slices/interfaces'
+
+import { mapCommandExecutionToUI } from '../utils/command-execution.mapper'
+import { CommandsHistoryDatabase, CommandHistoryResult } from './interface'
+
+export class CommandsHistorySQLite implements CommandsHistoryDatabase {
```
valkirilov, line 21: This is an i(example) implementation of the service that should work with the Commands History stored in the SQLite database. It should follow the common interface and keep all of the business login encapsulated inside it.

redisinsight/ui/src/services/commands-history/hooks/useCommandsHistory.ts
```diff
@@ -0,0 +1,56 @@
...
+      instanceId,
+      commandExecutionType,
+    )
+
+    if (error) {
+      dispatch(addErrorNotification(error))
+    }
+
+    return data || []
+  }
```
KrumTy: Unmemoized callback, later used in a useEffect. If set as a dependency in the array (as should) would cause infinite loop likely.

But then again, I think this can be extracted outside react lifecycle/hooks. commandExecutionType can be passed as a parameter (to the function or to a generator), redux store access and dispatch - it should be possible to trigger outside as well. I don't see a benefit in having this as a hook atm

redisinsight/ui/src/services/commands-history/utils/command-execution.mapper.ts
```diff
@@ -0,0 +1,12 @@
+import { EMPTY_COMMAND } from 'uiSrc/constants'
+import { CommandExecution, CommandExecutionUI } from 'uiSrc/slices/interfaces'
+
+export const mapCommandExecutionToUI = (
+  item: CommandExecution,
+): CommandExecutionUI => {
+  return {
+    ...item,
+    command: item.command || EMPTY_COMMAND,
+    emptyCommand: !item.command,
```
KrumTy, line 10: does it make sense to transform the data on api level instead?

valkirilov: What do you mean by "on API level"? It is a transformation that we currently have, so I have exported it as a util and reuse it in one more place.

KrumTy: I saw the data comes from apiService.get, so if it's coming from api, can it be transformed before sending the response?

redisinsight/ui/src/config/default.ts
```diff
@@ -69,6 +69,9 @@ export const defaultConfig = {
     localResourcesBaseUrl: process.env.RI_LOCAL_RESOURCES_BASE_URL,
     useLocalResources: booleanEnv('RI_USE_LOCAL_RESOURCES', false),
     indexedDbName: process.env.RI_INDEXED_DB_NAME || 'RI_LOCAL_STORAGE',
+    vectorSearchIndexedDbName:
```
valkirilov, line 72: Adding a separate IndexedDB for the Search page (as advised [here](https://github.com/redis/RedisInsight/pull/5043#discussion_r2415973113)). 

It will use the same interface like the Workbench, but the Commands History related to these pages will live independently.

redisinsight/ui/src/services/vectorSearchHistoryStorage.ts
```diff
@@ -0,0 +1,9 @@
+import { getConfig } from 'uiSrc/config'
+import { WorkbenchStorage } from './workbenchStorage'
+
+const riConfig = getConfig()
+
+export const vectorSearchCommandsHistoryStorage = new WorkbenchStorage(
```
valkirilov, line 6: Adding a separate IndexedDB for the Search page (as advised [here](https://github.com/redis/RedisInsight/pull/5043#discussion_r2415973113)). 

It will use the same interface like the Workbench, but the Commands History related to these pages will live independently.

redisinsight/ui/src/services/workbenchStorage.ts
```diff
@@ -283,9 +283,12 @@ export const wbHistoryStorage = new WorkbenchStorage(
 
 type CommandHistoryType = CommandExecution[]
 
-export async function getLocalWbHistory(dbId: string) {
+export async function getLocalWbHistory(
```
valkirilov, line 286: Extending all the methods to support picking the database instance (as advised [here](https://github.com/redis/RedisInsight/pull/5043#discussion_r2415973113)). 

The end goal is to keep it simple yet generic, so both the Search and Workbench pages will use the same methods for CRUD related to their Command History stored in IndexedDB.

redisinsight/ui/src/services/commands-history/commandsHistoryService.ts
```diff
@@ -0,0 +1,119 @@
...
+  CommandExecutionUI,
+  ResultsMode,
+  RunQueryMode,
+} from 'uiSrc/slices/interfaces'
+
+import { CommandsHistorySQLite } from './database/CommandsHistorySQLite'
+import { CommandsHistoryDatabase } from './database/interface'
+import { CommandsHistoryIndexedDB } from './database/CommandsHistoryIndexedDB'
+
+export class CommandsHistoryService {
```
valkirilov, line 16: This is the main service that should be used when working with Command History data:

- it defines all the necessary CRUD operations for dealing with this type of data
- and it does the magic to target the correct data source, based on a feature flag, hiding these implementation details from its consumers

redisinsight/ui/src/services/commands-history/hooks/useCommandsHistory.ts
valkirilov: As discussed offline, the only purpose behind this hook was to get the feature flag from Redux. There is no need to have a hook at all. So, we agreed that we should change it to a regular class instead, and it's been reimplemented. Thanks for your feedback.

redisinsight/ui/src/services/commands-history/utils/command-execution.mapper.ts
valkirilov: As discussed offline, this point is valid, and perhaps we don’t need to transform the data at all. However, this implementation detail was introduced a long time ago, and in this pull request, I’m making it reusable instead of copying and pasting the same code. 

While potential changes or the complete removal of this logic are outside the scope of this pull request, they’re welcome.

redisinsight/ui/src/services/workbenchStorage.ts
KrumTy: if all of these methods require a WorkbenchStorage instance as a first parameter, then why not make them methods of WorkbenchStorage class?

valkirilov: I'm not sure what the original idea was behind this implementation back then, when it was originally introduced. Maybe it was not considered that one day it would serve two different pages.

In my previous pull request, I have changed the original implementation of the `WorkbenchStorage` and it was advised [here](https://github.com/redis/RedisInsight/pull/5043#discussion_r2415973113) that it's not the correct approach to extend it. That's why I have followed the provided suggestion in this implementation, simply by extending the helper methods instead of the original class.

redisinsight/ui/src/services/commands-history/database/CommandsHistoryIndexedDB.ts
```diff
@@ -0,0 +1,131 @@
...
+import {
+  addCommands,
+  clearCommands,
+  findCommand,
+  getLocalWbHistory,
+  removeCommand,
+  wbHistoryStorage,
+  WorkbenchStorage,
+} from 'uiSrc/services/workbenchStorage'
+import { C } from 'msw/lib/glossary-2792c6da'
```
dantovska: nit: I think we don't need this?

dantovska (dismissed):

redisinsight/ui/src/services/commands-history/database/CommandsHistoryIndexedDB.ts
valkirilov: Thanks for spotting this, now it's removed 🙌

PR #5069, opened 2025-10-17
fix test runner

pd-redis:
This pull request focuses on updating the MSW (Mock Service Worker) API usage throughout the codebase. It also upgrades related dependencies and adjusts the Jest configuration to support these changes.

**MSW API Modernization and Type Cleanup:**

* update MSW version

* updated calls to follow current API of MSW

**Dependency and Jest Configuration Updates:**

* Upgraded `msw` to version `^2.11.5` and added `jest-fixed-jsdom` 

* Changed Jest configuration to use `jest-fixed-jsdom` as the test environment, added `until-async` to the `transformIgnorePatterns`, and included `customExportConditions` in `testEnvironmentOptions`. All of this is to support MSW 2 with jestdom

**Code Cleanup:**

* Removed unused re-export of `FormField` in `FormField.ts`.

### DIFF
--- a/redisinsight/ui/src/pages/home/components/db-status/DbStatus.spec.tsx
+++ b/redisinsight/ui/src/pages/home/components/db-status/DbStatus.spec.tsx
@@ -18,8 +18,22 @@ jest.mock('uiSrc/telemetry', () => ({
 
 const mockedProps = mock<Props>()
 const daysToMs = (days: number) => days * 60 * 60 * 24 * 1000
+let mockDate: Date
 
 describe('DbStatus', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+
+    // Set up fake timers
+    jest.useFakeTimers()
+    mockDate = new Date('2024-11-22T12:00:00Z')
+    jest.setSystemTime(mockDate)
+  })
+
+  afterEach(() => {
+    jest.useRealTimers()
+  })
+
   it('should render', () => {
     expect(render(<DbStatus {...mockedProps} />)).toBeTruthy()
   })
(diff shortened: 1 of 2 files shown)


redisinsight/ui/src/mocks/handlers/analytics/clusterDetailsHandlers.ts
```diff
@@ -11,12 +11,13 @@ import { Database as DatabaseInstanceResponse } from 'apiSrc/modules/database/mo
 
 export const INSTANCE_ID_MOCK = 'instanceId'
 
-const handlers: RestHandler[] = [
+const handlers: HttpHandler[] = [
   // useGetClusterDetailsQuery
-  rest.get<DatabaseInstanceResponse[]>(
+  http.get<any, DatabaseInstanceResponse[]>(
```
valkirilov, line 16: Looking at the `http.get` call, I'm a bit confused about the generic types. From what I understand about MSW's type signature, the generics are `<PathParams, ResponseBodyType, RequestBodyType>`. So:

1. Path params (`any`)
Since we're not using any path parameters in this URL pattern, could we use `Record<string, never>` or just omit it instead of `any`?

2. Response body type (`DatabaseInstanceResponse[]`) 
We're actually returning `CLUSTER_DETAILS_DATA_MOCK` which is typed as `ClusterDetails` (a single object, not an array). Shouldn't this be:
   ```typescript
   http.get<Record<string, never>, ClusterDetails>(...)
   ```

Or the `ClusterDetails` interface is actually a subset of `DatabaseInstanceResponse`, and it's fine.

3. General question 
When I remove these type parameters entirely, I don't see any TypeScript errors. Are these types actually being validated somewhere, or are they more for documentation purposes? Maybe I'm missing some context about how MSW uses these in the testing setup?

I'm just trying to make sure I understand the pattern correctly.

pd-redis: since this is purely test code, and TS types have no effect on actual results, I am in favor of omitting them overall

pawelangelow (approved):

PR #5070, opened 2025-10-17
RI-7219: remove "trial" from "free trial" labels

KrumTy:
### Description

copy changes

### DIFF
--- a/redisinsight/api/src/modules/cloud/database/cloud-database.capi.service.ts
+++ b/redisinsight/api/src/modules/cloud/database/cloud-database.capi.service.ts
@@ -87,7 +87,7 @@ export class CloudDatabaseCapiService {
   }
 
   /**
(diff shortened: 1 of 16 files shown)


redisinsight/api/src/modules/cloud/database/cloud-database.capi.service.ts
```diff
@@ -87,7 +87,7 @@ export class CloudDatabaseCapiService {
   }
 
   /**
-   * Creating free trial database along with subscription if needed
+   * Creating free database along with subscription if needed
```
valkirilov, line 90: I see "auto-replacement" here 😀

KrumTy: It was manual. There's just one instance of the word "trial" remaining now

PR #5095, opened 2025-10-21
RI-7663 Prevent the app from crashing when Redis commands are not parsable

valkirilov:
# Description 

Prevent the app from crashing when Redis commands are not parsable

| Before | After |
| - | - |
image|image

# Root Cause Analysis

We fetch all the available Redis commands and provide autocomplete intelligence in the Monco editor on the Workbench page. And we do get all of this information from external sources. A recent change in the RediSearch repository caused our app to crash because we can't parse the new command properly. With this pull request, we aim to improve the handling of these commands and simply disable the autocomplete feature, instead of causing the whole application to crash.

### DIFF
--- a/redisinsight/ui/src/components/monaco-laguages/MonacoLanguages.tsx
+++ b/redisinsight/ui/src/components/monaco-laguages/MonacoLanguages.tsx
@@ -67,14 +67,18 @@ const MonacoLanguages = () => {
       name?.startsWith(ModuleCommandPrefix.RediSearch),
     )
 
-    monaco.languages.setMonarchTokensProvider(
-      MonacoLanguage.RediSearch,
-      getRediSearchSubRedisMonarchTokensProvider(REDIS_SEARCH_COMMANDS),
-    )
-    monaco.languages.setMonarchTokensProvider(
-      MonacoLanguage.Redis,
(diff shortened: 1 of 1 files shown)


redisinsight/ui/src/components/monaco-laguages/MonacoLanguages.tsx
```diff
@@ -67,14 +67,18 @@ const MonacoLanguages = () => {
...
-    )
+    try {
+      monaco.languages.setMonarchTokensProvider(
+        MonacoLanguage.RediSearch,
+        getRediSearchSubRedisMonarchTokensProvider(REDIS_SEARCH_COMMANDS),
+      )
+      monaco.languages.setMonarchTokensProvider(
+        MonacoLanguage.Redis,
+        getRedisMonarchTokensProvider(REDIS_COMMANDS),
+      )
```
ArtemHoruzhenko, line 78: I don't think it is a good pattern. We must have different try/catch block for a single tokens provider. otherwise we might not setup monaco-editor for commands without issues.

valkirilov: Good point, thanks! Here it is:
- https://github.com/redis/RedisInsight/pull/5098

PR #5097, opened 2025-10-21
Release/2.70.3

valkirilov:
# Description 

Critical bug fix, to prevent the app from crashing when Redis commands are not parsable.


dantovska (approved):

KrumTy:
can we also include the actual fix from this PR? https://github.com/redis/RedisInsight/pull/5085
it's best to extract the formatter as a util and maybe add some unit tests..

PR #5104, opened 2025-10-23
RI-7637 Loading indicator for tables in Browser keys details

valkirilov:
# Description

Bring back the loading indicator for tables in Browser Keys details.

https://github.com/user-attachments/assets/8f8ce8d4-5054-4afa-87e6-6bd396a99c55

# How it was tested

1. Go to **Databases** and open a connection to an existing database instance, or create a new one
2. Go to the **Browser** page
3. Open an existing key of type SET, SORTED SET, LIST, or HASH (all of which come with a table with members)
4. Click on the refresh button in the header of the Key Details panel

Note: Usually, the indicator is shown, but it disappears very quickly if your connection is fast enough. You can make it stay for longer if you throttle your internet connection via the Dev Tools -> Network -> Throttle 

image

### DIFF
--- a/redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
+++ b/redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
@@ -66,12 +66,11 @@ interface LoaderContainerProps {
   children?: ReactNode
   style?: React.CSSProperties
   className?: string
-  isAbsolutePositioned?: boolean
+  absolute?: boolean
 }
 
 export const LoaderContainer = styled.div<LoaderContainerProps>`
-  position: ${({ isAbsolutePositioned }) =>
-    isAbsolutePositioned ? 'absolute' : 'relative'};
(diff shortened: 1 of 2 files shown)


dantovska (dismissed):

redisinsight/ui/src/components/base/display/progress-bar/progress-bar-loader.styles.ts
```diff
@@ -66,10 +66,13 @@ interface LoaderContainerProps {
   children?: ReactNode
   style?: React.CSSProperties
   className?: string
+  isAbsolutePositioned?: boolean
```
pd-redis: I don't like this name

pd-redis: maybe just `absolute`?!

valkirilov: Done, simplified.

PR #5118, opened 2025-10-29
Expose router history

KrumTy:
## Description 

Issue:
We currently cannot navigate outside of react lifecycle/hooks. E.g. if we want to navigate from a method, we need to call `const history = useHistory()` in a component and somehow pass the history as a prop.
Use case - we have a lot of cells in a table and we don't want to call useHistory for each one of them if we can access the object directly.

Solution:
- [BrowserRouter](https://v5.reactrouter.com/web/api/BrowserRouter) is wrapper over ReactRouter that sets up history under the hook, so the new implementation should behave identically, only we have access to the history object outside of react hooks
- history from the store and test-store was not used, so I removed those

### DIFF
--- a/redisinsight/ui/src/slices/store.ts
+++ b/redisinsight/ui/src/slices/store.ts
@@ -1,4 +1,3 @@
-import { createBrowserHistory } from 'history'
 import { configureStore, combineReducers } from '@reduxjs/toolkit'
 
 import { getConfig } from 'uiSrc/config'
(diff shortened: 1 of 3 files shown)


redisinsight/ui/src/slices/store.ts
```diff
@@ -59,8 +58,6 @@ import tagsReducer from './instances/tags'
 
 const riConfig = getConfig()
 
-export const history = createBrowserHistory()
```
KrumTy, line 62: this was not used and also it creates another instance - not the one from BrowserRouter, so it's also useless

PR #5119, opened 2025-10-29
Fix tests

ArtemHoruzhenko:
- Properly msw server to return success response for unhandled requests temporary
- Fix store issue in RediSearchIndexesList.spec.tsx


pd-redis (approved):
nice

PR #5124, opened 2025-10-30
[PoC] handle icon theme implicitly

KrumTy:
## Description

Having to access theme every time you render an icon is bad practice imo. This should be handled implicitly, in the icon component.
This is a PoC allows light+dark icon to be provided when defining an icon component with a single example (there are many that can be refactored afterwards)

### DIFF
--- a/redisinsight/ui/src/components/base/icons/iconRegistry.tsx
+++ b/redisinsight/ui/src/components/base/icons/iconRegistry.tsx
@@ -280,9 +280,7 @@ export const RedisJSONLightIcon = createIconComponent(RedisJSONLight)
 export const RedisSearchDarkIcon = createIconComponent(RedisSearchDark)
 export const RedisSearchLightIcon = createIconComponent(RedisSearchLight)
 export const RediStackDarkLogoIcon = createIconComponent(RediStackDarkLogoSvg)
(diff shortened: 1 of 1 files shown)


redisinsight/ui/src/components/base/icons/iconRegistry.tsx
```diff
@@ -270,9 +280,11 @@ export const RedisJSONLightIcon = createIconComponent(RedisJSONLight)
 export const RedisSearchDarkIcon = createIconComponent(RedisSearchDark)
 export const RedisSearchLightIcon = createIconComponent(RedisSearchLight)
 export const RediStackDarkLogoIcon = createIconComponent(RediStackDarkLogoSvg)
-export const RediStackDarkMinIcon = createIconComponent(RediStackDarkMinSvg)
 export const RediStackLightLogoIcon = createIconComponent(RediStackLightLogoSvg)
-export const RediStackLightMinIcon = createIconComponent(RediStackLightMinLight)
+export const RediStackMinIcon = createIconComponent(
+  RediStackLightMinLight,
+  RediStackDarkMinSvg,
+)
```
KrumTy, line 287: POC - when defining an icon that has multiple colors and its svg cant be switched via `{ fill: currentColor; }` we have are provided with 2 svgs; with this helper we provide both. If 2nd argument is omitted, same icon will be used for both themes

PR #5126, opened 2025-10-30
RI-7463 sentinel databases rerender

pd-redis:
Masters:
image

Result:
image


.storybook/preview-head.html
```diff
@@ -13,6 +13,13 @@
       min-height: 100vh;
     }
 
+    #storybook-root {
+      position: absolute;
+      inset: 0;
+      display: flex;
+      flex-direction: column;
+    }
```
dantovska, line 21: what is this style for, can't we avoid such styles?

.storybook/preview.tsx
```diff
@@ -18,12 +19,13 @@ import { type Theme } from 'uiSrc/components/base/theme/types'
...
 import { store } from 'uiSrc/slices/store'
 import Router from 'uiSrc/Router'
+import { StyledContainer } from './helpers/styles'
 
 const parameters: Parameters = {
   parameters: {
     layout: 'centered',
   },
-  actions: { argTypesRegex: '^on[A-Z].*' },
+  actions: { argTypesRegex: '^on.*' },
```
dantovska, line 28: can you add some explanation what this regex means?

redisinsight/ui/src/components/auto-discover/Header.tsx
```diff
@@ -0,0 +1,62 @@
...
+    <Row align="center" justify="between" grow={false}>
+      <Col align="start" justify="start">
+        <EmptyButton
+          icon={ArrowLeftIcon}
+          onClick={onBack}
+          data-testid="btn-back-adding"
+        >
+          {backButtonText}
+        </EmptyButton>
+        <Spacer size="m" />
```
dantovska: I see you also rely on the Spacer a lot. Is it not possible to solve this with gap on the Col?

redisinsight/ui/src/components/auto-discover/index.ts
```diff
@@ -26,12 +26,26 @@ export const SearchContainer = styled(FlexItem)`
...
 `
-export const Footer = styled(FlexItem).attrs({
-  grow: false,
-})`
+export const Footer = styled(FlexItem).attrs<{
+  grow?: boolean | number
+  padding?: React.ComponentProps<typeof FlexItem>['padding']
+}>(({ grow, padding }) => ({
+  grow: grow ?? false,
+  padding: padding ?? 6,
```
dantovska, line 34: I think this padding pattern is confusing .. is there a reason we are not using the space from the theme?

redisinsight/ui/src/components/auto-discover/index.ts
```diff
@@ -26,12 +26,26 @@ export const SearchContainer = styled(FlexItem)`
...
+}))`
   border-top: 1px solid
     ${({ theme }: { theme: Theme }) => theme.semantic.color.border.neutral400};
 `
+
+export const DatabaseContainer = styled(Col)`
+  position: relative;
+  padding: 20px 14px 0 14px;
```
dantovska: can we use space from the core theme and not hard coded pixels?

redisinsight/ui/src/components/base/forms/buttons/Button.tsx
```diff
@@ -59,7 +61,11 @@ export const IconSizes = {
   medium: '20px',
   large: '24px',
 }
-
+const Wrapper = styled.div`
+  svg {
+    display: block;
+  }
+`
```
dantovska, line 68: why do we need this display block style? I think we have agreed to avoid this usages...

redisinsight/ui/src/components/base/layout/flex/flex.styles.ts
```diff
@@ -284,33 +284,19 @@ export const VALID_GROW_VALUES = [
...
-  11, // '4.4rem: 55px',
-  12, // '4.8rem: 60px',
-  13, // '6.4rem: 80px',
-] as const
+export type PaddingType =
+  | keyof typeof flexItemStyles.padding
+  | null
+  | undefined
+  | true
+  | false
```
dantovska, line 293: what would a padding: true mean? why are we allowing this as a value?

redisinsight/ui/src/components/base/layout/spacer/spacer.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Spacer } from './spacer'
+import { SpacerProps } from './spacer.styles'
+
+const sizeToValue = {
+  xs: '0.2rem',
+  s: '0.4rem',
+  m: '0.8rem',
+  l: '1.6rem',
+  xl: '2rem',
+  xxl: '2.4rem',
```
dantovska, line 13: I think we agreed to avoid introducing new hard coded values but tend to remove the already used also.

.storybook/preview-head.html
pd-redis: no, it is for ensuring storybook canvas is 100% tall, it does not affect RI in anyway

.storybook/preview.tsx
pd-redis: Sure, you can read about it here: https://storybook.js.org/docs/essentials/actions#automatically-matching-args

redisinsight/ui/src/components/auto-discover/Header.tsx
pd-redis: Sure, good catch

redisinsight/ui/src/components/auto-discover/index.ts
pd-redis: ths is the padding prop from FlexItem component, it uses the theme values

redisinsight/ui/src/components/auto-discover/index.ts
pd-redis: We can, but for max widths etc we don't have values in theme

redisinsight/ui/src/components/base/forms/buttons/Button.tsx
pd-redis: no we haven't :)
we need svgs insife button icons to be targeted like this, because we can't target them with styled components
this is why I have added this Wrapper, no other reason

pd-redis: and we need svg to be display block, because if we pass an icon wrapped in tooltip, it gets miss-aligned

redisinsight/ui/src/pages/autodiscover-cloud/utils.tsx
```diff
@@ -1,66 +1,33 @@
...
+ */
 export const getSelectionColumn = <T extends object>({
   size = 50,
-  id = 'select-col',
-  enableResizing = false,
-  canSelectRow = (_row: RowDefinition<T>) => true,
-  setSelection,
-  onSelectionChange,
-}: Props): ColumnDefinition<T> => {
+  id = 'row-selection',
```
KrumTy, line 14: it's worth using this as data-test id as well

redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases-result/useSentinelDatabasesResultConfig.tsx
```diff
@@ -377,8 +324,95 @@ export const useSentinelDatabasesResultConfig = () => {
...
+
+export const useSentinelDatabasesResultConfig = () => {
+  const [items, setItems] = useState<ModifiedSentinelMaster[]>([])
+  const [isInvalid, setIsInvalid] = useState(true)
+
+  const { data: masters } = useSelector(sentinelSelector)
+
+  const countSuccessAdded = masters.filter(
+    ({ status }) => status === AddRedisDatabaseStatus.Success,
+  )?.length
```
KrumTy, line 93: can useMemo here

redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases-result/useSentinelDatabasesResultConfig.tsx
```diff
@@ -377,8 +324,95 @@ export const useSentinelDatabasesResultConfig = () => {
...
+
+  const dispatch = useDispatch()
+  const history = useHistory()
+
+  setTitle('Redis Sentinel Primary Groups Added')
```
KrumTy: maybe set it once in the useEffect bellow

redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases-result/useSentinelDatabasesResultConfig.tsx
```diff
@@ -377,8 +324,95 @@ export const useSentinelDatabasesResultConfig = () => {
...
+
+  const handleBackAdding = () => {
+    dispatch(resetLoadedSentinel(LoadedSentinel.MastersAdded))
+    history.push(Pages.home)
+  }
+
+  const handleViewDatabases = () => {
+    dispatch(resetDataSentinel())
+    history.push(Pages.home)
+  }
```
KrumTy: optional - if you use history from Router.tsx and dispatch from store.ts you can extract those outside the component

redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases-result/useSentinelDatabasesResultConfig.tsx
```diff
@@ -377,8 +324,95 @@ export const useSentinelDatabasesResultConfig = () => {
...
+
+        return { ...item, [field]: value }
+      }),
+    )
+  }
+
+  const columns: ColumnDef<ModifiedSentinelMaster>[] = useMemo(() => {
+    return colFactory(
+      handleChangedInput,
+      handleAddInstance,
```
KrumTy, line 153: handleAddInstance - memoized, not as dependency 
handleChangedInput - not memoized, not as dependency

redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases/components/SentinelDatabases/SentinelDatabases.tsx
```diff
@@ -48,13 +52,32 @@ const notFoundMsg = 'Not found.'
...
+        (acc, item) => {
+          if (item.id) {
+            acc[item.id as string] = true
+          }
+          return acc
+        },
+        {} as Record<NonNullable<ModifiedSentinelMaster['id']>, boolean>,
+      ),
+    )
+  }, [selection])
```
KrumTy: so table calls onSelectionChange with an object, then some magic happens, then selection array is updated and it is passed down to this component so the useEffect can process it and update the table selection? 

I might be getting it wrong, the flow seems confusing

redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases-result/useSentinelDatabasesResultConfig.tsx
pd-redis: yeah, but overhead of useMemo vs filtering sub 100 items ... not sure its worth it.
If we get some data from PMs for people discovering 100s of masters, it could be worth it

redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases/components/SentinelDatabases/SentinelDatabases.tsx
pd-redis: Well ....
that's leftover, removing

redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases-result/useSentinelDatabasesResultConfig.tsx
```diff
@@ -377,8 +337,93 @@ export const useSentinelDatabasesResultConfig = () => {
...
+  const countSuccessAdded = masters.filter(
+    ({ status }) => status === AddRedisDatabaseStatus.Success,
+  )?.length
+
+  const dispatch = useDispatch()
+  const history = useHistory()
+
+  useEffect(() => {
+    setTitle('Redis Sentinel Primary Groups Added')
+    if (!masters.length) {
```
KrumTy: does the title need to change back on component unmount? 
overall, quite weird mechanism we have here

redisinsight/ui/src/components/base/layout/flex/flex.styles.ts
pd-redis: true === '4', meaning it is the default padding applied if you use `<FlexItem padding/>`

pd-redis: But I also think, that this PR is not related to changing layout components, we can and should do this separately

KrumTy (approved):
maybe it's worth extracting some of the styled components to `.styles.ts` files, other than that LGTM

dantovska (dismissed):

PR #5130, opened 2025-10-31
DEV: Improve code style with turning on few ESlint rules

pawelangelow:
It's easier if you go commit by commit.

1. Turning on `prettier/prettier` rules - results in empty lines, adjusted imports, etc... `NO` business logic change here.
2. Turning on `@typescript-eslint/***` results in `one` business logic change - `redisinsight/api/src/main.ts` file and the rest are cosmetics - unused variables deleted, naming, etc...

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -261,9 +261,6 @@ module.exports = {
       rules: {
         semi: 'off',
         '@typescript-eslint/semi': 'off',
-        '@typescript-eslint/no-unused-vars': 'off',
-        '@typescript-eslint/no-use-before-define': 'off',
-        '@typescript-eslint/no-unused-expressions': 'off',
         'sonarjs/no-identical-functions': 'off',
         'sonarjs/prefer-immediate-return': 'off',
         'sonarjs/no-duplicate-string': 'off',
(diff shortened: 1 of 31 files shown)


valkirilov (approved):
Nice! Now, when I see how this PR fixed some limiting issues, I wonder whether it's a good idea to actually turn on the linter in the pipeline, so we don't produce new issues in the meantime. 

The rest of the rules are still disabled, so it won't block us at the moment - it will simply guard us from producing new issues.

pawelangelow:
> Nice! Now, when I see how this PR fixed some limiting issues, I wonder whether it's a good idea to actually turn on the linter in the pipeline, so we don't produce new issues in the meantime.
> 
> The rest of the rules are still disabled, so it won't block us at the moment - it will simply guard us from producing new issues.

That's for sure. When I have some time, I'll open a PR about it, as it requires a bit of changes of our whole process. What I imagine is:
- install dependencies
- run linter + prettier
- fail if previous step failed
- continue with tests with cached dependencies (so we don't wait again for `yarn install`)

but it will be explained in a feature PR. This one at least improves the Dev experience over the backend.

PR #5139, opened 2025-11-04
RI-7691: Polish Profiler styles

pawelangelow:
| Before    | After |
| -------- | ------- |
|  Screenshot 2025-11-04 at 10 19 36|  Screenshot 2025-11-04 at 10 18 59 |
|  Screenshot 2025-11-04 at 10 19 50 | Screenshot 2025-11-04 at 10 18 42 |
| Screenshot 2025-11-04 at 10 19 29 |  Screenshot 2025-11-04 at 10 19 17 |

Currently, the icon that should be in front of the result text is missing, so I added the most suitable one, IMO, from the existing set.

### DIFF
--- a/redisinsight/ui/src/components/monitor/Monitor/Monitor.tsx
+++ b/redisinsight/ui/src/components/monitor/Monitor/Monitor.tsx
@@ -58,7 +58,7 @@ const Monitor = (props: Props) => {
         <RiImage
           src={ProfilerImage}
           alt="Profiler"
-          style={{ userSelect: 'none' }}
+          style={{ userSelect: 'none', pointerEvents: 'none' }}
         />
         <Spacer size="l" />
         <Text>


redisinsight/ui/src/assets/img/profiler/magnifier.svg
KrumTy, line 1: I think you should be provided with a dark theme svg as well

pd-redis: We should stirve to use one image for both, if possible

redisinsight/ui/src/components/monitor/Monitor/Monitor.tsx
```diff
@@ -5,16 +5,20 @@ import AutoSizer from 'react-virtualized-auto-sizer'
...
-import { ColorText } from 'uiSrc/components/base/text'
+import { Col, FlexItem, Row } from 'uiSrc/components/base/layout/flex'
+import { PrimaryButton } from 'uiSrc/components/base/forms/buttons'
+import { ColorText, Text, Title } from 'uiSrc/components/base/text'
 import { SwitchInput } from 'uiSrc/components/base/inputs'
 import { RiIcon } from 'uiSrc/components/base/icons/RiIcon'
 import MonitorLog from '../MonitorLog'
 import MonitorOutputList from '../MonitorOutputList'
 
+import ProfilerImage from 'uiSrc/assets/img/profiler/magnifier.svg'
```
pd-redis, line 16: Why not import it as SVG component and output as such?

redisinsight/ui/src/components/monitor/Monitor/Monitor.tsx
```diff
@@ -43,58 +47,63 @@ const Monitor = (props: Props) => {
...
-              style={{ paddingLeft: 4 }}
-              data-testid="monitor-warning-message"
+    <Row
+      align="center"
+      style={{ margin: 48 }}
+      gap="xxl"
+      data-testid="monitor-not-started"
+    >
+      <StyledImagePanel align="center">
+        <StyledImage as="img" src={ProfilerImage} alt={'asd'} />
```
pd-redis: why do you need `as`? Also, my comment above stands

redisinsight/ui/src/assets/img/profiler/magnifier.svg
pawelangelow: IMO, we shouldn't block the PR with this. If we have such an image in the future, we'll address it in a separate one.

redisinsight/ui/src/components/monitor/Monitor/Monitor.tsx
pawelangelow: Not sure I got you, but check https://github.com/redis/RedisInsight/pull/5139/commits/04bd33a4c57698ebfc500e40c477f66bdaf6f148

redisinsight/ui/src/components/monitor/Monitor/Monitor.tsx
pawelangelow: Because I was using `` instead of `<RIImage />` and TS complained about `as`.

redisinsight/ui/src/components/monitor/Monitor/Monitor.tsx
pd-redis: What I meant is:
```typescript
import ProfilerImage from 'uiSrc/assets/img/profiler/magnifier.svg?react'

// ---

<ProfilerImage
          alt="Profiler"
          style={{ userSelect: 'none' }}
        />

```

Basically how icons work now

PR #5146, opened 2025-11-05
RI-7399: DatabasesList (v2)

KrumTy:
# What

Adds DatabasesList component that uses redis-ui table. It is behind a feature flag so that old DB List can be enabled as well.

# Testing

## Demo

https://github.com/user-attachments/assets/7e24effe-0772-4775-9e1f-6d5dd18a66ca

## Columns and resizing demo:

https://github.com/user-attachments/assets/b33556c7-072a-4b23-916a-c585eb75b65c

### DIFF
--- a/redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellControls/DatabasesListCellControls.spec.tsx
+++ b/redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellControls/DatabasesListCellControls.spec.tsx
@@ -59,7 +59,7 @@ const renderWithProvider = () =>
     <HomePageDataProviderProvider>
       <>
         <Observer />
-        {React.createElement(DatabasesListCellControls as any, { row })}
+        <DatabasesListCellControls {...({ row } as any)} />
       </>
     </HomePageDataProviderProvider>,
   )
--- a/redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellControls/methods/handlers.spec.ts
+++ b/redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellControls/methods/handlers.spec.ts
@@ -1,6 +1,7 @@
 import { TelemetryEvent, sendEventTelemetry } from 'uiSrc/telemetry'
 import { dispatch } from 'uiSrc/slices/store'
 import { Instance } from 'uiSrc/slices/interfaces'
+import { DBInstanceFactory } from 'uiSrc/mocks/factories/database/DBInstance.factory'
 
 import {
   handleDeleteInstances,
@@ -46,15 +47,7 @@ jest.mock('uiSrc/slices/instances/instances', () => ({
     mockSetEditedInstance(...(args as any)),
 }))
 
-const instance: Instance = {
-  id: '1',
-  name: 'A',
-  host: 'h1',
-  port: 6379,
-  provider: 'LOCALHOST',
-  modules: [],
-  version: null,
-}
+const instance: Instance = DBInstanceFactory.build()
 
 describe('DatabasesListCellControls handlers', () => {
   beforeEach(() => {
@@ -80,8 +73,8 @@ describe('DatabasesListCellControls handlers', () => {
       expect.objectContaining({
         event: TelemetryEvent.CONFIG_DATABASES_SINGLE_DATABASE_DELETE_CLICKED,
         eventData: expect.objectContaining({
-          databaseId: '1',
-          provider: 'LOCALHOST',
+          databaseId: instance.id,
+          provider: instance.provider,
         }),
       }),
     )
@@ -102,8 +95,8 @@ describe('DatabasesListCellControls handlers', () => {
       expect.objectContaining({
         event: TelemetryEvent.CONFIG_DATABASES_DATABASE_EDIT_CLICKED,
         eventData: expect.objectContaining({
-          databaseId: '1',
-          provider: 'LOCALHOST',
+          databaseId: instance.id,
+          provider: instance.provider,
         }),
       }),
     )
@@ -121,8 +114,8 @@ describe('DatabasesListCellControls handlers', () => {
       expect.objectContaining({
         event: TelemetryEvent.CONFIG_DATABASES_DATABASE_MANAGE_TAGS_CLICKED,
         eventData: expect.objectContaining({
-          databaseId: '1',
-          provider: 'LOCALHOST',
+          databaseId: instance.id,
+          provider: instance.provider,
         }),
       }),
     )
--- a/redisinsight/ui/src/pages/home/components/databases-list/hooks/useDatabaseListData.spec.ts
+++ b/redisinsight/ui/src/pages/home/components/databases-list/hooks/useDatabaseListData.spec.ts
@@ -1,4 +1,5 @@
 import { act } from '@testing-library/react'
+
 import {
   mockStore,
   initialStateDefault,
@@ -6,6 +7,8 @@ import {
 } from 'uiSrc/utils/test-utils'
 import { DatabaseListColumn } from 'uiSrc/constants'
 import { Instance } from 'uiSrc/slices/interfaces'
+import { DBInstanceFactory } from 'uiSrc/mocks/factories/database/DBInstance.factory'
+
 import { SELECT_COL_ID, ENABLE_PAGINATION_COUNT } from '../DatabasesList.config'
 
 import useDatabaseListData from './useDatabaseListData'
@@ -44,34 +47,9 @@ const getStoreWith = ({
 }
 
 const mockInstances: Instance[] = [
-  {
-    id: '1',
-    name: 'Instance A',
-    host: 'a.example',
-    port: 6379,
-    modules: [],
-    version: null,
-    visible: true,
-    tags: [{ id: '1', key: 'env', value: 'dev', createdAt: '', updatedAt: '' }],
-  },
-  {
-    id: '2',
-    name: 'Instance B',
-    host: 'b.example',
-    port: 6380,
-    modules: [],
-    version: null,
-    visible: false,
-  },
-  {
-    id: '3',
-    name: 'Instance C',
-    host: 'c.example',
-    port: 6381,
-    modules: [],
-    version: null,
-    visible: true,
-  },
+  DBInstanceFactory.build({ visible: true }),
+  DBInstanceFactory.build({ visible: false }),
+  DBInstanceFactory.build({ visible: true }),
 ]
 
 describe('useDatabaseListData', () => {
--- a/redisinsight/ui/src/pages/home/components/db-status/DbStatus.styles.ts
+++ b/redisinsight/ui/src/pages/home/components/db-status/DbStatus.styles.ts
@@ -1,5 +1,5 @@
 import styled from 'styled-components'
 
-export const IconWrapper = styled.div<any>`
+export const IconWrapper = styled.div<{ children?: React.ReactNode }>`
   margin-left: -${({ theme }) => theme.core.space.space150};
 `
--- a/redisinsight/ui/src/mocks/factories/database/DBInstance.factory.ts
+++ b/redisinsight/ui/src/mocks/factories/database/DBInstance.factory.ts
@@ -1,0 +1,23 @@
+import { Factory } from 'fishery'
+import { faker } from '@faker-js/faker'
+import { ConnectionType, Instance } from 'uiSrc/slices/interfaces'
+
+export const DBInstanceFactory = Factory.define<Instance>(() => ({
+  id: faker.string.uuid(),
+  name: faker.company.name(),
+  host: faker.internet.ip(),
+  port: faker.internet.port(),
+  username: faker.internet.userName(),
+  password: faker.internet.password(),
+  visible: faker.datatype.boolean(),
+  connectionType: faker.helpers.arrayElement([
+    ConnectionType.Standalone,
+    ConnectionType.Cluster,
+    ConnectionType.Sentinel,
+  ]),
+  nameFromProvider: faker.company.name(),
+  modules: [],
+  version: faker.system.semver(),
+  lastConnection: faker.date.past(),
+  provider: faker.company.name(),
+}))


pd-redis:
One comment on the interactions part, in my opinion, all clickable elements should have style `cursor: pointer;` to indicate to the user, that they can be clicked

redisinsight/ui/src/components/item-list/components/action-bar/ActionBar.tsx
```diff
@@ -6,45 +6,27 @@ import { CancelSlimIcon } from 'uiSrc/components/base/icons'
...
-          icon={CancelSlimIcon}
-          aria-label="Cancel selecting"
-          onClick={() => onCloseActionBar()}
-          data-testid="cancel-selecting"
-        />
-      </FlexItem>
-    </Row>
-  </div>
+const ActionBar = ({ selectionCount, actions, onCloseActionBar }: Props) => (
+  <Row centered className={styles.container} gap="l">
```
pd-redis: it seems like this is the only scss style left here, can you just wrap Row in styled component and remove dependency on styles.scss?

redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellControls/DatabasesListCellControls.styles.ts
```diff
@@ -0,0 +1,11 @@
+import styled from 'styled-components'
+
+import { IconButton } from 'uiSrc/components/base/forms/buttons'
+
+export const HoverableIconButton = styled(IconButton)`
+  opacity: 0;
+
```
pd-redis, line 8: Great idea. Maybe add some transition, so that opacity change is smooth?

redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellControls/DatabasesListCellControls.tsx
```diff
@@ -0,0 +1,132 @@
...
+    useSelector(appFeatureFlagsFeaturesSelector)
+
+  return (
+    <Row
+      justify="end"
+      align="center"
+      gap="xs"
+      onClick={(e) => e.stopPropagation()}
+    >
+      {databaseManagementFeature?.flag && (
```
pd-redis: Why not use `FeatureFlagComponent` here as well?

redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellHost/DatabasesListCellHost.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { HostPortContainer } from './DatabasesListCellHost.styles'
+
+const DatabasesListCellHost: IDatabaseListCell = ({ row }) => {
+  const instance = row.original
+  const { host, port, id } = instance
+  const text = `${host}:${port}`
+
+  return (
+    <HostPortContainer gap="m" data-testid="host-port">
+      <Text className="copyHostPortText">{text}</Text>
```
pd-redis: Wrap it in styled component?

redisinsight/ui/src/pages/home/components/databases-list/methods/handlers.ts
```diff
@@ -0,0 +1,68 @@
...
+import { store, dispatch } from 'uiSrc/slices/store'
+import { SortingState } from 'uiSrc/components/base/layout/table'
+import { history } from 'uiSrc/Router'
+import { localStorageService } from 'uiSrc/services'
+
+export const handleCheckConnectToInstance = async (instance: Instance) => {
+  const { id, provider, modules } = instance
+  const { contextInstanceId } = appContextSelector(store.getState())
+
+  const connectToInstance = (id = '') => {
```
pd-redis: Not big deal, but this seems like it could be declared outside of the handler

redisinsight/ui/src/pages/home/components/db-status/DbStatus.tsx
```diff
@@ -105,15 +105,13 @@ const DbStatus = (props: Props) => {
...
     return (
-      <RiTooltip
-        content="New"
-        position="top"
-        anchorClassName={cx(styles.statusAnchor)}
-      >
+      <RiTooltip content="New" position="top">
         <div
-          className={cx(styles.status, styles.new)}
+          style={{ marginLeft: -10 }}
```
pd-redis: styled.div`` :)

redisinsight/ui/src/components/item-list/components/action-bar/ActionBar.styles.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const Container = styled(Row)`
+  position: fixed;
+  z-index: 1;
+  left: 50%;
+  transform: translate(-50%, -50%);
+
+  width: 462px;
+  height: 50px;
+  bottom: calc(9vh + 9px);
```
valkirilov, line 13: Can you hint with a comment what is hidden behind these magic numbers, to make it easier to modify in the future (if necessary)? 

PS: I see these styles are copy/paste from the old table and maybe you're not sure about them either?

redisinsight/ui/src/pages/home/components/databases-list/components/BulkItemsActions/BulkItemsActions.tsx
```diff
@@ -0,0 +1,54 @@
...
+  DeleteAction,
+} from 'uiSrc/components/item-list/components'
+
+import {
+  handleDeleteInstances,
+  handleExportInstances,
+} from './methods/handlers'
+import { Instance } from 'uiSrc/slices/interfaces'
+
+const actionMessage = (action: string, length: number) =>
```
valkirilov, line 15: Nit: Do we have a habit of type-hinting the return types as well, or do we usually rely on inference?
```suggestion
const actionMessage = (action: string, length: number): string =>
```

redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellControls/methods/handlers.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    mockDeleteInstancesAction(...(args as any)),
+  fetchEditedInstanceAction: (...args: any[]) =>
+    // @ts-expect-error
+    mockFetchEditedInstanceAction(...(args as any)),
+  setEditedInstance: (...args: any[]) =>
+    // @ts-expect-error
+    mockSetEditedInstance(...(args as any)),
+}))
+
+const instance: Instance = {
```
valkirilov: Note: If you want, you can export this one as a reusable factory, just like we already have such for the RDI Instances ([example](https://github.com/redis/RedisInsight/blob/056ce218d70c86ad92a780d43f8666ce7382569f/redisinsight/ui/src/mocks/rdi/RdiInstance.factory.ts))

redisinsight/ui/src/pages/home/components/databases-list/components/BulkItemsActions/BulkItemsActions.tsx
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  return (
+    <ActionBar
+      selectionCount={items.length}
+      onCloseActionBar={onClose}
+      actions={[
+        <ExportAction<Instance>
+          selection={items}
+          onExport={(_, withSecrets) => {
```
valkirilov, line 35: nit: It's minor, and doesn't affect performance in any way, but I believe the code is more readable when we avoid passing multiple inline callbacks, and have them defined separately:
- it keeps the handlers separate from the rest of the code, and is easier to maintain and modify them in the future
- it keeps the component invocation simpler (here, we have two callback functions, each of with a two-line body, but they can be huge ([example](https://github.com/redis/RedisInsight/pull/5146/files#diff-1585c6d8d3d8963edd41746d53e67730b55397ac6b73eea63e751228049d158fR47)))

```typescript
const handleExportAction = useCallback((_: unknown, withSecrets: boolean) => {
  handleExportInstances(items, withSecrets);
  onClose();
}, [items, onClose, handleExportInstances]);

<ExportAction<Instance>
    selection={items}
    subTitle={actionMessage('exported', items.length)}
    onExport={handleExportAction}
/>
```

redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellControls/DatabasesListCellControls.spec.tsx
```diff
@@ -0,0 +1,121 @@
...
+  }, [openDialog])
+  return null
+}
+
+const renderWithProvider = () =>
+  render(
+    <HomePageDataProviderProvider>
+      <>
+        <Observer />
+        {React.createElement(DatabasesListCellControls as any, { row })}
```
valkirilov: Can't we simply init the component the way we usually do? Why do we need `React.createElement` and the cast to `any`?

```
<DatabasesListCellControls row={row} />
```

redisinsight/ui/src/pages/home/components/databases-list/hooks/useDatabaseListData.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+        shownColumns,
+      },
+    },
+  } as typeof initialStateDefault
+
+  return mockStore(state)
+}
+
+const mockInstances: Instance[] = [
+  {
```
valkirilov: Note: One more place where you can use the factory suggested in [this](https://github.com/redis/RedisInsight/pull/5146/files#r2497892319) comment.

redisinsight/ui/src/pages/home/components/databases-list/hooks/useDatabaseListData.ts
```diff
@@ -0,0 +1,77 @@
...
+    shownColumns,
+  } = useSelector(instancesSelector)
+  const [rowSelection, setRowSelection] = useState<RowSelectionState>({})
+  const resetRowSelection = useCallback(() => {
+    setRowSelection({})
+  }, [])
+
+  const paginationEnabledRef = useRef(false)
+  paginationEnabledRef.current =
+    // Workaround: table breaks if pagination is disabled after it was previously enabled
```
valkirilov, line 31: Ouch, is it something on our end, or is it a Redis UI problem? If it's a generic issue, maybe it's a good idea to report a bug for that behavior?

redisinsight/ui/src/pages/home/components/databases-list/DatabasesList.config.tsx
```diff
@@ -0,0 +1,127 @@
...
+import DatabasesListCellConnectionType from './components/DatabasesListCellConnectionType/DatabasesListCellConnectionType'
+import DatabasesListCellLastConnection from './components/DatabasesListCellLastConnection/DatabasesListCellLastConnection'
+import DatabasesListCellModules from './components/DatabasesListCellModules/DatabasesListCellModules'
+import DatabasesListCellControls from './components/DatabasesListCellControls/DatabasesListCellControls'
+import DatabasesListCellTags from './components/DatabasesListCellTags/DatabasesListCellTags'
+import { TagsCellHeader } from '../tags-cell/TagsCellHeader'
+
+export const SELECT_COL_ID = 'select-col-db'
+
+export const ENABLE_PAGINATION_COUNT = 15
```
valkirilov, line 28: nit: It's not a boolean, so we don't enable anything. And we're not counting anything either.

```suggestion
export const PAGINATION_ITEMS = 15
export const PAGE_SIZE = 15
```

redisinsight/ui/src/pages/home/components/databases-list/DatabasesList.tsx
```diff
@@ -0,0 +1,45 @@
...
+        stripedRows
+        rowSelectionMode="multiple"
+        paginationEnabled={paginationEnabled}
+        onRowClick={handleCheckConnectToInstance}
+        emptyState={emptyMessage}
+        onRowSelectionChange={setRowSelection}
+        rowSelection={rowSelection}
+        onSortingChange={handleSortingChange}
+        defaultSorting={DEFAULT_SORTING}
+        maxHeight="60rem" // this enables vertical scroll
```
valkirilov, line 38: nit: We can export this magic number as a constant?

redisinsight/ui/src/pages/home/components/db-status/DbStatus.styles.ts
```diff
@@ -0,0 +1,5 @@
+import styled from 'styled-components'
+
+export const IconWrapper = styled.div<any>`
```
valkirilov: Why is this `any` here? What do we usually pass to this component so that we can't type-hint it properly?

redisinsight/ui/src/styles/elastic.css
```diff
@@ -460,9 +460,9 @@
   caption,
   tbody,
   tfoot,
-  thead,
+  /* thead,
```
valkirilov, line 463: I thought you decided to remove these, not only comment them temporarily.

redisinsight/ui/src/components/item-list/components/action-bar/ActionBar.styles.ts
KrumTy: yes, those are copied over

also, this component is going to be redesigned (from feedback) so lets not bother

redisinsight/ui/src/pages/home/components/databases-list/components/BulkItemsActions/BulkItemsActions.tsx
KrumTy: Depends? This one is copied over from old component, just renamed. It returns a plain string and has no logic whatsoever.
What I don't like is the text being all over the place, ideally we'd use i18n or something but that's out of scope

redisinsight/ui/src/pages/home/components/databases-list/hooks/useDatabaseListData.ts
KrumTy: it's redis-table related, but I am not sure if it's fixed in later versions; need to check that separately

redisinsight/ui/src/pages/home/components/databases-list/DatabasesList.config.tsx
KrumTy: it is used for a boolean setting though - if the count is larger, then we enable pagination
the page size itself is not 15, it's 10-20 or whatever the user picks

redisinsight/ui/src/pages/home/components/databases-list/DatabasesList.tsx
KrumTy: I'm not sure how that helps, it's not going to be reused. We don't expose constants for every custom height or width of components do we?

redisinsight/ui/src/styles/elastic.css
KrumTy: we should ideally remove all of that, yes
let my comment commit stay as git blame for now

redisinsight/ui/src/pages/home/components/databases-list/components/BulkItemsActions/BulkItemsActions.tsx
KrumTy: the plus side of this is you know that this inline callback is not used anywhere else
whereas if you have 5-6 callback wrappers above the render, it is easy to call one in the other and get a complex mix as what was done in the previous db list version

I agree it makes sense when a callback is more complex than calling two functions and also if it needs memoization, which is not the case here (onExport is passed to a button)

redisinsight/ui/src/pages/home/components/databases-list/methods/handlers.ts
ArtemHoruzhenko, line 1: having separate handlers file for such a big component looks reasonable. well done! Need to do the same 👍

redisinsight/ui/src/pages/home/components/databases-list/DatabasesList.config.tsx
valkirilov: Oh, I see. It comes with some complex logic behind it. Disregard my comment then.

redisinsight/ui/src/pages/home/components/db-status/styles.module.scss
```diff
@@ -18,6 +18,7 @@
     text-align: center;
     color: var(--euiColorEmptyShade);
     font-size: 11px;
+    margin-left: -10px;
```
ArtemHoruzhenko, line 21: nit: this looks like a workaround

redisinsight/ui/src/pages/home/components/databases-list/DatabasesList.tsx
valkirilov: Yes, it's very very minor, my comment was driven by:
- the pure habit of not having magic numbers all over the codebase, as a convention 
- and also, if we name it properly, it can omit the need for the additional comment

redisinsight/ui/src/pages/home/components/db-status/styles.module.scss
KrumTy: it is, we want to display it to the side of the instance name but only if the db is new, without moving the name itself and also not affecting the whole column

redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellControls/methods/handlers.spec.ts
valkirilov: Lovely, it looks much better now!

redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellControls/DatabasesListCellControls.spec.tsx
valkirilov: I see there is still casting to any, but I don't want to dig into what's wrong with this component. At least now the invocation looks more natural, marking it as resolved.

redisinsight/ui/src/pages/home/components/databases-list/hooks/useDatabaseListData.spec.ts
valkirilov: Looks much better now, in my opinion. Thanks.

redisinsight/ui/src/pages/home/components/databases-list/components/BulkItemsActions/BulkItemsActions.tsx
pd-redis: I'm in favor of inference wherever possible

redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellHost/DatabasesListCellHost.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { RiTooltip } from 'uiSrc/components'
+import { Text } from 'uiSrc/components/base/text'
+import { IconButton } from 'uiSrc/components/base/forms/buttons'
+import { CopyIcon } from 'uiSrc/components/base/icons'
+
+import { handleCopyToClipboard } from './methods/handlers'
+import { IDatabaseListCell } from '../../DatabasesList.types'
+import { HostPortContainer } from './DatabasesListCellHost.styles'
+
+const DatabasesListCellHost: IDatabaseListCell = ({ row }) => {
```
pd-redis, line 12: TODO: we have pretty much the same cell in many tables, I think we should try to reuse it

PR #5147, opened 2025-11-05
RI-7687: rollback warning message for import form + adjust layout

ArtemHoruzhenko:
# What

- added back message before file picker
- changed width of modal
- now all content in modal takes full width
- user Banner component for warning to follow designs
- removed "subdued" color for error modal 
# Testing

|Before|After|
|-|-|
Screenshot 2025-11-06 at 10 09 23|Screenshot 2025-11-06 at 10 08 31
Screenshot 2025-11-06 at 10 09 36|Screenshot 2025-11-06 at 10 08 44

### DIFF
--- a/redisinsight/ui/src/components/import-file-modal/ImportFileModal.tsx
+++ b/redisinsight/ui/src/components/import-file-modal/ImportFileModal.tsx
@@ -66,87 +66,97 @@ const ImportFileModal = <T,>({
           </Modal.Content.Header.Title>
         </Modal.Content.Header.Compose>
 
-        <Modal.Content.Body.Compose width="61rem">
-          <Col gap="l">
-            {warning && <FlexItem>{warning}</FlexItem>}
-            <FlexItem>
-              {isShowForm && (
-                <>
-                  <RiFilePicker
-                    id="import-file-modal-filepicker"
-                    initialPromptText="Select or drag and drop a file"
-                    className={styles.fileDrop}
-                    isInvalid={isInvalid}
-                    onChange={onFileChange}
-                    display="large"
-                    accept={acceptedFileExtension}
-                    data-testid="import-file-modal-filepicker"
-                    aria-label="Select or drag and drop file"
-                  />
-                  {isInvalid && (
-                    <ColorText
-                      color="danger"
-                      className={styles.errorFileMsg}
-                      data-testid="input-file-error-msg"
-                    >
-                      {invalidMessage}
-                    </ColorText>
-                  )}
-                </>
-              )}
-              {loading && (
-                <div
-                  className={styles.loading}
-                  data-testid="file-loading-indicator"
-                >
-                  <Loader size="xl" />
-                  <Text color="subdued" style={{ marginTop: 12 }}>
-                    Uploading...
-                  </Text>
-                </div>
-              )}
-              {error && (
-                <div className={styles.result} data-testid="result-failed">
-                  <RiIcon type="ToastCancelIcon" size="xxl" color="danger500" />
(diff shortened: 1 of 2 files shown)


redisinsight/ui/src/components/import-file-modal/ImportFileModal.tsx
```diff
@@ -64,8 +66,9 @@ const ImportFileModal = <T,>({
           </Modal.Content.Header.Title>
         </Modal.Content.Header.Compose>
 
-        <Modal.Content.Body.Compose>
-          <Col align="center">
+        <Modal.Content.Body.Compose width="61rem">
```
dantovska: The whole structure inside the modal.content.body.compose is wrong .. there should not be a Col inside a compose component, we should have the <Modal.Content.Body> as a direct child and everything else inside it. Then, the width should be set to a styled component in a separate file, e.g. ImportFileModal.styles.ts and be something like:
```
export const StyledModalContentBody = styled(Modal.Content.Body)`
     width: 60rem # why 61? or can we use pixels? 
`
```

redisinsight/ui/src/components/import-file-modal/ImportFileModal.tsx
```diff
@@ -143,7 +144,7 @@ const ImportFileModal = <T,>({
               >
                 {submitBtnText || 'Import'}
               </PrimaryButton>
-            </>
+            </Modal.Content.Footer.Group>
           )}
           {data && <PrimaryButton onClick={onClose}>OK</PrimaryButton>}
```
dantovska: is there a reason why this button is not inside a group?

redisinsight/ui/src/components/upload-warning/UploadWarning.tsx
```diff
@@ -1,25 +1,19 @@
...
-      </FlexItem>
-      <FlexItem>
-        <Text className={styles.warningMessage}>
-          Use files only from trusted authors to avoid automatic execution of
-          malicious code.
-        </Text>
-      </FlexItem>
-    </Row>
-  </CallOut>
+  <UploadWarningBanner
```
dantovska, line 6: nice

redisinsight/ui/src/pages/rdi/pipeline-management/components/upload-modal/components/upload-dialog/UploadDialog.tsx
```diff
@@ -13,6 +13,10 @@ export interface Props {
   loading: boolean
 }
 
+const warningMessage =
+  'If a new pipeline is uploaded, existing pipeline configuration and transformation\n' +
```
dantovska: why is the \n needed here? visually it is not applied:
image

redisinsight/ui/src/components/import-file-modal/ImportFileModal.tsx
ArtemHoruzhenko: @dantovska Component support such property (I didn't use `styles` property). So I think this is original way for to do it.
Regarding 61rem: I've changed it ti 610px. I've took it from designs
Regarding `Col` - I've changed it but in general I don't think it is an issue. It might be an issue with Table but why here?

redisinsight/ui/src/components/import-file-modal/ImportFileModal.tsx
ArtemHoruzhenko: Putted it into groups but basically nothing is changed because it is not possible 3 buttons will be shown at the same time. But your suggestion is more safe. Thank you!

PR #5149, opened 2025-11-06
RI-7681: Apply new UI changes to DB Info section on Edit modal

dantovska:
# What

- Change Edit DB modal -> DB info section styling
- Change the styles for the Clone button 

# Testing

### Standalone

| Before | After |
| --- | --- |
| image | image |
| image | image |

### Cluster

| Before | After |
| --- | --- |
| image | image |
| image | image |

### Sentinel

| Before | After |
| --- | --- |
| image | image |
| image | image |

### Redis Stack

| Before | After |
| --- | --- |
| image | image |
| image | image |

### DIFF
--- a/redisinsight/ui/src/pages/home/components/form/DbInfo.tsx
+++ b/redisinsight/ui/src/pages/home/components/form/DbInfo.tsx
@@ -43,7 +43,7 @@ export const ListGroupItemLabelValue = ({
   <ListGroupItem
     label={
       <Row align="center" gap="m">
-        <Text color="ghost">{label}</Text>
+        <Text color="secondary">{label}</Text>
         <Text color="primary" data-testid={dataTestId}>
           {value}
         </Text>


valkirilov (approved):

redisinsight/ui/src/pages/home/components/form/sentinel/DbInfoSentinel.tsx
```diff
@@ -21,49 +21,63 @@
...
+  :hover {
+    opacity: 1;
+  }
+`
+
 const DbInfoSentinel = (props: Props) => {
   const { connectionType, nameFromProvider, sentinelMaster, host, port } = props
+
+  const handleCopy = (text = '') => {
+    navigator.clipboard.writeText(text)
```
pd-redis: We have similar functionality in so many places. We should extract it to utility function

redisinsight/ui/src/pages/home/components/form/DbInfo.styles.ts
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components'
+import { Group } from 'uiSrc/components/base/layout/list'
+
+export const DbInfoGroup = styled(Group)`
+  background-color: ${({ theme }) =>
+    theme.semantic.color.background.neutral200};
+  border: 1px solid ${({ theme }) => theme.semantic.color.border.neutral400};
```
pd-redis, line 7: I think this is starting to get out of hand, we use semantic.color.border in 26 places (at least), 21 are neutral500, one of 400, 600, 800, and one primary600.

This is more like a question to everybody - do you think it makes sense to define the colors we use in RI explicitly? Like borderColor, bgColor etc ?

KrumTy: I'm not sure keeping our version in sync with redis ui would be easy

redisinsight/ui/src/pages/home/components/form/DbInfo.tsx
```diff
@@ -60,104 +83,59 @@ const DbInfo = (props: Props) => {
...
-                color="default"
-                className={styles.dbInfoListValue}
-                data-testid="connection-type"
-              >
-                {capitalize(connectionType)}
-              </ColorText>
-            </Text>
-          }
+        <ListGroupItemLabelValue
+          label="Connection Type:"
```
KrumTy: nice, you can even create an array with label, value, dataTestId, isVisible for every field, like a config, and loop through it (map -> ListGroupItemLabelValue)

redisinsight/ui/src/pages/home/components/form/sentinel/DbInfoSentinel.tsx
```diff
@@ -21,49 +21,63 @@ export interface Props {
   sentinelMaster?: SentinelMaster
 }
 
+const StyledCopyButton = styled(IconButton)`
+  margin-bottom: 3px;
+  opacity: 0;
+
+  :hover {
```
KrumTy: is this only if you hover explicitly over the hidden button?
also, worth adding all styled components to XYZ.styles.ts file

redisinsight/ui/src/pages/home/components/form/sentinel/DbInfoSentinel.tsx
dantovska: @valkirilov noted me of that, aren't you handling this already? If so, please share so I can follow what you are doing?

redisinsight/ui/src/pages/home/components/form/sentinel/DbInfoSentinel.tsx
dantovska: yes, this is how it currently works and there weren't any designs to behave otherwise ... maybe I should bring it up. thanks !

redisinsight/ui/src/pages/home/components/form/DbInfo.styles.ts
valkirilov: Yeah, for sure, it might come with some struggles (keeping them in sync), but also, I see Redis UI introduces major breaking changes, so we're also kind of blocked now (we can't get the latest bug fixes and features because of that). 

And I think that soon we'll end up in a state where we don't treat it as a main source of truth, but instead build an intermediate abstract layer that depends on it. Just like we did with our components (`RiIcon`, `RiTabs`, etc.), we might need something similar for the color palette/theme. 

For sure, it's not something straightforward (maybe), but we should definitely consider it and think more about it, and how we can handle it in an efficient way.

dantovska: @pd-redis I don't know if it will be expected to have the same border everywhere? And we will need an effort to change that. I don't just want to introduce the variable, without changing everywhere and then we will end up like the eui colors that we are trying to get rid off ... borderLight, borderLighter ... It does makes sense if we do a redesign again and want to change border everywhere...  Anyway, I don't think it should start in this PR

PR #5155, opened 2025-11-07
RI-7687: Adjust RDI empty screen

ArtemHoruzhenko:
# What

Changed text, action button, positioning for empty RDI screen

# Testing

|Before|After|
|-|-|
Screenshot 2025-11-07 at 10 40 38|Screenshot 2025-11-07 at 10 40 20
Screenshot 2025-11-07 at 10 40 53|Screenshot 2025-11-07 at 10 40 07

### DIFF
--- a/redisinsight/ui/src/pages/rdi/home/empty-message/styles.ts
+++ b/redisinsight/ui/src/pages/rdi/home/empty-message/styles.ts
@@ -1,12 +1,11 @@
 import styled from 'styled-components'
 import { Theme } from 'uiSrc/components/base/theme/types'
+import { FlexItem } from 'uiSrc/components/base/layout/flex'
 
-export const EmptyPageContainer = styled.div`
-  display: flex;
+export const EmptyPageContainer = styled(FlexItem)`
   padding: ${({ theme }: { theme: Theme }) => theme.core.space.space300};
   border: 1px solid
     ${({ theme }: { theme: Theme }) => theme.semantic.color.border.neutral500};
   border-radius: ${({ theme }: { theme: Theme }) =>
     theme.components.card.borderRadius};
-  height: 100%;
 `
--- a/redisinsight/ui/src/pages/rdi/home/header/RdiHeader.tsx
+++ b/redisinsight/ui/src/pages/rdi/home/header/RdiHeader.tsx
@@ -14,27 +14,27 @@ export interface Props {
 const RdiHeader = ({ onRdiInstanceClick }: Props) => {
   const { data: instances } = useSelector(instancesSelector)
 
+  if (instances.length === 0) {
+    return null
+  }
+
   return (
-    <>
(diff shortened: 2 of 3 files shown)


redisinsight/ui/src/pages/rdi/home/header/RdiHeader.tsx
```diff
@@ -15,24 +15,26 @@ const RdiHeader = ({ onRdiInstanceClick }: Props) => {
...
+              </PrimaryButton>
+            </FlexItem>
+            <FlexItem className="searchContainer">
+              <SearchRdiList />
+            </FlexItem>
+          </Row>
+          <Spacer className="spacerDl" />
+        </div>
+      )}
+    </>
```
pawelangelow: Why not
```
if (instances.length === 0) {
return null;
}

return (
<div className="containerDl">
....
)
```

redisinsight/ui/src/pages/rdi/home/empty-message/styles.ts
```diff
@@ -0,0 +1,12 @@
+import styled from 'styled-components'
+import { Theme } from 'uiSrc/components/base/theme/types'
+
+export const EmptyPageContainer = styled.div`
```
valkirilov: Tip: Instead of a div, we can use `FlexGroup`, `Row` or `Col`, so you don't have to set `display: flex` after that.

```
export const EmptyPageContainer = styled(FlexGroup)`
    lotem: "ipsum";
`
```

redisinsight/ui/src/pages/rdi/home/empty-message/styles.ts
```diff
@@ -0,0 +1,12 @@
...
+  display: flex;
+  padding: ${({ theme }: { theme: Theme }) => theme.core.space.space300};
+  border: 1px solid
+    ${({ theme }: { theme: Theme }) => theme.semantic.color.border.neutral500};
+  border-radius: ${({ theme }: { theme: Theme }) =>
+    theme.components.card.borderRadius};
+  height: 100%;
+`
```
pawelangelow:
```
export const EmptyPageContainer = styled(FlexItem)`
padding: ${({ theme }: { theme: Theme }) => theme.core.space.space300};
  border: 1px solid
    ${({ theme }: { theme: Theme }) => theme.semantic.color.border.neutral500};
  border-radius: ${({ theme }: { theme: Theme }) =>
    theme.components.card.borderRadius};
```

flex item is by default `flex` and it has `grow` set to true (as I remember), so you won't need `height: 100%`. Practically the same, but sometimes `height: 100%` can cause you a bug if the parent container isn't positioned as well, meaning the child would get all it's height and can overflow somewhere.

pawelangelow:
And please fix the title of the PR - something like `RI-7687: Adjust RDI empty screen`

PR #5172, opened 2025-11-11
RI-7687: fix double scroll issue on the RDI statistics page

ArtemHoruzhenko:
# What

fixed issue with 2 scrolls appeared for statistics and main container at the same time
# Testing

Before
https://github.com/user-attachments/assets/d5780cf8-1190-4666-90d4-5e20d06c176b

After
https://github.com/user-attachments/assets/3923c01c-419a-473d-8d9c-ca6416ca3d9f

### DIFF
--- a/redisinsight/ui/src/templates/rdi-instance-page-template/styles.ts
+++ b/redisinsight/ui/src/templates/rdi-instance-page-template/styles.ts
@@ -2,5 +2,5 @@ import styled from 'styled-components'
 import { FlexItem } from 'uiSrc/components/base/layout/flex'
 
 export const RdiInstancePageContentContainer = styled(FlexItem)`
-  height: calc(100% - 230px);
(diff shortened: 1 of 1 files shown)


redisinsight/ui/src/templates/rdi-instance-page-template/styles.ts
```diff
@@ -0,0 +1,6 @@
+import styled from 'styled-components'
+import { FlexItem } from 'uiSrc/components/base/layout/flex'
+
+export const RdiInstancePageContentContainer = styled(FlexItem)`
+  height: calc(100% - 210px);
```
valkirilov, line 5: Tip: You can leave a note in a comment on what exactly 210 is (I guess it's the header + padding + something else), so we know how to calculate it in case we need to update it in the future.

PR #5175, opened 2025-11-11
RI-7681: Make info section fixed + apply scrolling properly

dantovska:
# What

### When there is not enough space for the content in the form modals, scroll should be applied.
For the edit forms, the scroll should not apply to the info section + tabs - it should remain fixed. It applies only on the form on each tab.

# Testing

Before:

https://github.com/user-attachments/assets/78815ab2-6f24-4d80-9fa1-59ffa8343519

After:

https://github.com/user-attachments/assets/46293401-ad5a-43a7-ac15-d55a9afd4685

### DIFF
--- a/redisinsight/ui/src/pages/home/components/database-panel-dialog/DatabasePanelDialog.tsx
+++ b/redisinsight/ui/src/pages/home/components/database-panel-dialog/DatabasePanelDialog.tsx
@@ -1,8 +1,5 @@
 import React, { useEffect, useState } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
-import styled from 'styled-components'
-import { Theme } from '@redis-ui/styles'
-
 
 import { Nullable } from 'uiSrc/utils'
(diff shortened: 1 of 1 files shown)


redisinsight/ui/src/pages/home/components/styles.module.ts
```diff
@@ -0,0 +1,26 @@
+import styled from 'styled-components'
+
+export const FixedWrapper = styled.div`
```
valkirilov, line 3: Tip: You can use `FlexGroup` or `Col` instead of a regular `div`, so you don't have to define the flex properties on your own.

PS: The same note applies for the other classes bellow, as well.

KrumTy (approved):

redisinsight/ui/src/pages/home/components/styles.module.ts
dantovska: Thanks! Using FlexGroup or Col add a lot of different styles with them, which I try to avoid and make it simple due to the sensitive scenario with the scrolling, that is why I decided to go with this approach.

PR #5176, opened 2025-11-12
[no-issue]: optimize FeatureFlagComponent

KrumTy:
# What

I noticed FeatureFlagComponent is overcomplicated. There's no need of cloning, arrays, mapping, function calls.

# Testing

Components wrapped in FeatureFlagComponent show up correctly

### DIFF
--- a/redisinsight/ui/src/components/feature-flag-component/FeatureFlagComponent.tsx
+++ b/redisinsight/ui/src/components/feature-flag-component/FeatureFlagComponent.tsx
@@ -29,13 +29,7 @@ const FeatureFlagComponent = (props: Props) => {
     return null
   }
 
-  const cloneElement = (child: React.ReactElement) => React.cloneElement(child)
-
-  return isArray(children) ? (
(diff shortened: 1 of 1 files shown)


redisinsight/ui/src/components/feature-flag-component/FeatureFlagComponent.tsx
```diff
@@ -29,13 +29,7 @@ const FeatureFlagComponent = (props: Props) => {
...
   }
 
-  const cloneElement = (child: React.ReactElement) => React.cloneElement(child)
-
-  return isArray(children) ? (
-    <>{React.Children.map(children, cloneElement)}</>
-  ) : (
-    cloneElement(children)
-  )
+  return <>{children}</>
```
pawelangelow, line 32:
```suggestion
  return children
```

KrumTy: The fragment is what makes the function’s return type a JSX.Element in a type-safe way. If we drop it, we need to assert types it seems.

PR #5181, opened 2025-11-12
Ri-7465 auto discovery layout

pd-redis:
# What

This PR refactors the Redis Cloud and Redis Cluster database discovery UI by extracting inline table column definitions
into dedicated, reusable components. These changes bring the current implementation closer to Figma designs while
improving code maintainability, testability, and consistency across the application.

### 1. **Column Definition Extraction**

- For **Redis Cluster**, **Redis Cloud Subscriptions**, **Redis Cloud Databases** and **Redis Cloud Databases Result** extracted column definitions into dedicated, reusable functions.

### 2. **Custom Hook Extraction**

- Created hooks for Redis Cluster and Redis Cloud database discovery logic.

### 3. **Storybook Stories**

- Added comprehensive stories for `RedisClusterDatabases` (15+ variants)
- Added stories for `RedisCloudDatabases` with different states
- Added stories for `RedisCloudSubscriptions` with multipage examples
- Added stories for `RedisCloudDatabasesResult` with success/error states
- Created factory functions for mock data generation

### 4. **UI/UX Improvements (Aligned with Figma Designs)**

- Improved `DatabaseListOptions` component with better styling and stories to match design specs
- Enhanced `MessageBar` component with variant support and comprehensive tests
- Fixed `HoverContent` to not wrap non-text titles per design guidelines
- Updated `ClusterConnectionForm` layout to align with Figma mockups
- Improved endpoint display with consistent copy functionality matching design system
- Updated column sizing and spacing to match Figma specifications

## Tech Decisions:

1. **Column Component Pattern**: Each column is now a factory function returning a `ColumnDef`
2. **Shared Components**: Leveraged existing `CopyBtn`, `CopyTextContainer`, and `CellText` components for consistency
3. **Factory Pattern**: Created factory functions for mock data to support comprehensive Storybook stories
4. **Hook Extraction**: Moved complex table logic to custom hooks, keeping page components focused on rendering

# Testing

## Manual Testing:

- Verified Redis Cluster database discovery flow works correctly
  - **Note**: To test Redis Cluster, a Redis Enterprise installation is needed. One can be found in the `tests/e2e/rte/redis-enterprise` folder
- Verified Redis Cloud subscription selection and database discovery
- Verified result pages display correctly with success/error states
- Tested column sorting, selection, and copy functionality
- Verified conditional column visibility (Capabilities, Options)
- Tested all Storybook stories render correctly

## Automation Tests:

- Added comprehensive `MessageBar` component tests covering all variants and states
[...]

PR #5182, opened 2025-11-13
RI-7714 Replace the Pub/Sub table with Redis UI

valkirilov:
# What

Replace the table used on the Pub/Sub page with Redis UI
- also, delete no longer necessary code related to old table
- added a default pagination, so we don't limit the table heigh with a fixed number

# Testing

1. Open Redis Insight and go to the **Databases** tab
2. Open an existing database, or establish a new connection to a new database
3. Go to the **Pub/Sub** page

### Empty State

4. Subscribe to receiving pub/sub messages

You should be able to see the empty state of the table, letting you know there are no messages yet

| Before | After |
| - | - |
Screenshot 2025-11-13 at 13 21 35|Screenshot 2025-11-13 at 13 17 17
Screenshot 2025-11-13 at 13 21 41|Screenshot 2025-11-13 at 13 17 27

### Listen for messages

5. Fill in the "**Channel**" and the "**Message**" fields at the bottom of the screen
6. Click on the "**Publish**" button

You should see how the table starts to grow with your messages

| Before | After |
| - | - |
Screenshot 2025-11-13 at 13 20 57|Screenshot 2025-11-14 at 16 42 06
Screenshot 2025-11-13 at 13 21 01|Screenshot 2025-11-14 at 16 42 18

### DIFF
--- a/redisinsight/ui/src/pages/pub-sub/components/messages-list/MessagesListWrapper.tsx
+++ b/redisinsight/ui/src/pages/pub-sub/components/messages-list/MessagesListWrapper.tsx
@@ -18,6 +18,7 @@ import PatternsInfo from '../patternsInfo'
 import { Wrapper } from './MessageListWrapper.styles'
 import { Table } from 'uiSrc/components/base/layout/table'
 import { PUB_SUB_TABLE_COLUMNS } from './MessagesListTable/MessagesListTable.config'
+import { PubSubTableColumn } from './MessagesListTable/MessagesListTable.constants'
 
 const MessagesListWrapper = () => {
   const {
@@ -76,8 +77,10 @@ const MessagesListWrapper = () => {
           <Table
             columns={PUB_SUB_TABLE_COLUMNS}
             data={messages}
-            maxHeight="50vh"
             stripedRows
+            enableSorting
+            paginationEnabled
+            defaultSorting={[{ id: PubSubTableColumn.Timestamp, desc: true }]}
             emptyState="No messages published yet"
           />
         </div>


redisinsight/ui/src/pages/pub-sub/components/messages-list/MessagesListWrapper.tsx
```diff
@@ -9,15 +8,27 @@ import { CommandsVersions } from 'uiSrc/constants/commandsVersions'
...
+  ChannelColumn,
+  MessageColumn,
+  TimestampColumn,
+} from './column-definitions'
+
+const columns: ColumnDef<IMessage>[] = [
+  TimestampColumn(),
+  ChannelColumn(),
+  MessageColumn(),
+]
```
KrumTy: why have functions that return static objects for this? `TimestampColumn` does not sound like a function name, more like a component.
For Tables I've used this pattern: https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/home/components/databases-list/DatabasesList.config.tsx#L44 file that exposes columns and other configs as static objects 

It's easier to manage because the col defs are not split across multiple files (cell render can be extracted since that's actually a component). Also, should the need of parameters arise (hopefully not), we can move from an array to a function that generates the array

valkirilov: I saw that both you and @pd-redis moved the column definitions to separate files, and I liked that. [[example](https://github.com/redis/RedisInsight/pull/5181/files#diff-b380c81f17297e105d2e8d599c76bd512ff0d66a543eec848f4ac2e23e6b5ae7R13)] [[example](https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/home/components/databases-list/components/DatabasesListCellHost/DatabasesListCellHost.tsx#L12)]

So, I went for the same approach to keep it consistent with your code. I have used the pattern inspired by [this](https://github.com/redis/RedisInsight/pull/5181/files#diff-be55448f298ce5067b7188ad5525aaa5af45fd08f4c7db29313a88121a6dbd82R51).

We can align on the approach offline, together.

KrumTy: I have two large table PRs approved and merged with the pattern shown above. It doesn't make sense sense aligning with unmerged PRs imo. 
Anyway, I have explained my reasoning there as well.

I'd be very happy if you point a flaw in my implementation and prove me wrong btw. And I agree with you that we should align with one way of doing it. That's why I'm writing this in the first place :)

KrumTy: Just so that I am more clear about the idea:

1. Define all columns as a static array ([example](https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/rdi/home/components/rdi-instances-list/RdiInstancesList.config.tsx#L13))
2. You can later [filter](https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/rdi/home/components/rdi-instances-list/hooks/useRdiInstancesListData.ts#L34) this columns array, if you want to show only some of the columns
3. Cells - declare those as custom/reusable components that implement the same [interface](https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/rdi/home/components/rdi-instances-list/RdiInstancesList.types.ts#L6). They can have styles/tests/etc. They can access state/context and so on. ([example](https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/rdi/home/components/rdi-instances-list/components/RdiInstancesListCell/RdiInstancesListCell.tsx))

This is a very robust solution with a clear separation of concerns. Generating columns run time and mixing them with components is not.

pawelangelow: I'm with @KrumTy for the column definitions. Since the PR he showed is merged, this is the link to the example: https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/home/components/databases-list/DatabasesList.config.tsx

valkirilov: Applied the suggested changes 🙌

PR #5186, opened 2025-11-13
RI-7713: polish import from file forms

dantovska:
# What

#### Import file form:
- attaching 10MB+ file error message 
- adjust spacing and text color

#### Invalid file content

- Replace icon, make it bigger - e.g. import json file with only empty array in it [] - 
 
####  Import result page 
- fix table borders, 
- Remove colors for success, partial success, failed
- Make only one section expanded at a time as in the old ui
- Add padding around the section so box-shadow is not cut off

# Testing

| Before | After |
| --- | --- |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

### DIFF
--- a/redisinsight/ui/src/pages/home/components/import-database/components/ResultsLog/ResultsLog.spec.tsx
+++ b/redisinsight/ui/src/pages/home/components/import-database/components/ResultsLog/ResultsLog.spec.tsx
@@ -27,7 +27,7 @@ describe('ResultsLog', () => {
 
     expect(screen.getByTestId('success-results-closed')).toBeInTheDocument()
     expect(screen.getByTestId('partial-results-closed')).toBeInTheDocument()
-    expect(screen.getByTestId('failed-results-closed')).toBeInTheDocument()
+    expect(screen.getByTestId('fail-results-closed')).toBeInTheDocument()
   })
 
   it('should open and collapse other groups', () => {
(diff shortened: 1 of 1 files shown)


redisinsight/ui/src/pages/home/components/import-database/components/ResultsLog/ResultsLog.tsx
```diff
@@ -45,79 +65,34 @@ const ResultsLog = ({ data }: Props) => {
...
+    <Row gap="s">
+      <Text data-testid="nav-group-title">{title}:</Text>
+      <Text data-testid="number-of-dbs">{length}</Text>
+    </Row>
   )
 
-  const getNavGroupState = (name: ResultsStatus) =>
-    openedNav === name ? 'open' : 'closed'
+  const getNavGroupState = (type: ImportDatabaseResultType) =>
+    openedNav === type ? 'open' : 'closed'
```
KrumTy: no point of declaring a function and calling it 3 times with the same argument

```suggestion
  const navState = openedNav === type ? 'open' : 'closed'
```

redisinsight/ui/src/pages/home/components/import-database/components/ResultsLog/ResultsLog.tsx
```diff
@@ -1,42 +1,62 @@
...
+      data: data?.partial ?? [],
+    },
+    {
+      type: ImportDatabaseResultType.Failed,
+      title: 'Failed to import',
+      data: data?.fail ?? [],
+    },
+  ]
 
   const CollapsibleNavTitle = ({
```
KrumTy: this component doesn't belong here. I'd delete it, steal its return jsx (3 lines) and put it where its called

redisinsight/ui/src/pages/home/components/import-database/components/ResultsLog/ResultsLog.tsx
```diff
@@ -1,42 +1,62 @@
...
     }
-  }
+  }, [openedNav])
+
+  const collapsibleNavData: TableResultData[] = [
+    {
+      type: ImportDatabaseResultType.Success,
+      title: 'Fully imported',
+      data: data?.success ?? [],
+    },
```
KrumTy: You can move that config out of the component. Instead of `data` key you have a `type` key that you can use for filtering/rendering required bellow.

I'd leave only the map part bellow though, move the rest under useMemo so the conditions can be visible

KrumTy (approved):

PR #5189, opened 2025-11-14
RI-7707 update the database analysis page

pd-redis:
# What

This PR migrates the database analysis page components from SCSS modules to styled-components, adds comprehensive Storybook stories, and resolves circular dependency issues.

### 1. Styled-components Migration
- **Charts**: Migrated `BarChart` and `DonutChart` from SCSS to styled-components with theme integration
- **Database Analysis Components**: Converted `TopNamespace`, `TopKeys`, `TableLoader`, `SummaryPerData`, `ExpirationGroupsView`, `GroupBadge`, and `TableTextBtn` to styled-components
- **Theme Integration**: Replaced CSS custom properties with theme values (`theme.semantic.color.*`, `theme.core.space.*`)

### 2. Storybook Stories
Added comprehensive interactive stories for:
- `TopNamespace` - 11 stories covering loading, empty states, data views, extrapolation, and custom delimiters
- `TopKeys` - 11 stories covering loading, various key types, TTL values, big keys, and custom delimiters  
- `GroupBadge` - 8 stories covering key types, command groups, compressed mode, delete functionality
- `ExpirationGroupsView` - Stories for bar chart TTL visualization
- `DatabaseAnalysisPageView` - Full page component stories
- `BarChart` and `DonutChart` - Enhanced existing stories

### 3. Styled-components Infrastructure
- Created global styles (`globalStyles.ts`) with CSS variables for Redis data types and command groups
- Added reusable mixins in `styles/mixins/styledComponents.ts`:
  - `truncateText` - Text overflow handling
  - `breakpoint` - Responsive media queries
  - `scrollbarStyles` - Custom scrollbar styling
  - `insightsOpen` - Insights panel responsive states
- Made `CellText` component props overridable while maintaining defaults

### 4. TypeScript & Testing Improvements
- Fixed type errors in `TopNamespace.spec.tsx` by creating `createMockedDatabaseAnalysis` helper
- Fixed type errors in `TopKeysTable.spec.tsx` by properly typing Key arrays
- Updated component props with proper TypeScript interfaces
- Renamed components for clarity: `Table` → `TopKeysTable` and `TopNamespacesTable`

# Testing

## Manual Testing
- Open a database with some content, go to Analysis tab, click on New Report button

## Visual Verification

https://github.com/user-attachments/assets/e7b25754-37d1-42bc-9264-3d7adf95032b


redisinsight/ui/src/styles/globalStyles.ts
```diff
@@ -0,0 +1,42 @@
...
+    --typeHashColor: #364cff;
+    --typeListColor: #008556;
+    --typeSetColor: #9c5c2b;
+    --typeZSetColor: #a00a6b;
+    --typeStringColor: #6a1dc3;
+    --typeReJSONColor: #3f4b5f;
+    --typeStreamColor: #6a741b;
+    --typeGraphColor: #14708d;
+    --typeTimeSeriesColor: #6e6e6e;
+
```
KrumTy, line 21: new patterns, affecting the whole app, should be ideally introduced in a separate PR with explanation.

Can't we use @ArtemHoruzhenko 's suggestion about extending theme with custom colors instead?

redisinsight/ui/src/pages/database-analysis/components/top-namespace/TopNamespace.spec.tsx
```diff
@@ -13,11 +13,33 @@ import {
...
+const createMockedDatabaseAnalysis = (
+  overrides: Partial<DatabaseAnalysis> = {},
+): DatabaseAnalysis =>
+  ({
+    id: 'analysis-id',
+    databaseId: 'database-id',
+    filter: { match: '*', count: 10000 },
+    delimiter: ':',
+    progress: { total: 100, scanned: 100 },
+    createdAt: new Date('2023-01-01'),
```
KrumTy: looks like factory can be used here

redisinsight/ui/src/pages/database-analysis/components/top-keys/TopKeysTable.tsx
```diff
@@ -87,7 +87,7 @@ const TopKeysTable = ({
     history.push(Pages.browser(instanceId))
   }
 
-  const columns: ColumnDefinition<Key>[] = [
+  const columns: ColumnDef<Key>[] = [
```
KrumTy, line 90: it is easy to extract columns outside the component - just need to put delimiter in a context or access it by other means

redisinsight/ui/src/components/base/forms/buttons/EmptyButton.tsx
```diff
@@ -22,22 +22,26 @@ export const EmptyButton = ({
...
+        <ButtonIcon
+          buttonSide="right"
+          icon={icon}
+          iconSide={iconSide}
+          loading={loading}
+          size={size}
+        />
+      </Row>
+    ) : (
+      children
```
KrumTy, line 44: why is this change needed? Can't you conditionally render just the buttons instead? Also why are both the same?

redisinsight/ui/src/pages/database-analysis/DatabaseAnalysisPageView.tsx
```diff
@@ -0,0 +1,52 @@
...
+import DatabaseAnalysisTabs from './components/data-nav-tabs'
+import {
+  type DatabaseAnalysis,
+  type ShortDatabaseAnalysis,
+} from 'apiSrc/modules/database-analysis/models'
+import { Nullable } from 'uiSrc/utils'
+import { Col } from 'uiSrc/components/base/layout/flex'
+import { Theme } from 'uiSrc/components/base/theme/types'
+
+import './styles.module.scss'
```
KrumTy: Can you explain why these global styles are needed here?

redisinsight/ui/src/pages/database-analysis/components/header/styles.module.scss
```diff
@@ -26,3 +26,7 @@
   color: var(--euiColorMediumShade);
   cursor: pointer;
 }
+
+.tooltipAnchor {
+  display: inline-flex;
+}
```
KrumTy, line 32: was this style reverted?

redisinsight/ui/src/styles/globalStyles.ts
pd-redis: Ideally these should not be custom, but I have not gotten a list of theme colors to use.

My suggestion for using css properties instead of even theme overrides is simply practical. 

example:
```typescript
--bg-color-neutral100: ${({ theme }) => theme.semantic.color.background.neutral100};
    // ....
background-color: ${({ theme }) => theme.semantic.color.background.neutral100};
    // vs
background-color: var(--bg-color-neutral100);
```

If --var is defined at root level, it is accessible everywhere and is a lot shorter to type.

In this case, the colors defined are used in most if not all "key" related components.
In case of theme changes, this can be the single place where we need to apply the changes.
If we use the approach to override the theme, we should not override it but define our theme based on redis-ui theme. Why? Because in case that redis-ui decides to change the structure from `theme.semantic.color.background.neutral100` to just `theme.components.page.background`, it will not be just matter of changing the override, but also everywhere the old path is used.

Of course this is just an opinion, the team can decide if we want to use any sort of override at all.

redisinsight/ui/src/pages/database-analysis/components/top-namespace/TopNamespace.spec.tsx
pd-redis: done

redisinsight/ui/src/components/base/forms/buttons/EmptyButton.tsx
pd-redis: Its ugly, but allows not to have extra wrapper if you don't pass an icon.
I don't like it as well, maybe we should use different approach.
Right now I need it in order to have "real" text like look of the button

redisinsight/ui/src/pages/database-analysis/components/header/styles.module.scss
pd-redis: Not sure what you mean by reverted, but I needed this style, because without it, the tooltip wrapped content cannot be aligned: center (or otherwise) properly.
In icon that is 20x20 px when wrapped in tooltip, effectively has height of 25 or something like that - the tooltip content container has that size, not the icon, so if it is centerd, it is actually offset by some pixels

redisinsight/ui/src/pages/database-analysis/DatabaseAnalysisPageView.tsx
pd-redis: wasn't removed when cleaning up scss classes

redisinsight/ui/src/pages/database-analysis/components/top-keys/TopKeysTable.tsx
pd-redis: Yes, but lets leave it as is for now, the focus of this PR is visuals

redisinsight/ui/src/components/base/forms/buttons/EmptyButton.tsx
KrumTy: does it make sense that both left and right icons are the same though?

pd-redis: The idea is (and was since this component exists) to be able to put icon on either side of the button. That is what `iconSide` prop does

pd-redis: The change here is not related to how icons are hadled when they are present, it is about what to do when there is no icon

pd-redis: https://github.com/redis/RedisInsight/blob/9d31cba76cfc3f66f21cd15298ff6120c314ac13/redisinsight/ui/src/components/base/forms/buttons/EmptyButton.tsx#L25

redisinsight/ui/src/styles/globalStyles.ts
ArtemHoruzhenko: @pd-redis I would prefer to use `theme` here so it will be aligned with everything else

ArtemHoruzhenko: For me it looks like we must extend existing theme with another semantic component (e.g. keyType.color = 'json' | 'hash' | ...)

pd-redis: I agree with you @ArtemHoruzhenko on both comments. But the colors in the designs are from the update we're supposed to do next. This will for sure be migrated to the theme once we're on the same page as designers
cc. @KrumTy  cc @DimoHG

redisinsight/ui/src/pages/database-analysis/components/base/TableTextBtn.tsx
```diff
@@ -1,43 +1,21 @@
 import styled, { css } from 'styled-components'
 import React from 'react'
 import { EmptyButton } from 'uiSrc/components/base/forms/buttons'
+import { Theme } from 'uiSrc/components/base/theme/types'
 
-const expandedStyle = css`
+const expandedStyle = css<{ theme: Theme }>`
```
KrumTy, line 6: can move all of this to TableTextBtn.styles.ts, and leave this a simple export

KrumTy (approved):
overall LGTM

PR #5197, opened 2025-11-17
Don't break the UI when the screen is smaller

valkirilov:
# What

Make sure we don't break the whole layout when the screen is smaller.

| Before | After |
| - | - |
image|Screenshot 2025-11-17 at 14 41 40


KrumTy (approved):
nice 💪

PR #5203, opened 2025-11-18
RI-7463 sentinel bugfixes

pd-redis:
# What
When the “Add Primary Group” is disabled, users see a tooltip hovering over any place in the button, not only the exclamation mark.
image

 Page titles and column names have uppercase for the first word, and lowercase for the following ones.
 “Here is a list of primary groups your Sentinel instance is managing.” is located on the first row. The text is 14px not 12px 
image

The database index misses an explanation. Since redis-ui does not support OOB composed tooltip, it is moved to the header by request of PM. Width of input field is also limited by request
image

# Testing

Add Sentinel cluster database, the changes applied here, can be seen in the "Auto-Discover Redis Sentinel Primary Groups" page

### DIFF
--- a/redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases/components/column-definitions/DbIndexColumn.tsx
+++ b/redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases/components/column-definitions/DbIndexColumn.tsx
@@ -5,7 +5,7 @@ import { SentinelInputFieldType } from 'uiSrc/components/input-field-sentinel/In
 import { type ColumnDef } from 'uiSrc/components/base/layout/table'
 import { ModifiedSentinelMaster } from 'uiSrc/slices/interfaces'
 import { RiIcon } from 'uiSrc/components/base/icons'
-import { ColumnDefinitionTitles } from './constants'
+import { ColumnDefinitionTitles } from 'uiSrc/pages/autodiscover-sentinel/constants/constants'
 import { Row } from 'uiSrc/components/base/layout/flex'
 
 const InputContainer = styled.div<React.HTMLAttributes<HTMLDivElement>>`
--- a/redisinsight/ui/src/pages/autodiscover-sentinel/constants/constants.ts
+++ b/redisinsight/ui/src/pages/autodiscover-sentinel/constants/constants.ts
@@ -6,4 +6,5 @@ export enum ColumnDefinitionTitles {
   NumberOfReplicas = '# of replicas',
   Password = 'Password',
   PrimaryGroup = 'Primary group',
+  Result = 'Result',
 }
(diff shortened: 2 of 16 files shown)


redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases/components/SentinelDatabases/SentinelDatabases.tsx
```diff
@@ -161,23 +161,26 @@ const SentinelDatabases = ({
...
-      <PrimaryButton
-        type="submit"
-        onClick={onClick}
-        disabled={isSubmitDisabled()}
-        loading={loading}
-        icon={isSubmitDisabled() ? TooltipIcon : undefined}
-        data-testid="btn-add-primary-group"
+      <RiTooltip
+        position="top"
+        title={title}
```
pawelangelow, line 169: Q: Is the tooltip displayed when `content` is `undefined`? If so, maybe we want to have `disabled ? title : undefined` for the title?

redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases/components/column-definitions/DbIndexColumn.tsx
```diff
@@ -1,42 +1,54 @@
...
+          <RiTooltip
+            anchorClassName="inputAppendIcon"
+            position="left"
+            content="Select the Redis logical database to work with in Browser and Workbench."
+          >
+            <RiIcon type="InfoIcon" style={{ cursor: 'pointer' }} />
+          </RiTooltip>
+        </Row>
+      )
+    },
```
pawelangelow, line 33: nit: you can return it directly
```
header: () => (<Row... )
```
instead of
```
header: () => { return (<Row ...) }
```

pd-redis: Probably, but I prefer explicit returns. Do you insist on this?

redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases/components/SentinelDatabases/SentinelDatabases.tsx
pd-redis: No, it is not, the tool tip is `disabled` when there is no need to be displayed.

pd-redis: And it is pretty easy to test in storybook or in actual import

redisinsight/ui/src/pages/autodiscover-sentinel/sentinel-databases/components/column-definitions/DbIndexColumn.tsx
pawelangelow: We don't have a style guide on that yet, so - no.

pawelangelow (approved):

PR #5204, opened 2025-11-18
RI-7731: Adjust RDI: General info - update text

dantovska:
# What

##### General info:
- Removed unneeded FlexItem component as there is a Row inside it and it was breaking responsiveness
- Added some unit tests for the component as there were none 
- Fix text to follow the text in the figma designs

# Testing

#### General info, it is located on the Analytics tab

| Before | After |
| --- | --- |
| image | image |
| image | image |
| image | image |

### DIFF
--- a/redisinsight/ui/src/mocks/factories/rdi/RdiPipelineStatus.factory.ts
+++ b/redisinsight/ui/src/mocks/factories/rdi/RdiPipelineStatus.factory.ts
@@ -1,0 +1,29 @@
+import { Factory } from 'fishery'
+import { faker } from '@faker-js/faker'
+
+import {
+  IRdiPipelineStatus,
+  PipelineState,
+  PipelineStatus,
+} from 'uiSrc/slices/interfaces'
+
+export const rdiPipelineStatusFactory = Factory.define<IRdiPipelineStatus>(
+  () => ({
+    rdiVersion: faker.system.semver(),
+    address: `${faker.internet.ip()}:${faker.internet.port()}`,
+    runStatus: faker.helpers.arrayElement([
+      PipelineStatus.Validating,
+      PipelineStatus.Starting,
+      PipelineStatus.Stopping,
+      PipelineStatus.Resetting,
+      PipelineStatus.Ready,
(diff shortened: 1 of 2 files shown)


redisinsight/ui/src/pages/rdi/statistics/status/Status.spec.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react'
+
+import { render, screen } from 'uiSrc/utils/test-utils'
+import { IRdiPipelineStatus } from 'uiSrc/slices/interfaces'
+import Status from './Status'
+
+const mockedProps: IRdiPipelineStatus = {
```
valkirilov: Do you think there is a chance this mock to be used somewhere else, so we can create a factory for it? Or it's the only place we use it, and we can keep it as it is.

valkirilov (dismissed):

redisinsight/ui/src/pages/rdi/statistics/status/Status.spec.tsx
dantovska: Yes, it can be used in other places - in the parent component for example. I'll create a factory. Tnx!

redisinsight/ui/src/mocks/factories/rdi/RdiPipelineStatus.factory.ts
```diff
@@ -0,0 +1,29 @@
...
+  IRdiPipelineStatus,
+  PipelineState,
+  PipelineStatus,
+} from 'uiSrc/slices/interfaces'
+
+export const rdiPipelineStatusFactory = Factory.define<IRdiPipelineStatus>(
+  () => ({
+    rdiVersion: faker.system.semver(),
+    address: `${faker.internet.ip()}:${faker.internet.port()}`,
+    runStatus: faker.helpers.arrayElement([
```
valkirilov, line 14: Tip: If you don't want to explicitly list particular enum values, you can use the following helper as well


```suggestion
    runStatus: faker.helpers.enumValue(PipelineStatus)
```

PR #5206, opened 2025-11-18
RI-7740: update instances navigation styles

KrumTy:
# What

- Updates the Link component to support underlining
- Updates styles for databases/rdi instances navigations

# Testing

| Before | After |
| --- | --- |
| Screenshot 2025-11-18 at 18 08 42 | Screenshot 2025-11-18 at 18 08 26 |
| Screenshot 2025-11-18 at 18 08 52 | Screenshot 2025-11-18 at 18 08 13 |

### DIFF
--- a/redisinsight/ui/src/components/instance-header/InstanceHeader.tsx
+++ b/redisinsight/ui/src/components/instance-header/InstanceHeader.tsx
@@ -44,6 +44,7 @@ import { EditIcon } from 'uiSrc/components/base/icons'
 import { Text } from 'uiSrc/components/base/text'
 import { NumericInput } from 'uiSrc/components/base/inputs'
 import { RiIcon } from 'uiSrc/components/base/icons/RiIcon'
+import { Link } from 'uiSrc/components/base/link/Link'
 import InstancesNavigationPopover from './components/instances-navigation-popover'
 import styles from './styles.module.scss'
 
@@ -159,8 +160,10 @@ const InstanceHeader = ({ onChangeDbIndex }: Props) => {
                       : 'Redis Databases'
                   }
                 >
-                  <Text
-                    className={styles.breadCrumbLink}
+                  <Link
+                    color="subdued"
+                    underline
+                    variant="inline"
                     aria-label={
                       server?.buildType === BuildType.RedisStack
                         ? 'Edit database'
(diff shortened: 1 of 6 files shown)


redisinsight/ui/src/components/instance-header/components/instances-navigation-popover/InstancesNavigationPopover.styles.ts
```diff
@@ -0,0 +1,7 @@
+import styled from 'styled-components'
+
+import { Row } from 'uiSrc/components/base/layout/flex'
+
+export const ButtonWrapper = styled(Row)`
+  cursor: pointer;
+`
```
ArtemHoruzhenko, line 7: not sure if this comment is valid but I saw we used `role="button"` property in other places to achieve this. or i'm missing something

redisinsight/ui/src/components/instance-header/InstanceHeader.tsx
```diff
@@ -159,19 +160,20 @@ const InstanceHeader = ({ onChangeDbIndex }: Props) => {
                       : 'Redis Databases'
                   }
                 >
-                  <Text
-                    className={styles.breadCrumbLink}
+                  <Link
+                    color="subdued"
```
valkirilov, line 164: In most of the places we go away from this legacy EUI `subdued` color, is there a reason why we use it here?

redisinsight/ui/src/components/instance-header/InstanceHeader.tsx
```diff
@@ -44,6 +44,7 @@ import { EditIcon } from 'uiSrc/components/base/icons'
 import { Text } from 'uiSrc/components/base/text'
 import { NumericInput } from 'uiSrc/components/base/inputs'
 import { RiIcon } from 'uiSrc/components/base/icons/RiIcon'
+import { Link } from 'uiSrc/components/base/link/Link'
 import InstancesNavigationPopover from './components/instances-navigation-popover'
 import styles from './styles.module.scss'
```
valkirilov, line 49: Tip: In the last few weeks, I have tried to remove all the SASS styles and migrate them to styled components when I deal with a particular component. 

If you find the styles of these components to be simple enough, you can go for it as well.

redisinsight/ui/src/components/instance-header/InstanceHeader.tsx
KrumTy: yes, the default color of our Link is overrided for some reason. This color is the bluish one

redisinsight/ui/src/components/instance-header/InstanceHeader.tsx
KrumTy: We have a dedicated tech debt task on the matter. I'd rather keep this PR short.
I usually do what you are saying as well, but only if I touch more components and half or fewer remain is sass

PR #5207, opened 2025-11-18
RI-7728: Show different statuses for RDI configuration

pawelangelow:
# What

With this PR, we show different statuses for the RDI configuration:
- all good (what we currently always display)
- modified file
- error in the file
- both error and modification

# Testing

## What we currently display = no errors, no changes state

Screenshot 2025-11-18 at 18 34 05

## New state - there are changes, but no error

Screenshot 2025-11-18 at 18 33 49

## New state - there is an error/s

Screenshot 2025-11-18 at 18 36 26

## New state - both: error and modification

Screenshot 2025-11-18 at 18 33 21

### DIFF
--- a/redisinsight/ui/src/pages/rdi/pipeline-management/components/navigation/cards/hooks/useConfigurationState.spec.ts
+++ b/redisinsight/ui/src/pages/rdi/pipeline-management/components/navigation/cards/hooks/useConfigurationState.spec.ts
@@ -1,5 +1,5 @@
 import { renderHook } from 'uiSrc/utils/test-utils'
-import { rdiPipelineSelector } from 'uiSrc/slices/rdi/pipeline'
+import { rdiPipelineSelector, initialState } from 'uiSrc/slices/rdi/pipeline'
 import { IStateRdiPipeline, FileChangeType } from 'uiSrc/slices/interfaces'
 import { useConfigurationState } from './useConfigurationState'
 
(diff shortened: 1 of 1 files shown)


redisinsight/ui/src/pages/rdi/pipeline-management/components/navigation/cards/hooks/useConfigurationState.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  it('should handle changes in other files without affecting config state', () => {
+    mockRdiPipelineSelector.mockReturnValue({
+      changes: {
+        job1: 'modified',
+        job2: 'added',
+        // no config changes
+      },
+      configValidationErrors: [],
+    } as any)
```
valkirilov: nit: Is there a way to not discard the types with `any`, or is it too complicated?

valkirilov (approved):

redisinsight/ui/src/pages/rdi/pipeline-management/components/navigation/cards/hooks/useConfigurationState.spec.ts
pawelangelow: kind of hacky one, but take a look https://github.com/redis/RedisInsight/pull/5207/commits/0d25da7028fbf1154986c5f24a2bada6dc739583

The problem is that I don't want to bring the whole slice here, just the properties we need (so that's why we can't use directly the slice type)

pawelangelow: Nah, I didn't like it, so I changed it like this: https://github.com/redis/RedisInsight/pull/5207/commits/ced945ec06183277202b6de7f301d70cd9815122

The reason not to like it is because of the explicit casting, which breaks the coupling between the types in the redux slice and the ones in the test file. A.k.a. from now on, if we have some modifications on the slice types, the tests would know of that.

valkirilov: > The problem is that I don't want to bring the whole slice here, just the properties we need (so that's why we can't use directly the slice type)

Yeah, that's what I thought the case was. But it looks much better now, so thanks! 🍕

PR #5212, opened 2025-11-19
RI-7681: Polish forms overall

dantovska:
# What

- Remove extra space above db index when cloning sentinel db
- Replace Spacers with parent Col element and gaps - Part 1 (still some work left to do)

# Testing

| Before | After | 
| --- | --- | 
| image | image |

### DIFF
--- a/redisinsight/ui/src/pages/home/components/form/sentinel/PrimaryGroupSentinel.tsx
+++ b/redisinsight/ui/src/pages/home/components/form/sentinel/PrimaryGroupSentinel.tsx
@@ -7,17 +7,15 @@ import { FormField } from 'uiSrc/components/base/forms/FormField'
 import { TextInput } from 'uiSrc/components/base/inputs'
 
 export interface Props {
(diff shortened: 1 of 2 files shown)


redisinsight/ui/src/pages/home/components/form/SSHDetails.tsx
```diff
@@ -67,167 +66,159 @@ const SSHDetails = (props: Props) => {
...
-                    value={Number(formik.values.sshPort)}
-                    onChange={(value) => formik.setFieldValue('sshPort', value)}
-                    onFocus={selectOnFocus}
-                  />
-                </FormField>
-              </FlexItem>
-            </Row>
+        <Col gap="l">
+          <Row gap="m" responsive className={flexGroupClassName}>
+            <FlexItem grow className={flexItemClassName}>
```
valkirilov: Note: Do we need to keep all these SASS classnames, or can we replace them with styled components?

PR #5217, opened 2025-11-19
RI-7745: Revert the POC for icon theming 

dantovska:
# What

### Revert the POC for icon theming handling in icon registry as it was breaking the workbench plugins

# Testing


dantovska:
decided to go with another approach fixing the plugins - https://github.com/redis/RedisInsight/pull/5215. this change is improvement and should stay.

PR #5233, opened 2025-11-25
Fe/bugfix/refactor ri tooltip to better use styled components

pd-redis:
# What

[Tech Debt]
Refactored `RiPopover` component to improve styled-components compatibility and extensibility while maintaining full backwards compatibility.

## Changes

- **Added `trigger` prop** (optional): New preferred prop for popover trigger element, works alongside existing `button` prop.
  - the idea is, that trigger sounds more close to what that prop is doing, right now you can consider this `alias` of button, with
    the idea to move to it entirely over time and remove `button` usages
- **Added `className` prop** (optional): New preferred prop for popover content wrapper styling, using standard prop name for better styled-components integration.
  - it is passed to Popover.Compose, so it RiPopover can be styled directly and that will be reflected on the content wrapper
- **Added `standalone` prop** (optional): When `true`, the trigger renders directly without a span wrapper to avoid DOM interference with styling/layout. Requires the trigger element to be a base DOM element (`div`, `span`, etc.) or a component that forwards refs.
- **Backwards compatibility**: Existing `button` and `panelClassName` props continue to work unchanged:
  - `button` always wraps in span by default (legacy behavior preserved)
  - `panelClassName` falls back when `className` is not provided
- **Migration warnings**: Console warnings when both old and new props are provided together

## Tech Decisions

- **Backwards compatible approach**: Instead of breaking changes, added new optional props that work alongside existing ones. This allows gradual migration without disrupting existing code.
- **Controwll wrapping with `standalone` prop**: Rather than automatically wrapping trigger in span,
  use an explicit `standalone` prop to control trigger wrapping.
  This gives developers clear if they need it.
- **Standard prop names**: Using `className` instead of `panelClassName` aligns with styled-components and other CSS-in-JS libraries that expect standard React prop names. And will work out of the box for custom styling of popover contents

## Why

This refactoring enables better integration with styled-components and similar styling libraries:

1. **Standard prop names**: Styled-components work seamlessly with standard `className` prop, making it easy to style RiPopover's component with styled-components.
2. **Use the best component for the usecase**:

- sometimes you need `inline` element, other times `block`, `flex` or whatever element for trigger of the popover. This is the Element that the user clicks on to show the popover.
- Right now to achieve this, you need an `anchorClassName` (found more than 20 usages in the codebase).
  It should be a css class stored somewhere (global or module stylesheet), which while it is not a problem in
  general, creates discrepancy with how we want to style things (using styled-components).
- If you don't care that defalut wrapper `span` is used, don't need to change anything.
- If you however need custom wrapper, just wrap your trigger with a basic `span`, `div` or `styled-component` and you are done.

# Testing

- No breaking changes - all ~40 files using `RiPopover` remain unaffected

## Manual Testing

- Verified existing `RiPopover` usages continue to work (e.g., `CancelButton`, `EditablePopover`, `ConfirmationPopover`, `TagsCellHeader`)
- Tested new `trigger` or `button` prop with `standalone={true}` (renders directly without span wrapper)
- Tested new `trigger` or `button` prop without `standalone` (wraps in span as expected)
- Tested `className` prop works correctly for styled-components integration
- Verified warnings appear in console when both `button`/`trigger` or `panelClassName`/`className` are provided

---

> [!NOTE]
> Refactors `RiPopover` to improve flexibility and styled-components compatibility while preserving backwards compatibility.
> 
> - Adds `trigger`, `className`, and `standalone` props; prefers `trigger` over legacy `button` and `className` over `panelClassName`, with console warnings on conflicts
> - Adjusts trigger rendering: wraps by default; renders directly when `standalone` is true; wraps scalars for `standalone`
> - Centralizes config in `RiPopover.constants` and renames to `ANCHOR_POSITION_MAP` and `PANEL_PADDING_SIZE_MAP`; updates types to `RiPopover.types` and re-exports via `index.ts`
> - Applies padding via `panelPaddingSize` using `PANEL_PADDING_SIZE_MAP`; preserves existing behavior for `onClickOutside`, focus, placement, and align
> - Adds comprehensive unit tests (`RiPopover.spec.tsx`) covering rendering, new/legacy props, conflict warnings, padding, and scalar triggers
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3abf5bd9cf300b762dc7a9d53e703e16c6824908. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/components/base/popover/RiPopover.tsx
+++ b/redisinsight/ui/src/components/base/popover/RiPopover.tsx
@@ -2,9 +2,12 @@ import React from 'react'
 import { Popover } from '@redis-ui/components'
 
 import * as keys from 'uiSrc/constants/keys'
-import { RiPopoverProps } from './types'
-import { anchorPositionMap, panelPaddingSizeMap } from './config'
 import { OutsideClickDetector } from 'uiSrc/components/base/utils'
+import { type RiPopoverProps } from './RiPopover.types'
+import {
+  ANCHOR_POSITION_MAP,
+  PANEL_PADDING_SIZE_MAP,
+} from './RiPopover.constants'
 
 export const RiPopover = ({
   isOpen,
@@ -65,10 +68,10 @@ export const RiPopover = ({
   }
 
   const placement =
-    anchorPosition && anchorPositionMap[anchorPosition]?.placement
-  const align = anchorPosition && anchorPositionMap[anchorPosition]?.align
+    anchorPosition && ANCHOR_POSITION_MAP[anchorPosition]?.placement
+  const align = anchorPosition && ANCHOR_POSITION_MAP[anchorPosition]?.align
   // TODO: maybe use wrapped popover instead of inline style?!
-  const padding = panelPaddingSize && panelPaddingSizeMap[panelPaddingSize]
+  const padding = panelPaddingSize && PANEL_PADDING_SIZE_MAP[panelPaddingSize]
   return (
     <Popover
       {...props}
@@ -80,18 +83,18 @@ export const RiPopover = ({
           closePopover?.(event as any)
         }
       }}
-    persistent={persistent}
+      persistent={persistent}
       content={
-      children && customOutsideDetector ? (
-        <OutsideClickDetector
-          onOutsideClick={(event) => closePopover?.(event as any)}
-        >
-          {children as JSX.Element}
-        </OutsideClickDetector>
-      ) : (
-        children
-      )
-    }
+        children && customOutsideDetector ? (
+          <OutsideClickDetector
+            onOutsideClick={(event) => closePopover?.(event as any)}
+          >
+            {children as JSX.Element}
+          </OutsideClickDetector>
+        ) : (
+          children
+        )
+      }
       // Props passed to the children wrapper:
       className={activeClassName}
       maxWidth={maxWidth}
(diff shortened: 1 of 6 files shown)


redisinsight/ui/src/components/base/popover/RiPopover.tsx
```diff
@@ -2,57 +2,110 @@ import React from 'react'
...
+      "[RiPopover]: Both 'button' and 'trigger' props are provided. Using 'trigger'. Please migrate to 'trigger' prop.",
+    )
+  }
+
+  // Warn if both panelClassName and className are provided
+  if (panelClassName !== undefined && className !== undefined) {
+    console.warn(
+      "[RiPopover]: Both 'panelClassName' and 'className' props are provided. Using 'className'. Please migrate to 'className' prop.",
+    )
+  }
```
pawelangelow, line 42: I guess we want to log these on dev only - `process.env.NODE_ENV !== 'production'`.

And additionally if we wrap them in `useEffect({}, [])` to display warnings once

pawelangelow:
Suggestion for the PR name: `Refactor RiPopover for styled-components compatibility`

redisinsight/ui/src/components/base/popover/RiPopover.tsx
pd-redis: We don't do any of that for any component and if accepted, I would prefer to leave these in to be loud and obnoxious, so that we don't leave them as is

pawelangelow: I don't have a strong opinion on it. It seems strange to have such logs for the production build to me.

PR #5243, opened 2025-11-26
RI-7774 Fix Copilot button styles

valkirilov:
See https://github.com/redis/RedisInsight/pull/5247


redisinsight/ui/src/components/triggers/copilot-trigger/CopilotTrigger.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+    width: 21px;
+    height: 21px;
+    color: var(--triggerIconActiveColor);
+  }
+
+  ${({ isOpen }) =>
+    isOpen &&
+    css`
+      background-color: ${({ theme }) =>
+        theme.semantic.color.background.primary200};
```
KrumTy, line 27: I wonder if it's possible to do this without nested template string literals

redisinsight/ui/src/components/triggers/copilot-trigger/CopilotTrigger.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+export const CopilotIconButton = styled(IconButton)<
+  { isOpen: boolean } & IconButtonProps
+>`
+  padding: ${({ theme }) => theme.core.space.space200};
+
+  // TODO: Remove this once size property is enabled for IconButton
+  svg {
+    width: 21px;
+    height: 21px;
+    color: var(--triggerIconActiveColor);
```
KrumTy, line 20: is there a theme color that can work here?

dantovska: maybe the color used for the lightbulb icon (insights) when it is highlighted will be a good fit?

redisinsight/ui/src/components/triggers/copilot-trigger/CopilotTrigger.styles.ts
valkirilov: Yes, we can. Applied to both files (the one changed in the PR + the ones that was used as an example ) https://github.com/redis/RedisInsight/pull/5247/commits/3c26fa929fed4fc5c9209c05f33bba1bc421eba0

redisinsight/ui/src/components/triggers/copilot-trigger/CopilotTrigger.styles.ts
valkirilov: Actually, we have the correct color in the palette, thanks! https://github.com/redis/RedisInsight/pull/5247/commits/e66da3ee123b1bbbf58aabba2280c4bfee4ddb3e

PR #5256, opened 2025-11-27
RI-7799: Bulk delete in batches

pawelangelow:
# What

The current bulk delete implementation keeps all deleted keys in memory. When the number grows large (e.g., around 1M keys), the backend eventually crashes.
This PR introduces batch processing to avoid that issue.

However, reports now only return the keys from the last processed batch, meaning the user can no longer see the complete list of deleted keys. So that's why I implemented https://github.com/redis/RedisInsight/pull/5257

# Testing

## Before

Fill a database with enough keys (depending on your RAM), and bulk delete them all.

https://github.com/user-attachments/assets/5efc348d-04ce-41ba-b0b6-4c16f38c7175

## After

https://github.com/user-attachments/assets/ef8a9404-5ec9-4bd7-9093-2b72154b4c85

### DIFF
--- a/redisinsight/api/config/default.ts
+++ b/redisinsight/api/config/default.ts
@@ -451,4 +451,8 @@ export default {
     queryMaxNestedElements:
       parseInt(process.env.RI_AI_QUERY_MAX_NESTED_ELEMENTS, 10) || 25,
   },
+  bulk_actions: {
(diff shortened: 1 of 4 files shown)


pawelangelow:
I'm closing this as I got better solution to the problem.

PR #5266, opened 2025-11-28
RI-7778 Add proper error messages for invalid cloud database endpoints

valkirilov:
# What

Fixes the issue where users see raw JavaScript errors like "Cannot read properties of null (reading 'split')" when adding cloud databases with invalid endpoints. Now returns user-friendly error messages instead.

# Testing

_Prerequisites: You need a Redis Cloud account with a PRO-tier subscription database with a disabled public endpoint._

1. Open Redis Insight and login to Redis Cloud (top right corner)
2. From the top right corner, click on your avatar and then click "**Import Cloud Databases**"
3. Select your subscription from the "**Redis Cloud Subscriptions**" screen
4. Select your databases from the "**Redis Cloud Databases**" screen

You'll be redirected to the "**Redis Enterprise Databases Added**" screen, once the import is done. Observe the errors in the Results column.

| Before | After |
| - | - |
image|Screenshot 2025-12-01 at 17 04 07

---

> [!NOTE]
> Introduce `CloudDatabaseEndpointInvalidException` and use it across autodiscovery and cloud jobs; add corresponding error code/message, UI code, tests, and docs updates.
> 
> - **Backend**:
>   - **Custom exception**: Add `CloudDatabaseEndpointInvalidException` (`api/src/modules/cloud/job/exceptions/cloud-database-endpoint-invalid.exception.ts`) and export from `exceptions/index.ts`.
>   - **Autodiscovery**: In `cloud-autodiscovery.service.ts`, validate `publicEndpoint`; on missing value, return fail response using the new exception instead of splitting the endpoint.
>   - **Cloud jobs**: In `create-free-database.cloud-job.ts` and `import-free-database.cloud-job.ts`, throw `CloudDatabaseEndpointInvalidException` when `publicEndpoint` is absent.
>   - **Constants**: Add `CustomErrorCodes.CloudDatabaseEndpointInvalid` and `ERROR_MESSAGES.CLOUD_DATABASE_ENDPOINT_INVALID`.
> - **Tests**:
>   - `cloud-autodiscovery.service.spec.ts`: Add parameterized tests for `null`/`undefined` `publicEndpoint`; ensure proper error structure and no DB creation.
> - **Frontend**:
>   - Update `ui/src/constants/customErrorCodes.ts` with `CloudDatabaseEndpointInvalid` code.
> - **Docs**:
>   - Backend rules: add guidance on custom exceptions with example.
>   - Testing rules: add `it.each` parameterized tests section.
>   - Commits rules: emphasize keeping commit bodies concise.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 90f362cf142d2087339bf990b3d07cc8d57b8b09. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/api/src/constants/error-messages.ts
+++ b/redisinsight/api/src/constants/error-messages.ts
@@ -145,7 +145,7 @@ export default {
   CLOUD_DATABASE_IMPORT_FORBIDDEN:
     'Adding your Redis Cloud database to Redis Insight is disabled due to a setting restricting database connection management.',
   CLOUD_DATABASE_ENDPOINT_INVALID:
-    'Database endpoint is not available or not fully provisioned yet.',
+    'Database endpoint is unavailable. It may still be provisioning or has been disabled.',
   CLOUD_PLAN_NOT_FOUND_FREE: 'Unable to find free cloud plan',
   CLOUD_SUBSCRIPTION_ALREADY_EXISTS_FREE: 'Free subscription already exists',
   COMMON_DEFAULT_IMPORT_ERROR: 'Unable to import default data',


redisinsight/api/src/modules/cloud/database/utils/cloud-data-converter.ts
```diff
@@ -17,6 +17,14 @@ export function convertRedisCloudModuleName(name: string): string {
   return REDIS_CLOUD_MODULES_NAMES[name] ?? name;
 }
 
+export function isValidCloudDatabaseEndpoint(
+  publicEndpoint: string | null | undefined,
+): boolean {
+  return (
+    !!publicEndpoint && !!publicEndpoint.trim() && publicEndpoint.includes(':')
```
KrumTy: is this the best way to check it or there can be a regex with more rules?

.ai/rules/backend.md
```diff
@@ -141,6 +141,43 @@ Use appropriate exception types:
...
+export class FeatureInvalidException extends HttpException {
+  constructor(
+    message = ERROR_MESSAGES.FEATURE_INVALID,
+    options?: HttpExceptionOptions,
+  ) {
+    const response = {
+      message,
+      statusCode: HttpStatus.BAD_REQUEST,
+      error: 'FeatureInvalid',
+      errorCode: CustomErrorCodes.FeatureInvalid,
```
KrumTy, line 173: does it make sense to enforce that these fields are provided for all exceptions (e.g. extract and implement a common interface)? Currently it's typed as `Record<string, any>` and if we have a common interface, we can have more predictive api responses

redisinsight/api/src/modules/cloud/database/utils/cloud-data-converter.ts
valkirilov: Actually, all we care about is whether there is a value for the publicEndpoint, but since we split it later in the code by `:`, the AI considered it makes sense to check it here as well. 

For sure, we can make the validation smarter, but I don't have the requirements for it, so here we're just improvising.

.ai/rules/backend.md
valkirilov: This is the current pattern, so I just plugged into it. 

What you suggest sounds reasonable to me, but I believe @ArtemHoruzhenko can advise on architectural changes like that.

redisinsight/api/src/constants/error-messages.ts
```diff
@@ -144,6 +144,8 @@ export default {
   CLOUD_DATABASE_ALREADY_EXISTS_FREE: 'Free database already exists',
   CLOUD_DATABASE_IMPORT_FORBIDDEN:
     'Adding your Redis Cloud database to Redis Insight is disabled due to a setting restricting database connection management.',
+  CLOUD_DATABASE_ENDPOINT_INVALID:
+    'Database endpoint is not available or not fully provisioned yet.',
```
ArtemHoruzhenko: Can we change the message? 
I would say that we must say something like: "Public endpoint is disabled for this database"

redisinsight/api/src/modules/cloud/database/utils/cloud-data-converter.ts
ArtemHoruzhenko: I would say we just need to check if we have not empty string. It should be enough.

redisinsight/api/src/constants/error-messages.ts
valkirilov: Of course, we can put whatever we want.

redisinsight/api/src/modules/cloud/database/utils/cloud-data-converter.ts
valkirilov: Ok, done. I have simplified the validation so we check only whether the `publicEndpoint` is present 👍

redisinsight/api/src/constants/error-messages.ts
valkirilov: After an internal sync, we'll go with the following error message (https://github.com/redis/RedisInsight/pull/5266/commits/90f362cf142d2087339bf990b3d07cc8d57b8b09):

> Database endpoint is unavailable. It may still be provisioning or has been disabled.

PR #5278, opened 2025-12-01
RI:7799: Stream bulk delete report

pawelangelow:
# What

Bulk delete can include a huge number of keys, and our current implementation keeps them in memory. So the robust solution in this case is to stream all the data to a report (successes and failures).

So these changes on a high level look like this:

high-level-overview

- User initiates bulk delete with "Download report" checkbox enabled => sends `generateReport: true`
- Backend creates bulk action and returns  `downloadUrl` in the overview, **but waits before processing** (via `waitForStreamIfNeeded()`)
- Frontend immediately triggers HTTP GET to `/bulk-actions/:id/report` (automatic, no button click)
- Backend attaches HTTP response stream to BulkAction via `setStreamingResponse(res)`:
  - Writes report header directly to HTTP stream
  - Resolves the promise => unblocks processing
- Processing runs with direct HTTP streaming:
  - Runner deletes keys via Redis
  - Calls `bulkAction.writeToReport()` => writes directly to HTTP response
  - Only tracks counts in memory (addFailed(1), addSuccess(1))
- When complete, `finalizeReport()` writes summary and calls `res.end()`
- Browser receives the complete file and saves it automatically

## Update

After the copilot suggestion - if the HTTP request that starts the actual delete process, for some reason, hangs, the user receives an error that the bulk delete didn't work. If they disable the report, the delete process starts immediately.

https://github.com/user-attachments/assets/d3fc2165-493a-4e05-895a-0c16d50de807

# Testing

1. Set up a database with a large enough set (that depends on the RAM, but let's say > 1M keys)
2. Bulk delete everything (`*`)

Expected result: All good
Actual result: it fails at some point, like in the following recording:

https://github.com/user-attachments/assets/ad47ef4d-649f-4371-8258-6d11b2622762

## After

https://github.com/user-attachments/assets/e33ed1be-9d97-461b-800e-6d9264cfd2c5

## The result log format

```
Bulk Delete Report
Command Executed for each key: UNLINK key_name
A summary is provided at the end of this file.
==================
string:926189__53.... - OK.  <--- when delete was successful
string:941__807a7..... - Error: NOPERM No permissions to access a key <--- when there is an error
...
=============
Summary:
=============
Status: completed
Processed: 1000 keys
Succeeded: 2 keys
Failed: 998 keys
```

---

> [!NOTE]
> Adds streaming bulk delete reports with a new download endpoint, updates processing to write directly to the stream, and introduces a UI checkbox that auto-starts the download.
> 
> - **Backend**:
>   - **Streaming reports**: Add `GET /bulk-actions/:id/report/download` in `bulk-actions.controller.ts` and `streamReport(id, res)` in `BulkActionsService` to set headers and attach the response stream.
>   - **BulkAction streaming**: Implement report flow in `models/bulk-action.ts` (`waitForStreamIfNeeded`, `setStreamingResponse`, `writeToReport`, `finalizeReport`, timeout handling) and include `downloadUrl` and `error` in `getOverview`.
>   - **DTO/Provider**: Add `generateReport?: boolean` to `CreateBulkActionDto`; pass through in `BulkActionsProvider.create`.
>   - **Runners**: Update simple runner to stop accumulating errors/keys in memory and call `bulkAction.writeToReport(...)` while maintaining counters.
>   - **Interfaces**: Extend `IBulkActionOverview` with `downloadUrl` and `error`; extend `IBulkAction` with `writeToReport`.
> - **Frontend**:
>   - **UI toggle**: Add "Download report" checkbox (Redux `bulkDelete.generateReport`) to control report generation.
>   - **Auto-download**: On create, include `generateReport`; when `downloadUrl` arrives, trigger download via `triggerDownloadFromUrl` using `getBaseUrl()`.
>   - **Status/UX**: Show failure banner with server `error`; remove legacy summary download button.
>   - **Utils**: Add `triggerDownloadFromUrl` helper.
> - **Tests**:
>   - Add/expand unit tests for streaming flow (service, BulkAction, runner) and UI checkbox behavior; remove obsolete summary button tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9ce17acf10cd4ed4399186499c1aa825cdec52cb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/api/src/modules/bulk-actions/models/bulk-action.spec.ts
+++ b/redisinsight/api/src/modules/bulk-actions/models/bulk-action.spec.ts
@@ -512,7 +512,7 @@ describe('AbstractBulkActionSimpleRunner', () => {
     });
 
     describe('setStreamingResponse', () => {
-      it('should set streaming response and resolve promise', () => {
+      it('should set streaming response and resolve promise when waiting', async () => {
         const bulkActionWithReport = new BulkAction(
           mockCreateBulkActionDto.id,
           mockCreateBulkActionDto.databaseId,
@@ -523,12 +523,18 @@ describe('AbstractBulkActionSimpleRunner', () => {
           true,
         );
 
+        // Start waiting for stream
+        const waitPromise = bulkActionWithReport['waitForStreamIfNeeded']();
+
+        // Set streaming response
         bulkActionWithReport.setStreamingResponse(mockResponse);
+
+        await waitPromise;
 
         expect(bulkActionWithReport['streamingResponse']).toBe(mockResponse);
       });
 
-      it('should write header when response is set', () => {
+      it('should write header when response is set', async () => {
         const bulkActionWithReport = new BulkAction(
           mockCreateBulkActionDto.id,
           mockCreateBulkActionDto.databaseId,
@@ -539,12 +545,40 @@ describe('AbstractBulkActionSimpleRunner', () => {
           true,
         );
 
+        // Start waiting for stream
+        const waitPromise = bulkActionWithReport['waitForStreamIfNeeded']();
+
+        // Set streaming response
         bulkActionWithReport.setStreamingResponse(mockResponse);
+
+        await waitPromise;
 
         expect(mockResponse.write).toHaveBeenCalled();
         const headerCall = mockResponse.write.mock.calls[0][0];
         expect(headerCall).toContain('Bulk Delete Report');
         expect(headerCall).toContain('Command Executed for each key:');
+      });
+
+      it('should immediately end response when called without waiting', () => {
+        const bulkActionWithReport = new BulkAction(
+          mockCreateBulkActionDto.id,
+          mockCreateBulkActionDto.databaseId,
+          mockCreateBulkActionDto.type,
+          mockBulkActionFilter,
+          mockSocket,
+          mockBulkActionsAnalytics() as any,
+          true,
+        );
+
+        // Call setStreamingResponse without waitForStreamIfNeeded being called first
+        bulkActionWithReport.setStreamingResponse(mockResponse);
+
+        // Should immediately end the response
+        expect(mockResponse.write).toHaveBeenCalledWith(
+          'Unable to generate report. Please try again.\n',
+        );
+        expect(mockResponse.end).toHaveBeenCalled();
+        expect(bulkActionWithReport['streamingResponse']).toBeNull();
       });
     });
 
@@ -727,24 +761,6 @@ describe('AbstractBulkActionSimpleRunner', () => {
         jest.useRealTimers();
       });
 
-      it('should resolve immediately when streamingResponse is already set', async () => {
-        const bulkActionWithReport = new BulkAction(
-          mockCreateBulkActionDto.id,
-          mockCreateBulkActionDto.databaseId,
-          mockCreateBulkActionDto.type,
-          mockBulkActionFilter,
-          mockSocket,
-          mockBulkActionsAnalytics() as any,
-          true,
-        );
-
-        bulkActionWithReport.setStreamingResponse(mockResponse);
-
-        await bulkActionWithReport['waitForStreamIfNeeded']();
-
-        expect(bulkActionWithReport['streamingResponse']).toBe(mockResponse);
-      });
-
       it('should clear timeout when stream is set before timeout', async () => {
         jest.useFakeTimers();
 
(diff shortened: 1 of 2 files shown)


redisinsight/api/src/modules/bulk-actions/interfaces/bulk-action.interface.ts
```diff
@@ -7,4 +7,5 @@ export interface IBulkAction {
   getFilter(): BulkActionFilter;
   changeState(): void;
   getSocket(): Socket;
+  writeToReport(keyName: string, success: boolean, error?: string): void;
```
ArtemHoruzhenko: key name shouldn't be a string. We must use Buffers

redisinsight/api/src/modules/bulk-actions/models/runners/simple/abstract.bulk-action.simple.runner.ts
```diff
@@ -86,18 +86,17 @@ export abstract class AbstractBulkActionSimpleRunner extends AbstractBulkActionR
     res: [Error | null, RedisClientCommandReply][],
   ) {
     this.summary.addProcessed(res.length);
-    this.summary.addKeys(keys.filter((_, index) => !res[index][0]));
-
-    const errors = [];
 
     res.forEach(([err], i) => {
+      const keyName = keys[i].toString();
```
ArtemHoruzhenko: We must use Buffers

redisinsight/api/src/modules/bulk-actions/models/bulk-action-summary.spec.ts
```diff
@@ -81,12 +81,30 @@ describe('BulkActionSummary', () => {
...
+      ]);
+    });
+  });
   describe('getOverview', () => {
     it('should get overview and clear errors', async () => {
       expect(summary['processed']).toEqual(0);
       expect(summary['succeed']).toEqual(0);
       expect(summary['failed']).toEqual(0);
       expect(summary['errors']).toEqual([]);
+      expect(summary['keys']).toEqual([]);
```
ArtemHoruzhenko, line 107: do we still have keys property?

redisinsight/api/src/modules/bulk-actions/bulk-actions.controller.ts
```diff
@@ -0,0 +1,37 @@
...
+    description: 'Bulk action id',
+    type: String,
+  })
+  @Get(':id/report/download')
+  async downloadReport(
+    @Param() { id }: BulkActionIdDto,
+    @Res() res: Response,
+  ): Promise<void> {
+    await this.service.streamReport(id, res);
+  }
```
ArtemHoruzhenko, line 36: This solution does not support multi-tenancy we need to rework it in the future

ArtemHoruzhenko (approved):
We can go as is with assumption that we will back to it soon and enhance

redisinsight/ui/src/components/bulk-actions-config/BulkActionsConfig.tsx
```diff
@@ -45,6 +47,16 @@ const BulkActionsConfig = () => {
...
 
+  const triggerDownload = useCallback((downloadUrl: string) => {
+    const link = document.createElement('a')
+    // Build full URL using API base URL
+    link.href = `${getBaseUrl()}${downloadUrl}`
+    link.style.display = 'none'
+    document.body.appendChild(link)
+    link.click()
+    document.body.removeChild(link)
+  }, [])
```
KrumTy: this doesn't need to memozied/inside the component; it should be in some generic util

redisinsight/ui/src/components/bulk-actions-config/BulkActionsConfig.tsx
```diff
@@ -101,8 +113,20 @@ const BulkActionsConfig = () => {
...
+        onBulkDeleting(response)
+
+        // Trigger download if report generation is enabled
+        if (
+          generateReport &&
+          'downloadUrl' in response &&
+          response.downloadUrl
+        ) {
+          triggerDownload(response.downloadUrl)
+        }
```
KrumTy: why not put this logic directly inside onBulkDeleting?

redisinsight/api/src/modules/bulk-actions/models/bulk-action-summary.spec.ts
pawelangelow: Yeah, because of the other bulk action. I want to remove it from the API, but we need to re-implement the bulk import.

pawelangelow: It's not used tho, just not breaking the API.

PR #5296, opened 2025-12-03
RI-7776 Fix cloud discovery search and notifications

ArtemHoruzhenko:
# What

Fixes two issues in the cloud database discovery flow:

1. **Multiple notifications bug**: The "Successfully discovered database(s)..." notification was showing multiple times when interacting with the search field
2. **Limited search**: Search in subscriptions table only worked for name and id fields

### Root Cause

When search filtered items, the notification counts were recalculated from filtered data. This caused the `variant` prop to change (e.g., from "attention" to "success" when filtered results had no failures), which triggered the MessageBar's useEffect to show a new toast.

### Solution

- Calculate notification counts from original `subscriptions` prop instead of filtered `items` state
- Expand search fields to include all relevant columns:
  - Subscriptions: name, id, type, provider, region, status
  - Database results: added statusAdded (success/fail)

# Testing

1. Navigate to Redis Cloud database discovery
2. Verify the summary notification appears only once when data loads
3. Type in the search field - notification should NOT reappear
4. Search by different fields (provider, region, status) - should filter correctly

---

Fixes #RI-7776

---

> [!NOTE]
> Stabilizes notifications by deriving counts from original data and broadens search across subscriptions and database results, with tests added for search behavior.
> 
> - **UI (Cloud Discovery)**:
>   - **Notifications**: Compute `countStatusActive`/`countStatusFailed` from `subscriptions` (not filtered `items`) in `RedisCloudSubscriptions.tsx` to avoid re-triggering toasts on search.
>   - **Search Enhancements**:
>     - `RedisCloudSubscriptions.tsx`: Search now matches `name`, `id`, `type`, `provider`, `region`, `status`.
>     - `RedisCloudDatabasesResult.tsx`: Search now includes `statusAdded` in addition to existing fields.
> - **Tests**:
>   - `RedisCloudSubscriptions.spec.tsx`: Add comprehensive search tests (name, id, provider, region, status, type; case-insensitive) and test scaffolding with `faker`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 93f8edbc0ac7b409b9b28da2ff6f6797550bc450. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/pages/autodiscover-cloud/redis-cloud-subscriptions/RedisCloudSubscriptions/RedisCloudSubscriptions.tsx
+++ b/redisinsight/ui/src/pages/autodiscover-cloud/redis-cloud-subscriptions/RedisCloudSubscriptions/RedisCloudSubscriptions.tsx
@@ -75,12 +75,15 @@ const RedisCloudSubscriptions = ({
     }
   }, [subscriptions, loading])
 
-  const countStatusActive = items.filter(
-    ({ status, numberOfDatabases }: RedisCloudSubscription) =>
-      status === RedisCloudSubscriptionStatus.Active && numberOfDatabases !== 0,
-  )?.length
+  // Calculate counts from original subscriptions to prevent notification re-triggering on search
+  const countStatusActive =
+    subscriptions?.filter(
+      ({ status, numberOfDatabases }: RedisCloudSubscription) =>
+        status === RedisCloudSubscriptionStatus.Active &&
+        numberOfDatabases !== 0,
+    )?.length || 0
 
-  const countStatusFailed = items.length - countStatusActive
+  const countStatusFailed = (subscriptions?.length || 0) - countStatusActive
 
   const handleSubmit = () => {
     onSubmit(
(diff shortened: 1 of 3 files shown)


redisinsight/ui/src/pages/autodiscover-cloud/redis-cloud-subscriptions/RedisCloudSubscriptions/RedisCloudSubscriptions.tsx
```diff
@@ -106,7 +109,11 @@ const RedisCloudSubscriptions = ({
       subscriptions?.filter(
         (item: RedisCloudSubscription) =>
           item.name?.toLowerCase()?.indexOf(value) !== -1 ||
-          item.id?.toString()?.toLowerCase().indexOf(value) !== -1,
+          item.id?.toString()?.indexOf(value) !== -1 ||
```
pawelangelow, line 112: I'm wondering can't we add such util:
```
export const filterBySearchTerm = <T extends Record<string, unknown>>(
  items: T[],
  searchTerm: string,
  fields: (keyof T)[],
): T[] => {
  const term = searchTerm.toLowerCase()
  
  return items.filter((item) =>
    fields.some((field) => {
      const value = item[field]
      return String(value ?? '').toLowerCase().includes(term)
    }),
  )
}


or

export const matchesSearch = (
  fieldValue: unknown,
  searchTerm: string, // expected to be already lowercased
): boolean => String(fieldValue ?? '').toLowerCase().includes(searchTerm)
```
and then this syntax would be easier to be read and moreover we have it in the `RedisCloudDatabasesResult.tsx` as well.


Maybe not for this PR, but actually, why not 😸

PR #5301, opened 2025-12-03
RI-7781: fix copy module name test to verify clipboard functionality

pd-redis:
Previous test did not actually test anything related to copying

# What

Fix the `DatabaseListModules` test that was supposed to verify copy-to-clipboard functionality but wasn't actually testing it.

# Testing

1. Run the test:
  redisinsight/ui/src/components/database-list-modules/DatabaseListModules.spec.tsx
   2. Verify both tests pass:
   - `should render`
   - `should copy module name to clipboard when clicked`

References: #RI-7781

---

> [!NOTE]
> Updates DatabaseListModules test to mock clipboard and assert module click copies the correct name.
> 
> - **Tests**:
>   - `redisinsight/ui/src/components/database-list-modules/DatabaseListModules.spec.tsx`:
>     - Mock `navigator.clipboard.writeText` and assert it’s called with the selected module’s term.
>     - Rename test to clarify behavior: copy module name on click.
>     - Remove unnecessary `act` usage and redundant render/assertions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2e9d0a9e77894d3c787a6a5f9b72f864acda0579. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/components/database-list-modules/DatabaseListModules.spec.tsx
+++ b/redisinsight/ui/src/components/database-list-modules/DatabaseListModules.spec.tsx
@@ -4,7 +4,7 @@ import {
   RedisDefaultModules,
   DATABASE_LIST_MODULES_TEXT,
 } from 'uiSrc/slices/interfaces'
-import { fireEvent, render, act } from 'uiSrc/utils/test-utils'
+import { fireEvent, render } from 'uiSrc/utils/test-utils'
 import { AdditionalRedisModule } from 'apiSrc/modules/database/models/additional.redis.module'
 import { DatabaseListModules } from './DatabaseListModules'
 import { DatabaseListModulesProps } from './DatabaseListModules.types'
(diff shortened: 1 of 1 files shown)


redisinsight/ui/src/components/database-list-modules/DatabaseListModules.spec.tsx
```diff
@@ -33,27 +33,22 @@ describe('DatabaseListModules', () => {
...
+    })
+
     const { queryByTestId } = render(
       <DatabaseListModules {...instance(mockedProps)} modules={modulesMock} />,
     )
 
     const term = DATABASE_LIST_MODULES_TEXT[RedisDefaultModules.Search]
-
     const module = queryByTestId(`${term}_module`)
+    expect(module).toBeInTheDocument()
```
pawelangelow, line 48:
```suggestion
    const module = getByTestId(`${term}_module`)
```

pd-redis: You don't like assertions?

pawelangelow: I do but, it's the same with less syntax. If `get*` fails to find it, it will throw the same as `toBeInDocument`. It's a nit comment, non-blocking.

pd-redis: Yes, but more explicit assert shows the intention. Which for tests is good. I think.

Issue #5310, opened 2025-12-04
[Possible Bug]: CPU Metrics when multiple IO Threads


PinoOG:
**Preconditions** (Any important steps we need to know)
in redis.conf

io-threads X (where X != 1)
io-threads-do-reads yes

**Steps to reproduce** (How to reproduce what you found step by step)
// It depends on the redis node cpu usage (displayed by btop/htop)

**Actual behavior** (A short description of what you found)
When redis-server cpu usage is more than 100% on btop/htop the value in redis insight only reflects up to 100%

**Expected behavior** (A short description of what you expected to find)
Detecting when io-threads are enabled and displaying the CPU output in a format which could be like this %/THREAD_AMOUNT*100 

Example: 4 threads 250/400%

**Screenshots** (Paste or drag-and-drop a screenshot or a link to a recording)
// Not really needed

**Additional context** (Operating system, version of Redis Insight, Redis database version, Redis module version, database type, connection type, logs, or any other information)
os: Linux 6.12.48+deb13-amd64 x86_64
redis-server: 8.2.2
redis-insight: 3.0.0

valkirilov:
Thanks for reporting this odd issue @PinoOG, I managed to reproduce it. We'll take a deeper look at whether it's actually a bug, and we'll let you know when we have more info and progress on this.

<img width="1605" height="919" alt="Image" src="https://github.com/user-attachments/assets/cd312d07-472c-4e20-a49a-5430e9974b80" />

For clarity, here is the Docker setup that I used:

- `docker-compose.yml`
```yml
version: "3.9"

services:
  redis:
    image: redis:8.2
    container_name: redis-io-threads-test
    command: ["redis-server", "/usr/local/etc/redis/redis.conf"]
    volumes:
      - ./redis.conf:/usr/local/etc/redis/redis.conf:ro
    ports:
      - "6379:6379"
    networks:
      - redis-net

  redisinsight:
    image: redis/redisinsight:latest
    container_name: redisinsight-io-threads-test
    depends_on:
      - redis
    ports:
      - "5540:5540"
    networks:
      - redis-net
    # For RedisInsight in Docker, default URL is http://localhost:5540

  loadgen:
    image: redis:8.2
    container_name: redis-loadgen
    depends_on:
      - redis
    networks:
      - redis-net
    entrypoint: ["/bin/sh", "-c"]
    command: >
      "sleep 5;
      echo 'Starting redis-benchmark load loop...';
      while true; do
[...]

PR #5321, opened 2025-12-09
RI-7856 Extend CPU usage to go beyond 100% and reflect shards/threads

valkirilov:
# What

Update the indicator to properly display CPU usage when Redis I/O threads or cluster shards push CPU usage above 100%.

- Added `maxCpuUsagePercentage` to the backend and implemented calculation logic:
  - **Cluster**: Number of primary nodes x number of threads per node × 100%
  - **Standalone**: Detect I/O threads via `CONFIG GET io-threads` and multiply by 100%
- Updated CPU tooltip on the frontend to show "current% / max%" format when max CPU > 100%

### Examples
- 1 standalone with 4 threads: 4 * 100 = 400%
- 3 shards, each with 4 threads: (4+4+4) * 100 = 1200%
- 3 shards, default (1 thread each): (1+1+1) * 100 = 300%

References #5310

# Testing

## Redis database with multiple threads

Use the [following setup](https://github.com/redis/RedisInsight/issues/5310#issuecomment-3615911262) to start a Redis server with 4 threads via Docker.

Then, simply connect to the database and observe the reported CPU usage at the top of the screen.

| Before | After |
| - | - |
image|image

## Standalone Redis

We don't change the default behavior. We show a single number for the current CPU usage when we have a standalone instance with a single thread.

Screenshot 2025-12-09 at 14 45 16

---

> [!NOTE]
> Expose uncapped CPU usage with new maxCpuUsagePercentage derived from threads/shards and update UI to display current/max.
> 
> - **Backend/API**:
>   - Add `maxCpuUsagePercentage` to `DatabaseOverview` and compute it in `DatabaseOverviewProvider`:
>     - Standalone: `CONFIG GET io-threads` × 100.
>     - Cluster: sum primary nodes' I/O threads × 100.
>   - Stop clamping `cpuUsagePercentage` at 100 (return raw sum across shards); add logging for I/O thread fetch failures.
> - **Tests**:
>   - Extend provider specs to cover max CPU calculation (standalone/cluster), error handling, and CPU >100% cases.
>   - API e2e: response schema includes `maxCpuUsagePercentage`; cluster test asserts value equals primary count × 100.
> - **Frontend**:
>   - `OverviewMetrics` displays CPU >100% and shows tooltip as `current% / max%` when `maxCpuUsagePercentage > 100`.
>   - Add helpers `getCpuDisplayValue`/`getCpuTooltipValue`; wire `maxCpuUsagePercentage` through props and state (`instances.ts`).
>   - Add unit tests for CPU display/tooltip behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 44da1f9ec88cabbb26c3d2831665a8aa174415d7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/api/src/modules/database/providers/database-overview.provider.ts
+++ b/redisinsight/api/src/modules/database/providers/database-overview.provider.ts
@@ -295,54 +295,69 @@ export class DatabaseOverviewProvider {
   }
 
   /**
-   * Calculates maximum CPU percentage based on number of nodes/shards
-   * For clusters: max = number of primary nodes * 100% (always returned, no cap)
-   * For standalone: detect I/O threads via CONFIG GET (no estimation)
-   *
-   * Example of calculation:
-   * 1 standalone: 1 * 100 = 100%
-   * 2 shards: 2 * 100 = 200%
-   * 4 threads: 4 * 100 = 400%
+   * Gets the io-threads configuration for a Redis client/node
+   * Returns the number of io-threads, defaulting to 1 if detection fails
+   * @param nodeClient The Redis client to query
+   * @returns The number of io-threads (defaults to 1)
    */
-  private async calculateMaxCpuPercentage(
-    client: RedisClient,
-  ): Promise<number | undefined> {
-    // For cluster, return the number of primary nodes * 100%
-    if (client.getConnectionType() === RedisClientConnectionType.CLUSTER) {
-      try {
-        const primaryNodes = await client.nodes(RedisClientNodeRole.PRIMARY);
-        const maxCpu = primaryNodes.length * 100;
-
-        return maxCpu;
-      } catch (error) {
-        // If we can't get nodes, fall through to standalone logic
-        this.logger.warn(
-          'Error occurred when trying to calculate max CPU usage percentage for cluster',
-          error,
-        );
-      }
-    }
-
-    // For standalone, detect I/O threads via CONFIG GET
+  private async getIoThreadsForNode(nodeClient: RedisClient): Promise<number> {
     try {
-      const ioThreadsResult = await client.call(
+      const ioThreadsResult = await nodeClient.call(
         ['config', 'get', 'io-threads'],
         {
           replyEncoding: 'utf8',
         },
       );
 
-      const ioThreads = parseInt(ioThreadsResult?.[1] || '1', 10);
-
-      return ioThreads * 100;
+      return parseInt(ioThreadsResult?.[1] || '1', 10);
     } catch (error) {
       this.logger.warn(
-        'Error occurred when trying to calculate max CPU usage percentage',
+        'Error getting io-threads, defaulting to 1 thread',
         error,
       );
+      return 1;
+    }
+  }
+
+  /**
+   * Calculates maximum CPU percentage based on number of nodes/shards
+   * For clusters: sum of io-threads across all primary nodes * 100%
+   * For standalone: detect I/O threads via CONFIG GET
+   *
+   * Example of calculation:
+   * 1 standalone with 4 threads: 4 * 100 = 400%
+   * 3 shards, each with 4 threads: (4+4+4) * 100 = 1200%
+   * 3 shards, default (1 thread each): (1+1+1) * 100 = 300%
+   */
+  private async calculateMaxCpuPercentage(
+    client: RedisClient,
+  ): Promise<number | undefined> {
+    // For cluster, sum io-threads across all primary nodes
+    if (client.getConnectionType() === RedisClientConnectionType.CLUSTER) {
+      try {
+        const primaryNodes = await client.nodes(RedisClientNodeRole.PRIMARY);
+        let totalIoThreads = 0;
+
+        // Get io-threads for each primary node
+        for (const node of primaryNodes) {
+          const ioThreads = await this.getIoThreadsForNode(node);
+          totalIoThreads += ioThreads;
+        }
+
+        return totalIoThreads * 100;
+      } catch (error) {
+        // If we can't get nodes, return undefined (don't fall through to standalone logic)
+        this.logger.warn(
+          'Error occurred when trying to calculate max CPU usage percentage for cluster',
+          error,
+        );
+        return undefined;
+      }
     }
 
-    return undefined;
+    // For standalone, detect I/O threads via CONFIG GET
+    const ioThreads = await this.getIoThreadsForNode(client);
+    return ioThreads * 100;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


redisinsight/api/src/modules/database/providers/database-overview.provider.ts
```diff
@@ -47,6 +49,13 @@ export class DatabaseOverviewProvider {
       totalKeysPerDb = calculatedTotalKeysPerDb;
     }
 
+    const cpuUsagePercentage = this.calculateCpuUsage(
+      clientMetadata.databaseId,
+      nodesInfo,
+    );
+
+    const maxCpuUsagePercentage = await this.calculateMaxCpuPercentage(client);
```
ArtemHoruzhenko: very nit but why don't we do the same as for other fields (invoke function inside return block?) seems like we don't use this constants below

redisinsight/api/src/modules/database/providers/database-overview.provider.ts
```diff
@@ -291,6 +298,73 @@ export class DatabaseOverviewProvider {
...
+   *
+   * Example of calculation:
+   * 2 shards: 2 * 100 = 200%
+   * 4 threads: 4 * 100 = 400%
+   */
+  private async calculateMaxCpuPercentage(
+    client: RedisClient,
+  ): Promise<number | undefined> {
+    // For cluster, return the number of primary nodes * 100%
+    if (client.getConnectionType() === RedisClientConnectionType.CLUSTER) {
```
ArtemHoruzhenko, line 336: what about standalone? I thought that this method should process both connections

redisinsight/api/src/modules/database/providers/database-overview.provider.ts
```diff
@@ -291,6 +298,73 @@ export class DatabaseOverviewProvider {
...
+        } else if (typeof ioThreadsValue === 'number') {
+          ioThreads = ioThreadsValue;
+        } else {
+          // If it's a Buffer, convert to string first
+          ioThreads = parseInt(ioThreadsValue?.toString() || '1', 10);
+        }
+
+        if (ioThreads > 1) {
+          return ioThreads * 100;
+        }
```
ArtemHoruzhenko: why we don't return 100?
1 - undefined
2 - 200
3 - 300
...

redisinsight/api/src/modules/database/providers/database-overview.provider.ts
```diff
@@ -291,6 +298,73 @@ export class DatabaseOverviewProvider {
...
+        let ioThreads: number;
+
+        if (typeof ioThreadsValue === 'string') {
+          ioThreads = parseInt(ioThreadsValue, 10);
+        } else if (typeof ioThreadsValue === 'number') {
+          ioThreads = ioThreadsValue;
+        } else {
+          // If it's a Buffer, convert to string first
+          ioThreads = parseInt(ioThreadsValue?.toString() || '1', 10);
+        }
```
ArtemHoruzhenko: I feel like we don't need all these checks. Let's add replyEncoding for call command so we will be sure what returned

redisinsight/api/src/modules/database/providers/database-overview.provider.ts
```diff
@@ -291,6 +298,73 @@ export class DatabaseOverviewProvider {
...
+        'config',
+        'get',
+        'io-threads',
+      ]);
+
+      if (
+        ioThreadsResult &&
+        Array.isArray(ioThreadsResult) &&
+        ioThreadsResult.length >= 2
+      ) {
```
ArtemHoruzhenko: I feel like it might be simplified. E.g. `const ioThreads = parseInt(ioThreadsResult?.[1] || '1', 10);`
without having any if here and below

redisinsight/ui/src/components/database-overview/components/OverviewMetrics/OverviewMetrics.tsx
```diff
@@ -58,10 +59,45 @@ export interface IMetric {
...
+}
+
+export function getCpuTooltipValue(
+  cpuUsagePercentage: number | null,
+  maxCpuUsagePercentage?: number | null,
+): string | null {
+  const hasMaxCpu =
+    maxCpuUsagePercentage !== undefined &&
+    maxCpuUsagePercentage !== null &&
+    maxCpuUsagePercentage > 100
```
ArtemHoruzhenko, line 77: you see. here you already have check for 100% so we can return from API what we determined - 100% not undefined

valkirilov: Agree, will do 👍

redisinsight/api/src/modules/database/providers/database-overview.provider.ts
valkirilov: Done 👍

redisinsight/api/src/modules/database/providers/database-overview.provider.ts
valkirilov: Yes, it's handled below:
- check if we deal with a cluster and count its nodes (and respectively the threads per node)
- otherwise, treat it as a standalone and check whether we have multiple threads within

redisinsight/api/src/modules/database/providers/database-overview.provider.ts
valkirilov: Totally agree, we don't need this constant. Updated to work as you advised 👍

redisinsight/api/src/modules/database/providers/database-overview.provider.ts
valkirilov: I agree, the trick here was to add `replyEncoding`, which I was unaware of. 

Now, when we pass it, we don't need these complex and unnecessary conditional checks, and the code looks simpler (and still works the same way). Thanks for pointing it out!

redisinsight/api/src/modules/database/providers/database-overview.provider.ts
valkirilov: Yes, somehow related to [this comment](https://github.com/redis/RedisInsight/pull/5321#discussion_r2609448037), but now that we have the `replyEncoding` passed, these checks look much simpler. Thanks!

redisinsight/api/src/modules/database/providers/database-overview.provider.ts
ArtemHoruzhenko: yeah, forgot to remove this comment

PR #5407, opened 2026-01-26
RI-7907: Add Azure authentication service with MSAL

pawelangelow:
# What

Implements Azure Entra ID (formerly Azure AD) authentication infrastructure using MSAL (Microsoft Authentication Library).

Backend:

- AzureAuthService - OAuth login flow with PKCE, token management, silent token acquisition
- AzureAuthController - Endpoints: /azure/auth/login, /azure/auth/callback, /azure/auth/logout, /azure/auth/status
- Azure configuration constants (client ID, authority, scopes, redirect URI)

# Testing

- Start the API: `yarn dev:api`
- Open in browser: http://localhost:5540/api/azure/auth/login
- Complete Azure login flow by the generated URL (open a new tab with it)
- Check API console logs for `success`/`failure`

---

> [!NOTE]
> Adds backend infrastructure for Azure Entra ID authentication.
> 
> - New global `AzureModule` with `AzureAuthService` (PKCE OAuth flow, MSAL client, silent token acquisition for `redis.azure.com` and `management.azure.com`) and `AzureAuthController` exposing `GET /azure/auth/login`, `GET /azure/auth/callback`, `GET /azure/auth/status`, `POST /azure/auth/logout/:accountId`
> - Config constants for authority, client ID, scopes, and redirect URI; models for token/status responses
> - Integrated `AzureModule` into `AppModule`
> - Added `@azure/msal-node` dependency and updated `yarn.lock`
> - Unit tests for service covering state handling, callback, status, logout, and token retrieval paths
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1d58c79e9b43ae6b8fafdb1b91e82b2c2f5e1ff3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/api/src/modules/azure/auth/azure-auth.service.ts
+++ b/redisinsight/api/src/modules/azure/auth/azure-auth.service.ts
@@ -195,12 +195,35 @@ export class AzureAuthService {
     }
   }
 
+  /**
+   * Get an access token for Azure Cache for Redis.
+   *
+   * This token is used to authenticate directly with Azure Redis databases
+   * using Entra ID (Azure AD) authentication instead of access keys.
+   *
+   * @see https://learn.microsoft.com/en-us/azure/azure-cache-for-redis/cache-azure-active-directory-for-authentication
+   */
   async getRedisTokenByAccountId(
     accountId: string,
   ): Promise<AzureTokenResult | null> {
     return this.getTokenByAccountId(accountId, AZURE_REDIS_SCOPE);
   }
 
+  /**
+   * Get an access token for Azure Resource Manager (ARM) API.
+   *
+   * This token is used to call Azure Management APIs for autodiscovery:
+   * - List subscriptions the user has access to
+   * - List Azure Cache for Redis instances in each subscription
+   * - Get connection details (host, port, SSL settings)
+   *
+   * Note: Azure AD doesn't allow requesting tokens for multiple resources
+   * (redis.azure.com AND management.azure.com) in a single OAuth request.
+   * We request the Redis scope during login and acquire this scope silently
+   * when needed for autodiscovery.
+   *
+   * @see https://learn.microsoft.com/en-us/rest/api/redis/
+   */
   async getManagementTokenByAccountId(
     accountId: string,
   ): Promise<AzureTokenResult | null> {


redisinsight/api/src/modules/azure/constants.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * Azure AD authority URL for multi-tenant authentication.
+ * Uses 'common' endpoint to allow any Azure AD tenant.
+ */
+export const AZURE_AUTHORITY = 'https://login.microsoftonline.com/common';
+
+/**
+ * Azure App Registration Client ID.
+ */
+export const AZURE_CLIENT_ID = '61f3d82d-2bf3-432a-ba1b-c056e4cf0fd0';
```
valkirilov, line 10: I bet that nobody will try to change this, and build their own version using their own App Registration ID, however I still feel the need to write this note - I think this should be configurable via an environment variable 🤔 

- if we need to target different environments (local/staging/production), it might help us to have it easily configurable
- I'm not 100% sure, but we might need this for the enterprise builds as well, if they come with some specific requirements 
- and we already have the pattern in place to read from `config/default.ts`, so I think it shouldn't require huge changes

PS: Looking at the list, I think these are hood candidates for env variables:
- `AZURE_CLIENT_ID`
- `AZURE_OAUTH_REDIRECT_PATH`
- `AZURE_AUTHORITY` - although I'm not sure whether we'll need to change this one, I think it will be the best match for most cases

redisinsight/api/src/modules/azure/auth/azure-auth.service.ts
```diff
@@ -0,0 +1,254 @@
...
+  AZURE_AUTHORITY,
+  AZURE_CLIENT_ID,
+  AZURE_REDIS_SCOPE,
+  AZURE_MANAGEMENT_SCOPE,
+  AZURE_OAUTH_REDIRECT_PATH,
+  AZURE_OAUTH_SCOPES,
+  AzureAuthStatus,
+} from '../constants';
+
+export interface AzureTokenResult {
```
valkirilov: nit: I see on many places, we have a pattern to offload the interfaces in separate files:
- https://github.com/redis/RedisInsight/tree/main/redisinsight/api/src/modules/bulk-actions/interfaces
- https://github.com/redis/RedisInsight/blob/main/redisinsight/api/src/modules/statics-management/providers/auto-updated-statics.interface.ts
- https://github.com/redis/RedisInsight/tree/main/redisinsight/api/src/modules/cloud/common/models

PS: We can update the AI rules, to follow this pattern (if there is no such already).

redisinsight/api/src/modules/azure/auth/azure-auth.service.ts
```diff
@@ -0,0 +1,254 @@
...
+   */
+  async handleCallback(
+    code: string,
+    state: string,
+  ): Promise<{ status: AzureAuthStatus; account?: AccountInfo }> {
+    const pca = this.getMsalClient();
+
+    const verifier = this.authRequests.get(state);
+    if (!verifier) {
+      this.logger.warn(`No auth request found for state: ${state}`);
```
valkirilov, line 120: Question: I'm not sure what is the default pattern in the codebase (probably the agent based this decision on something), but why it's a warning, and not an error - what do you think?

redisinsight/api/src/modules/azure/auth/azure-auth.service.ts
```diff
@@ -0,0 +1,254 @@
...
+    }
+  }
+
+  async getRedisTokenByAccountId(
+    accountId: string,
+  ): Promise<AzureTokenResult | null> {
+    return this.getTokenByAccountId(accountId, AZURE_REDIS_SCOPE);
+  }
+
+  async getManagementTokenByAccountId(
```
valkirilov, line 227: Question: I read the comments in the code and that we have two different scopes (Redis + Management), but I'm not sure what is Azure Management, and why we need it. 

Is it something defined and required by Azure?

redisinsight/api/src/modules/azure/auth/azure-auth.controller.ts
```diff
@@ -0,0 +1,114 @@
...
+      return {
+        status: AzureAuthStatus.Failed,
+        error: 'Missing code or state parameter',
+      };
+    }
+
+    return this.azureAuthService.handleCallback(code, state);
+  }
+
+  @Get('status')
```
valkirilov, line 75: Question: What will be the purpose of this API endpoint on the front-end?

My assumption is the modal for "Adding an Azure Managed Database", where we let the user login/logout?

redisinsight/api/src/modules/azure/constants.ts
pawelangelow: That’s definitely a valid scenario, and we considered it. For now, though, we decided to proceed with the current setup (i.e., using only our app registration). I had similar questions myself, but in my opinion, any change here would require a new release anyway, so it makes more sense to keep this in the codebase. And most importantly, these aren’t secrets.

redisinsight/api/src/modules/azure/auth/azure-auth.service.ts
pawelangelow: This is not supposed to happen, so it's more of a debugging enhancement :D (I usually run the backend with >debug level logging, because it's too verbose). If you have a suggestion for it - shoot, I don't have a strong opinion on it.

redisinsight/api/src/modules/azure/auth/azure-auth.service.ts
pawelangelow: Azure Redis API (`https://redis.azure.com`) - Used to connect to and interact with Azure Cache for Redis databases directly (authentication, data operations)
Azure Management API (`https://management.azure.com`) - Used to manage Azure resources programmatically (list subscriptions, list Redis instances, get connection details, create/delete resources)

Why we need both:

- For the autodiscovery feature (future work), we need to:
  - List user's subscriptions → Management API
  - List Redis databases in those subscriptions → Management API
  - Get connection details (host, port, keys) → Management API
  - Actually connect to the database → Redis API token
  
Azure limitation:
Azure AD doesn't allow requesting tokens for multiple resources in a single OAuth request (AADSTS70011 error). So we:
- Request the Redis scope during initial login
- Acquire the Management scope silently later using acquireTokenSilent() when needed

Is it required by Azure?
Yes - these are Azure's defined resource scopes. Each Azure service has its own scope/resource identifier, and you need the appropriate token to call that service's API.

redisinsight/api/src/modules/azure/auth/azure-auth.controller.ts
pawelangelow: Yeah, basically, we'll need it inside the UI - yeah, for the modal, autodiscovery (to see which user is selected), and maybe somewhere else to showcase again, which is the user

pawelangelow: I did wonder whether to include this in the PR since it’s not really used yet, but IMO it’s easier this way - at least the changes are grouped by context. Still figuring out the right balance for PR splitting 🤔

redisinsight/api/src/modules/azure/constants.ts
valkirilov: Yes, they're not secrets, but are configurations. And it makes sense to have them on a single place, like we already have plenty of them in the config:

- https://github.com/redis/RedisInsight/blob/main/redisinsight/api/config/default.ts

Anyway, it's not critical at this point, and I believe Artem knows better the BE codebase, and will advice in case this is necessary 👍

redisinsight/api/src/modules/azure/auth/azure-auth.controller.ts
valkirilov: Yup, no problem to be here. I just wanted to be sure what's behind it. Thanks :)

redisinsight/api/src/modules/azure/constants.ts
pawelangelow: Well, again, maybe. I thought of `api/config/default.ts`, but they are more env-related. I guess we won't use a different config for Azure, so I placed them in a different place. [RDI](https://github.com/redis/RedisInsight/blob/main/redisinsight/api/src/modules/rdi/constants/index.ts) and a few other modules define module-level constants, and I liked it 😄

Issue #5412, opened 2026-01-28
[Bug]:Object contains forbidden constructor property


fukuaile:
When using RedisJSON, an ERRO occurs if any key contains the word "constructor". Using other words or other data types, such as HASH, works without issue. This problem did not occur when using the previous RESP.app.

<img width="827" height="765" alt="Image" src="https://github.com/user-attachments/assets/241d707f-effd-4104-a938-8f9d8ddefa7e" />

<img width="660" height="176" alt="Image" src="https://github.com/user-attachments/assets/4c3573e9-bb1d-4a4d-976f-d57737ad2a30" />

valkirilov:
Wat? 😀
Very nice catch, thanks for reporting it! 

It seems like it's a side effect of the dependencies we use to parse JSON — they come with some default protections that prevent parsing certain JavaScript terms like `constructor`, `__proto__`, etc. (to guard against prototype pollution attacks).

We'll take care of it and let you know once the fix is in place. Thanks again!

valkirilov:
We have released a new version of Redis Insight [v3.4.1](https://github.com/redis/RedisInsight/releases/tag/3.4.1), and the issue should be gone now. You can give it a try, and in case you find any other issues, feel free to reopen it and let us know 👍 

Thanks again for reporting it, and sorry for the delay 🙏

PR #5413, opened 2026-01-28
RI-7973 Enhance JSON parse to safely handle prototype pollution

valkirilov:
# What
 
Update JSON parser configuration to allow `constructor` and `__proto__` properties to be preserved during JSON parsing

_Note: Since RedisInsight displays arbitrary user data from Redis, is it safe to do so?_

Reference #5412

# Testing

1. Create a JSON key with such proprietary values
```
JSON.SET mykey $ '{"constructor":"example value"}'
```
2. Try to open the key on the Browser page

| Before | After |
| - | - |
image|image

---

> [!NOTE]
> Ensures JSON values with potentially dangerous property names are preserved when reading from RedisJSON.
> 
> - Configure `JSONBigInt` in `rejson-rl.service.ts` with `protoAction: 'preserve'` and `constructorAction: 'preserve'`
> - Add unit tests in `rejson-rl.service.spec.ts` verifying handling of `constructor`, `"__proto__"` (as raw string), and nested occurrences during `getJson`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 23ebeab3cdf0fa5961e54f2a48f902f78f7eca1d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/api/src/modules/browser/rejson-rl/rejson-rl.service.ts
+++ b/redisinsight/api/src/modules/browser/rejson-rl/rejson-rl.service.ts
@@ -35,7 +35,12 @@ import { RedisClient } from 'src/modules/redis/client';
 import { DatabaseService } from 'src/modules/database/database.service';
 
 const MODULES_CONFIG = config.get('modules') as Config['modules'];
-const JSONbig = JSONBigInt();
+
+// Allow JSON keys named 'constructor' and '__proto__' to be preserved since RedisInsight displays arbitrary user data from Redis
+const JSONbig = JSONBigInt({
+  protoAction: 'preserve',
+  constructorAction: 'preserve',
+});
 
 @Injectable()
 export class RejsonRlService {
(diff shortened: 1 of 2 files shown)


redisinsight/api/src/modules/browser/rejson-rl/rejson-rl.service.ts
pd-redis, line 1: We have had 2 CVEs in the last couple of months related to loose parsing of user data - React Server Components and N8N, both of them allowed arbitrary code execution because of interpreting strings (JSON) as javascript. Therefore, my opinion is that ANY parsing of JS keywords (such as constructor) should be avoided

valkirilov:
We're going to tackle this in another way. Instead of disabling the default precondition’s behavior of the parser, we'll simply communicate back with the users in a better way, that their JSON contains reserved keywords, and it's better for them to use the CLI/Workbench commands instead.

redisinsight/api/src/modules/browser/rejson-rl/rejson-rl.service.ts
valkirilov: Yup, closing this pull request, we'll tackle this one in another way 👍 https://github.com/redis/RedisInsight/pull/5413#issuecomment-3871263825

PR #5416, opened 2026-01-28
RI-000: add possibility to run multiple UI

ArtemHoruzhenko:
# What

Just tech changes. Added possibility to run UI on specified port via env variable. So we can run both UI and API from different folders at the same time when each will live on own ports
# Testing

---

> [!NOTE]
> Allows running multiple UI instances or customizing the dev port.
> 
> - Updates `vite.config.mjs` to set `server.port` from `RI_UI_DEV_PORT` with fallback to `8080`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e8ead74ed7f4fb176ce224e2c32f3b59c51ff4ec. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/ui/vite.config.mjs
+++ b/redisinsight/ui/vite.config.mjs
@@ -86,7 +86,7 @@ export default defineConfig({
     },
   },
   server: {
-    port: 8080,
+    port: parseInt(process.env.RI_UI_DEV_PORT, 10) || 8080,
     fs: {
       allow: ['..', '../../node_modules/monaco-editor', 'static', 'defaults'],
     },


redisinsight/ui/vite.config.mjs
```diff
@@ -86,7 +86,7 @@ export default defineConfig({
     },
   },
   server: {
-    port: 8080,
+    port: parseInt(process.env.RI_UI_DEV_PORT, 10) || 8080,
```
pd-redis, line 89: We should use helper function IMO, some less experienced users may try to start the app with ports less than 1000 and get errors

pd-redis: Also, isn't there an option to make it so that vite choses next port if current is occupied?

ArtemHoruzhenko: I don't think we should to automate this, imo it should be under developer control. Moreover since it is developer command - I don't think we need to even validate it.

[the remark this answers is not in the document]
ArtemHoruzhenko: 0 port? seriously?

KrumTy, line 89: shouldn't `RI_UI_DEV_PORT` be added in some env.example file at least?

pd-redis: It is automatic out of the box - https://vite.dev/config/server-options#server-port
But I guess having this option does not harm if someone wants to use it

ArtemHoruzhenko: we don't have some env.example file atm. we can add it if needed for needed envs.

pawelangelow:
is this still alive 😄

PR #5421, opened 2026-01-30
RI-7900 Index Details component

valkirilov:
# What

Introduce Index Details component
    
- list index fields, values, and their types
- comes with readonly/editable mode to select certain rows or edit the types

_PS: You can review the Storybook examples for more details and usage instructions._

```
yarn sb
```

http://localhost:6006/?path=/docs/pages-vectorsearch-indexdetails--docs

| Readonly mode | Editable mode |
| - | - |
image|image

# Testing

## Basic Readonly Mode

Display fields without any interactivity:

```javascript
const fields: IndexField[] = [
  { id: '1', name: 'title', value: 'Introduction to Redis', type: FieldTypes.TEXT },
  { id: '2', name: 'price', value: 29.99, type: FieldTypes.NUMERIC },
  { id: '3', name: 'category', value: 'books', type: FieldTypes.TAG },
]

<IndexDetails 
  fields={fields} 
  mode={IndexDetailsMode.Readonly} 
/>
```

| Readonly mode | Complex Tooltips |
| - | - |
image|image

## Editable Mode with Selection

Enable row selection and edit actions:

```javascript
const [rowSelection, setRowSelection] = useState<RowSelectionState>({})

  const handleFieldEdit = (field: IndexField) => {
    // Open edit modal for the field
    console.log('Edit field:', field)
  }

  // Get selected fields
  const selectedFields = fields.filter((field) => rowSelection[field.id])

  return (
    <IndexDetails
      fields={fields}
      mode={IndexDetailsMode.Editable}
      rowSelection={rowSelection}
      onRowSelectionChange={setRowSelection}
      onFieldEdit={handleFieldEdit}
    />
  )
```

| Editable mode | Edit field callback |
| - | - |
image|image

---

> [!NOTE]
> **Low Risk**
> Low risk UI-only addition with Storybook + unit tests; the only shared change is a backwards-compatible `FieldTag` prop signature update to support custom `data-testid` values.
> 
> **Overview**
> Adds a new `IndexDetails` component under `pages/vector-search` that renders index fields in a table, supporting **readonly** display and an **editable** mode with multi-row selection and per-row “edit type” actions.
> 
> Introduces supporting column/cell components and header tooltips, Storybook stories, unit tests, and a `fishery`/`faker` `indexFieldFactory` for generating sample fields. Also updates shared `FieldTag` to accept an optional `dataTestId` (with a default) and moves its props to a new `FieldTag.types.ts`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f66b9474c5e03eb507d1d28585d2d52daa9a85d3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.config.tsx
+++ b/redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.config.tsx
@@ -1,100 +1,38 @@
-import React, { useMemo } from 'react'
-import { ColumnDef, Row, Table } from 'uiSrc/components/base/layout/table'
+import { ColumnDef } from 'uiSrc/components/base/layout/table'
 import {
-  IndexDetailsColumn,
+  GetColumnsOptions,
   IndexDetailsMode,
   IndexField,
 } from './IndexDetails.types'
+import {
+  SELECTION_COLUMN,
+  NAME_COLUMN,
+  VALUE_COLUMN,
+  TYPE_COLUMN_READONLY,
+  TYPE_COLUMN_EDITABLE,
+  createActionsColumn,
+} from './IndexDetails.columns'
 
-import { FieldNameCell } from './components/FieldNameCell/FieldNameCell'
-import { FieldValueCell } from './components/FieldValueCell/FieldValueCell'
-import { FieldTypeCell } from './components/FieldTypeCell/FieldTypeCell'
-import { FieldActionsCell } from './components/FieldActionsCell/FieldActionsCell'
-import { ColumnHeader } from './components/ColumnHeader/ColumnHeader'
-import { FieldNameTooltip } from './components/FieldNameCell/FieldNameTooltip'
-import { FieldValueTooltip } from './components/FieldValueCell/FieldValueTooltip'
-import { FieldTypeTooltip } from './components/FieldTypeCell/FieldTypeTooltip'
+const READONLY_COLUMNS: ColumnDef<IndexField>[] = [
+  NAME_COLUMN,
+  VALUE_COLUMN,
+  TYPE_COLUMN_READONLY,
+]
 
-type ColumnConfig = ColumnDef<IndexField> & {
-  modes: IndexDetailsMode[]
-}
+const EDITABLE_BASE_COLUMNS: ColumnDef<IndexField>[] = [
+  SELECTION_COLUMN,
+  NAME_COLUMN,
+  VALUE_COLUMN,
+  TYPE_COLUMN_EDITABLE,
+]
 
-interface UseColumnsOptions {
-  mode: IndexDetailsMode
-  onFieldEdit?: (field: IndexField) => void
-}
-
-export const useIndexDetailsColumns = ({
+export const getIndexDetailsColumns = ({
   mode,
   onFieldEdit,
-}: UseColumnsOptions): ColumnDef<IndexField>[] => {
-  const isEditable = mode === IndexDetailsMode.Editable
+}: GetColumnsOptions): ColumnDef<IndexField>[] => {
+  if (mode === IndexDetailsMode.Readonly) {
+    return READONLY_COLUMNS
+  }
 
-  const rowSelectionColumn = Table.useRowSelectionColumn<IndexField>()
-
-  return useMemo(() => {
-    const columns: ColumnConfig[] = [
-      {
-        ...rowSelectionColumn,
-        id: IndexDetailsColumn.Selection,
-        modes: [IndexDetailsMode.Editable],
-      },
-      {
-        id: IndexDetailsColumn.Name,
-        modes: [IndexDetailsMode.Editable, IndexDetailsMode.Readonly],
-        accessorKey: IndexDetailsColumn.Name,
-        enableSorting: false,
-        header: () => (
-          <ColumnHeader label="Field name" tooltip={<FieldNameTooltip />} />
-        ),
-        cell: ({ row }: { row: Row<IndexField> }) => (
-          <FieldNameCell field={row.original} />
-        ),
-      },
-      {
-        id: IndexDetailsColumn.Value,
-        modes: [IndexDetailsMode.Editable, IndexDetailsMode.Readonly],
-        accessorKey: IndexDetailsColumn.Value,
-        enableSorting: false,
-        header: () => (
-          <ColumnHeader
-            label="Field sample value"
-            tooltip={<FieldValueTooltip />}
-          />
-        ),
-        cell: ({ row }: { row: Row<IndexField> }) => (
-          <FieldValueCell field={row.original} />
-        ),
-      },
-      {
-        id: IndexDetailsColumn.Type,
-        modes: [IndexDetailsMode.Editable, IndexDetailsMode.Readonly],
-        accessorKey: IndexDetailsColumn.Type,
-        enableSorting: false,
-        header: () => (
-          <ColumnHeader
-            label={isEditable ? 'Suggested indexing type' : 'Indexing type'}
-            tooltip={<FieldTypeTooltip />}
-          />
-        ),
-        cell: ({ row }: { row: Row<IndexField> }) => (
-          <FieldTypeCell field={row.original} />
-        ),
-      },
-      {
-        id: IndexDetailsColumn.Actions,
-        modes: [IndexDetailsMode.Editable],
-        enableSorting: false,
-        size: 40,
-        header: '',
-        cell: ({ row }: { row: Row<IndexField> }) => (
-          <FieldActionsCell field={row.original} onEdit={onFieldEdit} />
-        ),
-      },
-    ]
-
-    return columns
-      .filter((col) => col.modes.includes(mode))
-      .map(({ modes: _, ...columnDef }) => columnDef)
-  }, [mode, isEditable, onFieldEdit, rowSelectionColumn])
+  return [...EDITABLE_BASE_COLUMNS, createActionsColumn(onFieldEdit)]
 }
--- a/redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.stories.tsx
+++ b/redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.stories.tsx
@@ -21,21 +21,21 @@ const SAMPLE_FIELDS: IndexField[] = [
   indexFieldFactory.build({ id: '5', type: FieldTypes.VECTOR }),
 ]
 
+const handleFieldEdit = (field: IndexField) => {
+  // eslint-disable-next-line no-alert
+  alert(
+    `Edit field: "${field.name}"\n\n` +
+      `Current type: ${field.type.toUpperCase()}\n` +
+      `Sample value: ${field.value}\n\n` +
+      `In the real flow, a modal will appear here to edit the field type and settings.`,
+  )
+}
+
 // Stateful wrapper for interactive stories (controlled component)
 const IndexDetailsWrapper = (props: IndexDetailsProps) => {
   const [rowSelection, setRowSelection] = useState<RowSelectionState>(
     props.rowSelection || {},
   )
-
-  const handleFieldEdit = (field: IndexField) => {
-    // eslint-disable-next-line no-alert
-    alert(
-      `Edit field: "${field.name}"\n\n` +
-        `Current type: ${field.type.toUpperCase()}\n` +
-        `Sample value: ${field.value}\n\n` +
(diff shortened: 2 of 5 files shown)


redisinsight/ui/src/pages/vector-search/components/command-view/CommandView.stories.tsx
```diff
@@ -33,6 +33,7 @@ const withMonacoSetup = (Story: React.ComponentType) => {
 
 const meta: Meta<typeof CommandView> = {
   component: CommandView,
+  title: 'Pages/VectorSearch/CommandView',
```
pd-redis: Adding a title will change the nav structure in SB, did you add it on purpose, or it was pushed by the llm?

redisinsight/ui/src/pages/vector-search/components/index-details/components/FieldActionsCell/FieldActionsCell.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { EditIcon } from 'uiSrc/components/base/icons'
+import { IconButton } from 'uiSrc/components/base/forms/buttons/IconButton'
+import { RiTooltip } from 'uiSrc/components/base/tooltip'
+import { FieldActionsCellProps } from './FieldActionsCell.types'
+
+export const FieldActionsCell = ({ field, onEdit }: FieldActionsCellProps) => {
+  const handleClick = (e: React.MouseEvent) => {
+    e.stopPropagation()
```
pd-redis, line 10: maybe call preventDefault as well, just for completeness

redisinsight/ui/src/pages/vector-search/components/index-details/components/FieldTypeCell/FieldTypeCell.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import { FieldTag } from 'uiSrc/components/new-index/create-index-step/field-box/FieldTag'
+import { FieldTypeCellProps } from './FieldTypeCell.types'
+
+export const FieldTypeCell = ({ field }: FieldTypeCellProps) => (
+  <span data-testid={`index-details-field-type-${field.id}`}>
```
pd-redis: why does this need to be wrapped in span?

redisinsight/ui/src/pages/vector-search/components/index-details/components/FieldTypeCell/FieldTypeTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+    </Text>
+
+    <Text color="secondary">
+      Defines how Redis searches this field and how it behaves at query time.
+      Available indexing types:
+    </Text>
+
+    <Spacer size="s" />
+    {FIELD_TYPE_DESCRIPTIONS.map(({ type, description }) => (
+      <Row key={type} gap="s" style={{ whiteSpace: 'nowrap' }}>
```
pd-redis: If this is really required, lets make it as styled component

redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.config.tsx
```diff
@@ -0,0 +1,101 @@
+import React, { useMemo } from 'react'
+import { Row } from '@tanstack/react-table'
```
pd-redis: You can import this from our table

redisinsight/ui/src/pages/vector-search/components/index-details/components/FieldTypeCell/FieldTypeTooltip.tsx
valkirilov: Yeah, it was intentional, because of a minor visual glitch, I didn't want to introduce a styled component only for this one liner, but you're correct, let's try to be consistent. Fixed in https://github.com/redis/RedisInsight/pull/5421/changes/2ed221263cd8f6c327cbac9c04e99ffec06c9c89

| Bad | Good |
| - | - |
image|image

redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.config.tsx
valkirilov: My bad, fixed in https://github.com/redis/RedisInsight/pull/5421/changes/41521ee3f859bbb052b8ae3e7e8ed0e68ac637a1

redisinsight/ui/src/pages/vector-search/components/index-details/components/FieldActionsCell/FieldActionsCell.tsx
valkirilov: Agree, added in https://github.com/redis/RedisInsight/pull/5421/changes/41521ee3f859bbb052b8ae3e7e8ed0e68ac637a1

redisinsight/ui/src/pages/vector-search/components/command-view/CommandView.stories.tsx
valkirilov: Yes, it was a decision made by the LLM, I have removed it in https://github.com/redis/RedisInsight/pull/5421/changes/3b0aa2547adb3aa2289b091176c48cd52d208129

redisinsight/ui/src/pages/vector-search/components/index-details/components/FieldTypeCell/FieldTypeCell.tsx
valkirilov: The span is present only to let us pass the test id, like we do for the rest of the columns. They use the `Text` component, which supports out-of-the-box `data-testid`, but here we rely on the `FieldTag`, which doesn’t. 

Now it's fixed in https://github.com/redis/RedisInsight/pull/5421/changes/2efaaa4caf3fa75e2b7247ba4a3adc53204b1f0f

redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const IndexDetails = ({
+  fields,
+  mode = IndexDetailsMode.Editable,
+  showBorder = false,
+  rowSelection,
+  onRowSelectionChange,
+  onFieldEdit,
+}: IndexDetailsProps) => {
+  const isEditable = mode === IndexDetailsMode.Editable
```
KrumTy, line 19: why not a boolean directly? So many checks and converting enum to boolean otherwise. `isEditable` prop (affects multiple files)

redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const selectedFields = props.fields.filter((field) => rowSelection[field.id])
+  const isEditable = props.mode === IndexDetailsMode.Editable
+
+  return (
+    <Col>
+      <IndexDetails
+        {...props}
+        rowSelection={rowSelection}
+        onRowSelectionChange={setRowSelection}
+        onFieldEdit={handleFieldEdit}
```
KrumTy, line 49: renders the useMemo hook in useIndexDetailsColumns obsolete 

hint: handleFieldEdit can be a constant, move it to the hook as it's not used anywhere else

redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.config.tsx
```diff
@@ -0,0 +1,100 @@
...
+  onFieldEdit?: (field: IndexField) => void
+}
+
+export const useIndexDetailsColumns = ({
+  mode,
+  onFieldEdit,
+}: UseColumnsOptions): ColumnDef<IndexField>[] => {
+  const isEditable = mode === IndexDetailsMode.Editable
+
+  const rowSelectionColumn = Table.useRowSelectionColumn<IndexField>()
```
KrumTy: does that return a memoized object?
I think you can have this instead https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/home/components/databases-list/DatabasesList.config.tsx#L51-L55

redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.config.tsx
```diff
@@ -0,0 +1,100 @@
...
+        cell: ({ row }: { row: Row<IndexField> }) => (
+          <FieldActionsCell field={row.original} onEdit={onFieldEdit} />
+        ),
+      },
+    ]
+
+    return columns
+      .filter((col) => col.modes.includes(mode))
+      .map(({ modes: _, ...columnDef }) => columnDef)
+  }, [mode, isEditable, onFieldEdit, rowSelectionColumn])
```
KrumTy: I think it's worth having two static column objects over this. Or you can literally have the same thing in terms of operations and still produce the 2 static objects. Then decide which to use based on `mode` 

if the two comments I wrote for this file applied (plus one for the onFieldEdit I wrote elsewhere) , you'd be able to drop this hook and work with static objects (way simpler).

redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.config.tsx
valkirilov: This is how Redis UI does it in their docs, but I didn't check how it works under the hood.
- https://github.com/redislabsdev/redis-ui/blob/master/docs/stories/Table/RowSelection.stories.tsx#L34

I'll compare it with your suggestion, thanks 🙂

redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.tsx
valkirilov: Having the enum makes it easier to extend it further in the future. Yes, for now we have two options (read and edit mode), so a boolean will be enough, but I believe this flexibility provided by the enum is not a debt, but something we can extend easily in the future.

And `isEditable ` is a syntax sugar, to keep the rest of the code more readable (by humans), but I don't think it adds any overhead overall. So I'm fine with it 🙂

KrumTy: what other modes in future do you foresee that would not benefit from explicit editable and non editable mode?

valkirilov: It could be anything, so let's be creative here:
- change the flow depending on whether the input is a Hash or JSON
- or allow some custom behavior based on external context

I might be missing something though - does this kind of flexible pattern raise concerns for you?

KrumTy: it's just weird api imo, that's all
mode=editable and mode=hashkey - these are different concepts entirely, I see no point of hash/json types being deprived of a edit/readonly functionality, or even worse - having combinations of if/elses. Or mode=hashkey_editable, which gets you back to the boolean field's superiority

redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.stories.tsx
valkirilov: It's just a Storybook, but since it can be used as an example later, it makes sense to have optimized code here as well. Thanks for pointing it out - I made the suggested changes in https://github.com/redis/RedisInsight/pull/5421/changes/f66b9474c5e03eb507d1d28585d2d52daa9a85d3

redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.config.tsx
valkirilov: I didn't dig deeply into the Redis UI implementation, because I went for the simplified approach, so we can safely drop the hook and go for static column definitions via objects. Changes in https://github.com/redis/RedisInsight/pull/5421/changes/f66b9474c5e03eb507d1d28585d2d52daa9a85d3

redisinsight/ui/src/pages/vector-search/components/index-details/IndexDetails.config.tsx
valkirilov: Suggested changes applied in https://github.com/redis/RedisInsight/pull/5421/changes/f66b9474c5e03eb507d1d28585d2d52daa9a85d3 . Thanks 🙂

PR #5423, opened 2026-01-30
RI-7948: init e2e playwright project

KrumTy:
# What

Sets up the foundational E2E test infrastructure using Playwright with support for both Chromium browser and Electron desktop testing.

## Key additions:

- AI guidelines: [rules/e2e-testing.md](https://github.com/redis/RedisInsight/blob/fbbba493f2b63775635f2317641d4aac3a02d693/.ai/rules/e2e-testing.md), [commands/e2e-fix.md](https://github.com/redis/RedisInsight/blob/fbbba493f2b63775635f2317641d4aac3a02d693/.ai/commands/e2e-fix.md), [e2e-generate.md](https://github.com/redis/RedisInsight/blob/fbbba493f2b63775635f2317641d4aac3a02d693/.ai/commands/e2e-generate.md)
- VSCode extension recommendations (Playwright) and added Playwright MCP server
- Project-based test organization (chromium and electron projects) with separate configurations
- Unified fixture architecture that handles both browser and Electron modes
- API helper for fast test setup/teardown via REST API
- Page Object Model with BasePage providing common navigation methods
- Environment-based configuration (.env, .env.ci, .env.staging)
- Setup/teardown projects for global initialization per test mode
- two sample groups of tests (browser -> add keys and key filtering)

## Test infrastructure:

- Worker-scoped Electron app fixture (single app instance per worker) - requires path to build
- Automatic page reload in Electron mode to sync with  beforeAll-created data
- Browser mode navigates to baseURL on test start

# Testing

Please refer to the [README.md](https://github.com/redis/RedisInsight/blob/712508f03050cd1d228527e9e364edffdf2025c7/tests/e2e-playwright/README.md#prerequisites) for local setup and running

---

> [!NOTE]
> **Medium Risk**
> Mostly new test tooling, but it adds Husky pre-commit hooks and new Playwright/Electron fixture orchestration that can impact developer workflows and CI reliability.
> 
> **Overview**
> Adds a new standalone E2E test package under `tests/e2e-playwright/`, including Playwright project config for `chromium` and `electron` runs (with setup/teardown projects), shared fixtures that can launch an Electron app per worker, and an `ApiHelper` for fast REST-based setup/teardown.
> 
> Introduces initial Page Object Model scaffolding (e.g., `BasePage`, `InstancePage`, `BrowserPage` and Browser components) plus sample Browser specs for adding keys and key-filtering patterns, and updates `TEST_PLAN.md` to mark those cases as implemented.
> 
> Adds supporting developer/AI tooling: `.ai` E2E rules + commands, Playwright MCP + VSCode extension recommendation, and E2E-local lint/format config with Husky/lint-staged pre-commit checks; also updates `.eslintignore` to exclude `tests/e2e-playwright`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b3506437e858835375fcab77a9ff247436cb0860. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.ai/rules/e2e-testing.md
+++ b/.ai/rules/e2e-testing.md
@@ -24,7 +24,9 @@ tests/e2e-playwright/
 ├── helpers/             # API helpers for setup/teardown
 ├── pages/               # Page Object Models
 │   ├── BasePage.ts      # Base class for all pages
-│   └── {feature}/       # Feature-specific pages
+│   ├── InstancePage.ts  # Base class for database instance pages
+│   ├── components/      # Shared components (InstanceHeader, NavigationTabs, BottomPanel)
+│   └── {feature}/       # Feature-specific pages (browser/, cli/, etc.)
 ├── test-data/           # Test data factories
 ├── tests/               # Test specs organized by project
 │   ├── main/            # Default parallel tests
@@ -80,26 +82,48 @@ Create a new project folder when tests:
 
 ## Page Objects
 
-### Always Extend BasePage
+### Page Object Hierarchy
+
+```
+BasePage (abstract)
+  ├── DatabasesPage           # Databases list page
+  ├── SettingsPage            # Settings page
+  └── InstancePage (abstract) # Base for all database instance pages
+        ├── instanceHeader    # Database name, stats, breadcrumb
+        ├── navigationTabs    # Browse, Workbench, Analyze, Pub/Sub
+        ├── bottomPanel       # CLI, Command Helper, Profiler
+        └── BrowserPage       # Browser-specific (extends InstancePage)
+              └── WorkbenchPage (future)
+              └── AnalyzePage (future)
+              └── PubSubPage (future)
+```
+
+### Extend the Appropriate Base Class
+
+- **BasePage** - For standalone pages (DatabasesPage, SettingsPage)
+- **InstancePage** - For pages within a connected database (BrowserPage, WorkbenchPage, etc.)
 
 Page objects are **stateless** - they don't store database objects. Pass `databaseId` to navigation methods.
 
 ```typescript
+// For database instance pages - extend InstancePage
 import { Page, Locator } from '@playwright/test';
-import { BasePage } from '../BasePage';
+import { InstancePage } from '../InstancePage';
 
-export class MyPage extends BasePage {
-  readonly myButton: Locator;
+export class WorkbenchPage extends InstancePage {
+  readonly editor: Locator;
 
   constructor(page: Page) {
     super(page);
-    this.myButton = page.getByTestId('my-button');
+    this.editor = page.getByTestId('workbench-editor');
   }
 
-  // Pass databaseId as parameter - don't store it in the class
+  // InstancePage provides: instanceHeader, navigationTabs, bottomPanel
+  // Plus navigation methods: navigateToBrowser(), openCli(), etc.
+
   async goto(databaseId: string): Promise<void> {
-    await this.gotoHome();
-    await this.connectToDatabase(databaseId);
+    await this.gotoDatabase(databaseId);
+    await this.navigationTabs.gotoWorkbench();
     await this.waitForLoad();
   }
 }
@@ -111,7 +135,7 @@ Break large pages into components:
 
 ```typescript
 // pages/feature/FeaturePage.ts
-export class FeaturePage extends BasePage {
+export class FeaturePage extends InstancePage {
   readonly dialog: FeatureDialog;
   readonly list: FeatureList;
 
--- a/tests/e2e-playwright/README.md
+++ b/tests/e2e-playwright/README.md
@@ -198,13 +198,33 @@ tests/e2e-playwright/
 
 Page objects are organized into component-based POMs for better maintainability:
 
-- **Page-level POMs** (`DatabasesPage`) - High-level page actions
-- **Component POMs** (`AddDatabaseDialog`, `DatabaseList`) - Reusable UI components
+```
+BasePage (abstract)
+  ├── DatabasesPage           # Databases list page
+  ├── SettingsPage            # Settings page
+  └── InstancePage (abstract) # Base for all database instance pages
+        ├── instanceHeader    # Database name, stats, breadcrumb
+        ├── navigationTabs    # Browse, Workbench, Analyze, Pub/Sub
+        ├── bottomPanel       # CLI, Command Helper, Profiler
+        └── BrowserPage       # Browser-specific (extends InstancePage)
+              └── WorkbenchPage (future)
+              └── AnalyzePage (future)
+              └── PubSubPage (future)
+```
+
+- **BasePage** - Common navigation methods for all pages
+- **InstancePage** - Base class for pages within a connected database (provides shared header, tabs, bottom panel)
+- **Component POMs** (`AddDatabaseDialog`, `KeyList`) - Reusable UI components
 
 ```typescript
 // Access component POMs through the page
 await databasesPage.addDatabaseDialog.fillForm(config);
-await databasesPage.databaseList.getRow(name);
+await browserPage.keyList.selectKey(keyName);
+
+// InstancePage provides common components
+await browserPage.instanceHeader.getDatabaseName();
+await browserPage.navigationTabs.gotoWorkbench();
+await browserPage.bottomPanel.openCli();
 ```
 
 ### Test Structure
--- a/tests/e2e-playwright/pages/browser/BrowserPage.ts
+++ b/tests/e2e-playwright/pages/browser/BrowserPage.ts
@@ -1,80 +1,45 @@
 import { Page, Locator, expect } from '@playwright/test';
-import { BasePage } from '../BasePage';
+import { InstancePage } from '../InstancePage';
 import { AddKeyDialog, BulkActionsPanel, KeyDetails, KeyList } from './components';
 
 /**
  * Browser Page Object
  * Main page for browsing Redis keys
+ *
+ * Extends InstancePage which provides:
+ * - instanceHeader: Database name, stats, breadcrumb
+ * - navigationTabs: Browse, Workbench, Analyze, Pub/Sub tabs
+ * - bottomPanel: CLI, Command Helper, Profiler buttons
  */
-export class BrowserPage extends BasePage {
-  // Components
+export class BrowserPage extends InstancePage {
+  // Browser-specific components
   readonly addKeyDialog: AddKeyDialog;
   readonly bulkActionsPanel: BulkActionsPanel;
   readonly keyDetails: KeyDetails;
   readonly keyList: KeyList;
 
-  // Navigation tabs
-  readonly browseTab: Locator;
-  readonly workbenchTab: Locator;
-  readonly analyzeTab: Locator;
-  readonly pubSubTab: Locator;
-
-  // Action buttons
+  // Browser-specific action buttons
   readonly addKeyButton: Locator;
   readonly bulkActionsButton: Locator;
 
-  // Database info
-  readonly databaseName: Locator;
-  readonly databaseSelector: Locator;
-  readonly logicalDatabaseButton: Locator;
-  readonly cpuUsage: Locator;
-  readonly commandsPerSec: Locator;
-  readonly totalMemory: Locator;
-  readonly totalKeys: Locator;
-
-  // Bottom panel
-  readonly cliButton: Locator;
-  readonly commandHelperButton: Locator;
-  readonly profilerButton: Locator;
-
-  // Key details panel
+  // Browser-specific key details panel
   readonly keyDetailsPanel: Locator;
   readonly noKeySelectedMessage: Locator;
 
   constructor(page: Page) {
     super(page);
 
-    // Initialize components
+    // Initialize browser-specific components
     this.addKeyDialog = new AddKeyDialog(page);
     this.bulkActionsPanel = new BulkActionsPanel(page);
     this.keyDetails = new KeyDetails(page);
     this.keyList = new KeyList(page);
 
-    // Navigation tabs
-    this.browseTab = page.getByRole('tab', { name: 'Browse' });
-    this.workbenchTab = page.getByRole('tab', { name: 'Workbench' });
-    this.analyzeTab = page.getByRole('tab', { name: 'Analyze' });
-    this.pubSubTab = page.getByRole('tab', { name: 'Pub/Sub' });
-
-    // Action buttons
+    // Browser-specific action buttons
     this.addKeyButton = page.getByText('Add key');
     this.bulkActionsButton = page.getByRole('button', { name: /bulk actions/i });
 
-    // Database info
-    this.databaseName = page.locator('[data-testid="db-name"], [data-testid="database-name"]');
-    this.databaseSelector = page.getByRole('button', { name: /db0/i });
-    this.logicalDatabaseButton = page.getByRole('button', { name: /^db\d+$/i });
-    this.cpuUsage = page.locator('[data-testid="cpu-usage"]');
-    this.commandsPerSec = page.locator('[data-testid="commands-per-sec"]');
-    this.totalMemory = page.locator('[data-testid="total-memory"]');
-    this.totalKeys = page.locator('[data-testid="total-keys"]');
-
-    // Bottom panel buttons
-    this.cliButton = page.getByText('CLI').locator('..');
-    this.commandHelperButton = page.getByText('Command Helper').locator('..');
-    this.profilerButton = page.getByText('Profiler').locator('..');
-
-    // Key details
+    // Browser-specific key details panel
     this.keyDetailsPanel = page.locator('[data-testid="key-details"]');
     this.noKeySelectedMessage = page.getByText(/Select the key from the list/);
   }
@@ -104,24 +69,6 @@ export class BrowserPage extends BasePage {
     }
   }
 
-  async navigateToWorkbench(): Promise<void> {
-    await this.workbenchTab.click();
-    await this.page.waitForURL(/workbench/);
-  }
-
-  async navigateToAnalyze(): Promise<void> {
-    await this.analyzeTab.click();
-  }
-
-  async navigateToPubSub(): Promise<void> {
-    await this.pubSubTab.click();
-    await this.page.waitForURL(/pub-sub/);
-  }
-
-  async openCli(): Promise<void> {
-    await this.cliButton.click();
-  }
-
   async expectKeyInList(keyName: string): Promise<void> {
     const exists = await this.keyList.keyExists(keyName);
     expect(exists).toBe(true);
@@ -130,30 +77,6 @@ export class BrowserPage extends BasePage {
   async expectKeyNotInList(keyName: string): Promise<void> {
     const exists = await this.keyList.keyExists(keyName);
     expect(exists).toBe(false);
-  }
-
-  /**
-   * Get the logical database index displayed in the header
-   * @returns The logical database index (e.g., "db0", "db1") or null if not visible
-   */
-  async getLogicalDatabaseIndex(): Promise<string | null> {
-    try {
-      const button = this.logicalDatabaseButton;
-      if (await button.isVisible()) {
-        const text = await button.textContent();
-        return text?.trim() || null;
-      }
-      return null;
-    } catch {
-      return null;
-    }
-  }
-
-  /**
-   * Check if the logical database button is visible in the header
-   */
-  async isLogicalDatabaseButtonVisible(): Promise<boolean> {
-    return this.logicalDatabaseButton.isVisible();
   }
 }
 
--- a/tests/e2e-playwright/pages/InstancePage.ts
+++ b/tests/e2e-playwright/pages/InstancePage.ts
@@ -1,0 +1,117 @@
+import { Page } from '@playwright/test';
+import { BasePage } from './BasePage';
+import { InstanceHeader, NavigationTabs, BottomPanel } from './components';
+
+/**
+ * Base class for all database instance pages (Browser, Workbench, Analyze, Pub/Sub)
+ *
+ * Provides common components:
+ * - Instance header (database name, stats, breadcrumb)
+ * - Navigation tabs (Browse, Workbench, Analyze, Pub/Sub)
+ * - Bottom panel (CLI, Command Helper, Profiler)
+ *
+ * Specific pages (BrowserPage, WorkbenchPage, etc.) should extend this class.
+ */
+export abstract class InstancePage extends BasePage {
+  // Common components for all instance pages
+  readonly instanceHeader: InstanceHeader;
+  readonly navigationTabs: NavigationTabs;
+  readonly bottomPanel: BottomPanel;
+
+  constructor(page: Page) {
+    super(page);
+
+    // Initialize common components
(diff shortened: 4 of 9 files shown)


.ai/rules/e2e-testing.md
```diff
@@ -0,0 +1,471 @@
...
+
+## Location
+
+All E2E tests are in `tests/e2e-playwright/`. This is a **standalone package** - no imports from `redisinsight/ui/` or `redisinsight/api/`.
+
+## Test Plan
+
+**Always refer to `tests/e2e-playwright/TEST_PLAN.md`** for:
+- Test coverage status (✅ implemented, 🔲 not implemented)
+- Test priorities (🔴 critical, 🟠 smoke, 🟢 regression)
```
valkirilov: nit: I think we dropped these from the test plan.

.ai/rules/e2e-testing.md
```diff
@@ -0,0 +1,471 @@
...
+## Test Data
+
+### Use Factories with Faker
+
+```typescript
+import { faker } from '@faker-js/faker';
+
+export const TEST_PREFIX = 'test-';
+
+export function generateConfig(overrides?: Partial<Config>): Config {
```
valkirilov: Tip: We can replicate the setup from the front-end and combine [faker](https://fakerjs.dev/) with [fishery](https://github.com/thoughtbot/fishery), to bootstrap the factory creation. It doesn't add an additional overhead, but it's a nice add-on with handy helpers 🙂

.ai/rules/e2e-testing.md
```diff
@@ -0,0 +1,471 @@
...
+
+1. **Navigate to the page**: `browser_navigate_Playwright` to target URL
+2. **Take snapshot**: `browser_snapshot_Playwright` to see element tree
+3. **Interact with elements**: `browser_click_Playwright` to trigger dialogs, dropdowns, etc.
+4. **Wait for async content**: `browser_wait_for_Playwright` for dynamic content
+5. **Document findings**: Add discovered UI patterns to `TEST_PLAN.md` under the feature section
+
+### What to Look For
+
+- `data-testid` attributes → use with `page.getByTestId()`
```
valkirilov, line 329: nit: I'm not sure whether it will affect the LLM, but in all places the bullet for the test-ids is the first one in the list, but we agreed on having this as the last resort to select an item. We should put test IDs for the containers, but we should aim to select by text/role, to mimic the actual way users interact with our UI. 
- https://playwright.dev/docs/locators
- https://testing-library.com/docs/queries/about/#priority

So, having this in mind, I'm not sure whether it makes sense to reorder the bullets, so it's not the first option in the list.

.ai/rules/e2e-testing.md
```diff
@@ -0,0 +1,471 @@
...
+- Use fixed timeouts (`waitForTimeout`)
+- Use CSS selectors for dynamic content
+- Leave test data after tests
+- Import from `redisinsight/ui/` or `redisinsight/api/`
+- Hardcode test data (use faker)
+- Assume element structure without verification
+
+## Navigation (IMPORTANT)
+
+**All navigation must use UI-based methods, NOT URL navigation.**
```
valkirilov, line 378: Note: Ahh, I see your point (also, the note below, regarding the Electron build). It's sad, because such navigation might be handy if we have multiple tests targeting a very complex use case: 
- every time we should select a database
- navigate to the page
- and so on

And all this to happen through UI, and test stuff that is not actually related to the current test suite/case, which is more expensive and time-consuming (compared to direct URL navigation, when you know what you actually target). 

But yeah, I understand the intent and (why) it will no longer be possible 🤷‍♂️

Still, having having utils to navigate through the UI keep the code clear and maintainable 🙂

.ai/rules/e2e-testing.md
```diff
@@ -0,0 +1,471 @@
...
+
+  // Tests can run in parallel - they share the database but don't modify shared state
+});
+```
+
+### 2. Use Serial Only When Tests Truly Depend on Each Other
+
+```typescript
+// Only use .serial when tests modify state that subsequent tests depend on
+test.describe.serial('Workflow that modifies state', () => {
```
valkirilov, line 460: Note: I would prefer removing this rule from the list. I don't say we won't use it as a last resort whenever we need it. But I think it's better not to put it in the LLM context as an easy option. 

> Prompt engineering best practices focus on being specific, providing clear context, examples, and data, defining the desired output, and **giving instructions on what to do rather than what to avoid**.

Reference: https://www.digitalocean.com/resources/articles/prompt-engineering-best-practices

Let's keep it for us to suggest this fix, only whenever we need it 🙂

tests/e2e-playwright/.husky/pre-commit
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env sh
```
valkirilov, line 1: nit: I think we discussed these pre-commit hooks and we decided to rely on the CI for now, but anyway, it's minor 🙂

tests/e2e-playwright/config/databases/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  standaloneV7Config,
+  standaloneV8Config,
+  standaloneEmptyConfig,
+  standaloneBigConfig,
+  clusterConfig,
+  sentinelConfig,
+  sshConfig,
+};
+
+// Backward compatibility - aggregated config object
```
valkirilov: nit: Why do we seek backwards compatibility, this is a new project? 

Or it's based on the way we have the tests in the testcafe folder, and the LLM decided to put it?

tests/e2e-playwright/config/databases/standalone.ts
```diff
@@ -0,0 +1,34 @@
+import { getEnv, getEnvNumber } from '../env';
+
+/**
+ * Standalone Redis configurations
+ */
+export const standaloneConfig = {
```
valkirilov: nit: Can we type-hint all these configurations with a generic interface?

tests/e2e-playwright/config/env.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (value === undefined) {
+    throw new Error(`Missing required environment variable: ${key}`);
+  }
+  return value;
+}
+
+/**
+ * Get environment variable as number
+ */
+export function getEnvNumber(key: string, defaultValue?: number): number {
```
valkirilov, line 42: nit: Can't we combine this with `getEnv`? Why do we need a special variation of the same function that will work with numbers, only when we explicitly use it? 

I think it should work like that, by default, unless we have many environment variables being numbers, but typed as strings 🤷‍♂️

PS: Same is valid for `getEnvOptional`. Yes, these are simple functions and they don't do any harm, yet seems odd to have them.

tests/e2e-playwright/helpers/api.ts
```diff
@@ -0,0 +1,358 @@
+import { request, APIRequestContext } from '@playwright/test';
+import { AddDatabaseConfig, DatabaseInstance } from 'e2eSrc/types';
+import { TEST_DB_PREFIX } from 'e2eSrc/test-data/databases';
+
+/**
+ * API Helper for database operations
+ * Used for test setup/teardown to avoid slow UI interactions
+ */
+export class ApiHelper {
```
valkirilov, line 9: nit: Is there a potential for this file to grow enormously in the future, because currently it does contain a few categories of utils:
- database related - create/delete database
- keys related - create keys of each type
- other, like settings and eula agreements

So I wonder whether it makes sense to split the helpers, or you think it's too much of premature optimization, and it's fine like we have it now. 

PS: I'm asking, because you know the end results (having the AI agent write all the tests).

tests/e2e-playwright/pages/browser/BrowserPage.ts
```diff
@@ -0,0 +1,164 @@
...
+ * Main page for browsing Redis keys
+ */
+export class BrowserPage extends BasePage {
+  // Components
+  readonly addKeyDialog: AddKeyDialog;
+  readonly bulkActionsPanel: BulkActionsPanel;
+  readonly keyDetails: KeyDetails;
+  readonly keyList: KeyList;
+
+  // Navigation tabs
```
valkirilov: nit: Minor, but these relate more to the general layout, but not to the Browser page itself.
We're going to have them in the rest of the pages, so we're either duplicating them, or exporting them into the base page object.

PS: Same is valid for the "Database info" and "Bottom panel" sections below.

Edit: Now I see we already have it in `BasePage`, so it looks like a duplication then 🤔

tests/e2e-playwright/pages/browser/BrowserPage.ts
```diff
@@ -0,0 +1,164 @@
...
+    this.keyList = new KeyList(page);
+
+    // Navigation tabs
+    this.browseTab = page.getByRole('tab', { name: 'Browse' });
+    this.workbenchTab = page.getByRole('tab', { name: 'Workbench' });
+    this.analyzeTab = page.getByRole('tab', { name: 'Analyze' });
+    this.pubSubTab = page.getByRole('tab', { name: 'Pub/Sub' });
+
+    // Action buttons
+    this.addKeyButton = page.getByTestId('btn-add-key');
```
valkirilov: nit: I think we can select this button via its role and label, instead of going straight for the test-id (based on our discussion for preferred selectors precedence).

tests/e2e-playwright/pages/browser/BrowserPage.ts
```diff
@@ -0,0 +1,164 @@
...
+    await expect(this.addKeyDialog.title).toBeVisible();
+  }
+
+  async closeAddKeyDialog(): Promise<void> {
+    if (await this.addKeyDialog.isVisible()) {
+      await this.addKeyDialog.clickCancel();
+    }
+  }
+
+  async navigateToWorkbench(): Promise<void> {
```
valkirilov: nit: related to https://github.com/redis/RedisInsight/pull/5423/changes#r2753742292, but I think this should live in `WorkbenchPage.goto()`, so it's not related to the Browser page object.

PS: The same applies to the rest below - `navigateToAnalyze`, `navigateToPubSub`, `openCli`.

tests/e2e-playwright/pages/browser/BrowserPage.ts
```diff
@@ -0,0 +1,164 @@
...
+   */
+  async goto(databaseId: string): Promise<void> {
+    await this.gotoDatabase(databaseId);
+    await this.waitForLoad();
+  }
+
+  async waitForLoad(): Promise<void> {
+    await this.page.waitForLoadState('domcontentloaded');
+    // Wait for either key list to load (Total: or Results:) or no keys message
+    await Promise.race([
```
valkirilov: nit: Looks like `KeyList.waitForKeysLoaded`, or we don't want it identical?

tests/e2e-playwright/pages/browser/components/KeyDetails.ts
```diff
@@ -0,0 +1,887 @@
...
+
+    // Set-specific
+    this.addMembersButton = page.getByRole('button', { name: 'Add Members' });
+    this.setGrid = page.getByTestId('set-details');
+
+    // ZSet-specific (Sorted Set)
+    this.zsetGrid = page.getByTestId('zset-details');
+    this.scoreSortButton = page.getByRole('button', { name: /Score/ });
+
+    // Stream-specific
```
valkirilov, line 125: nit: I think it will be easier if we split this into separate page objects, based on the various key types. 

Right now, it's a huge file ~1000 lines of code that the LLM wrote, but I believe it will make it easier if we have the split, so you know the specific UI elements related to each key type. 

So, it's minor refactoring, but I would have all the key related helpers below moved out.

tests/e2e-playwright/pages/browser/components/BulkActionsPanel.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  async close(): Promise<void> {
+    await this.closeButton.click();
+    await this.content.waitFor({ state: 'hidden', timeout: 5000 });
+  }
+
+  async isOpen(): Promise<boolean> {
+    try {
+      await this.content.waitFor({ state: 'visible', timeout: 2000 });
```
valkirilov, line 77: nit: Do we need to pass explicitly the timeout fallback, instead of relying on the default value. 

And why do we go for different magic numbers instead 🤔

tests/e2e-playwright/pages/browser/components/BulkActionsPanel.ts
```diff
@@ -0,0 +1,167 @@
...
+    // Wait for the status completed message to appear
+    await this.statusCompleted.waitFor({ state: 'visible', timeout: 30000 });
+  }
+
+  async performBulkDelete(): Promise<void> {
+    // Click delete button
+    await this.clickDelete();
+    // Confirm in the dialog
+    await this.confirmDelete();
+    // Wait for completion
```
valkirilov: nit: I know this is LLM generated, but these comments say 1:1 the line bellow that. Is there a way to tweak the granularity of the comments, and slightly reduce them?

tests/e2e-playwright/pages/cli/CliPanel.ts
```diff
@@ -0,0 +1,135 @@
...
+   */
+  async executeCommand(command: string): Promise<void> {
+    // Focus on the command input
+    await this.commandInput.focus();
+    // Type the command
+    await this.page.keyboard.type(command);
+    // Press Enter to execute
+    await this.page.keyboard.press('Enter');
+    // Wait a bit for the command to execute
+    await this.page.waitForTimeout(500);
```
valkirilov: nit: Is there a smarter way to verify the command is executed, because static timeouts are usually not the best way to do that (it's a hidden flakiness)?

PS: I believe this will be used with combination with `waitForOutput` and it will be enough to verify the result, without the static await.

tests/e2e-playwright/pages/cli/CliPanel.ts
```diff
@@ -0,0 +1,135 @@
...
+   */
+  async isOpen(): Promise<boolean> {
+    return this.hideButton.isVisible();
+  }
+
+  /**
+   * Execute a command in the CLI
+   */
+  async executeCommand(command: string): Promise<void> {
+    // Focus on the command input
```
valkirilov: nit: One more example of useless comments.

Reference: https://github.com/redis/RedisInsight/pull/5423/changes#r2754064531

tests/e2e-playwright/pages/BasePage.ts
```diff
@@ -0,0 +1,225 @@
...
+   * Press keyboard shortcut
+   */
+  async pressKey(key: string): Promise<void> {
+    await this.page.keyboard.press(key);
+  }
+
+  /**
+   * Get text content of an element
+   */
+  async getText(locator: Locator): Promise<string | null> {
```
valkirilov: nit: It's minor and not a problem, but I think the LLM went too much here, in creating aliases for one-line utils that are part of the Playwright API already.

tests/e2e-playwright/test-data/browser/index.ts
```diff
@@ -0,0 +1,130 @@
...
+ */
+export function generateKeyName(suffix?: string): string {
+  const uniqueId = faker.string.alphanumeric(8);
+  return suffix ? `${TEST_KEY_PREFIX}${suffix}-${uniqueId}` : `${TEST_KEY_PREFIX}${uniqueId}`;
+}
+
+/**
+ * Generate String key data
+ */
+export function getStringKeyData(overrides?: Partial<StringKeyData>): StringKeyData {
```
valkirilov: nit: Very minor, but since these utils are generators (as confirmed by the jsdoc), I would rather call them `generate*` instead of `get*`

tests/e2e-playwright/test-data/browser/index.ts
```diff
@@ -0,0 +1,130 @@
...
+      { field: faker.word.noun(), value: faker.lorem.word() },
+    ],
+    ...overrides,
+  };
+}
+
+/**
+ * Generate List key data
+ */
+export function getListKeyData(overrides?: Partial<ListKeyData>): ListKeyData {
```
valkirilov: I'll put a reference to my comment for the factories, because I think this is the place where they come in handy and can truly shine 🙂

- https://github.com/redis/RedisInsight/pull/5423/changes#r2753533520

tests/e2e-playwright/.eslintrc.json
```diff
@@ -0,0 +1,19 @@
+{
```
valkirilov, line 1: Question: Is there a conflict between this config and the main one in the root directory, that is responsible for the linking of the entire project? 

We should be able to run the linter in CI for these e2e test files as well, and if we can make it without additional setup, it will be nice 👍 

PS: I'm not against having a single config, or split between different directories, but I remember @pawelangelow spent some time on it, so maybe he can advise better.

PS2: Same applies for the prettier config.

tests/e2e-playwright/README.md
```diff
@@ -0,0 +1,307 @@
...
+
+## Prerequisites
+
+1. Start Redis instances using Docker Compose:
+   ```bash
+   cd tests/e2e
+   docker-compose -f rte.docker-compose.yml up -d
+   ```
+
+2. Start RedisInsight application on `http://localhost:8080`
```
valkirilov: nit: An example will be handy here, for clarity and completeness 🙂

```
yarn dev:api
yarn dev:ui
```

tests/e2e-playwright/tests/main/browser/key-filtering/key-filtering.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+    });
+
+    test(`should filter keys with question mark (?) single character wildcard`, async ({
+      browserPage,
+    }) => {
+      // Search for keys matching the pattern with ? wildcard (matches single char)
+      await browserPage.keyList.searchKeys(`${TEST_KEY_PREFIX}item?-${uniqueSuffix}`);
+
+      // Wait for results
+      await browserPage.page.waitForTimeout(500);
```
valkirilov: nit: This might lead to flakiness in some cases - it's better to ensure we wait for the UI to be rendered (async), instead of relying on a static timeout.

.ai/rules/e2e-testing.md
KrumTy: I can remove it but I added it because during PoC there was a certain problem and AI started rewriting tests to be serial. So I decided to explicitly tell it not to do it

.ai/rules/e2e-testing.md
KrumTy: Yes, that can be seen as a drawback. However it's just navigation and it happens very fast from what I've seen. There are navigation helpers inside the page objects.

.ai/rules/e2e-testing.md
KrumTy: Well it's an unordered list. Lets find an issue in action and then adjust the rule

tests/e2e-playwright/config/databases/index.ts
KrumTy: leftover

tests/e2e-playwright/config/env.ts
KrumTy: It is for type safety. We can't know a variable's type by the env file alone. Combining all into 1 function would cost us type safety that's all

tests/e2e-playwright/helpers/api.ts
KrumTy: This is the helper from the PoC more or less so it covers a lot of stuff already. I don't expect it to grow a lot more but we can always split if needed and export from a barrel. It's just for setting up the data for an e2e to happen (via api calls)

tests/e2e-playwright/tests/main/browser/key-filtering/key-filtering.spec.ts
KrumTy: yes but we also want relatively fast runs. So if it starts causing flakiness with 500ms we might want to see what's causing the search to be that slow (it's redis after all).. and address that issue. Or fallback to your suggestion

tests/e2e-playwright/.eslintrc.json
KrumTy: lets keep the tests decoupled, I will include lint checks in the workflows

tests/e2e-playwright/pages/browser/components/BulkActionsPanel.ts
KrumTy: the default timeout for the playwright project is 30s, we need to pass overrides when we want to speed it up. 
And having constants over "magic numbers" won't help a lot as each "wait" should be tweaked individually

.ai/rules/e2e-testing.md
valkirilov: Thanks! 🙌

tests/e2e-playwright/pages/browser/components/BulkActionsPanel.ts
```diff
@@ -0,0 +1,161 @@
...
+    this.deleteSummary = page.getByTestId('bulk-delete-summary');
+    this.deleteCompletedSummary = page.getByTestId('bulk-delete-completed-summary');
+    this.deleteInfo = page.getByTestId('bulk-actions-info');
+    this.downloadReportCheckbox = page.getByRole('checkbox', { name: 'Download report' });
+    this.statusCompleted = page.getByTestId('bulk-status-completed');
+
+    // Upload Data elements
+    this.fileInput = page.getByTestId('bulk-upload-file-input');
+    this.uploadContainer = page.getByTestId('bulk-upload-container');
+    this.uploadButton = page.getByTestId('bulk-action-warning-btn');
```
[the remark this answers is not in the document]
KrumTy: these are the correct ids, the buttons are in different tabs and not visible at the same time

tests/e2e-playwright/.husky/pre-commit
KrumTy: yeah, it's not enforced on a team level and most people prefer doing it manually for some reason

this won't change anything for people who do not setup husky

tests/e2e-playwright/pages/browser/components/KeyDetails.ts
KrumTy: leaving it as it is for now since it's one form and I've set grouping by action rather than key type in tests; we might go too granular otherwise

tests/e2e-playwright/pages/BasePage.ts
```diff
@@ -0,0 +1,218 @@
...
+    await locator.clear();
+    await locator.fill(value);
+  }
+
+  /**
+   * Click and wait for navigation
+   */
+  async clickAndWaitForNavigation(locator: Locator): Promise<void> {
+    await Promise.all([this.page.waitForURL(/.*/, { waitUntil: 'domcontentloaded' }), locator.click()]);
+  }
```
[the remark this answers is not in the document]
KrumTy: these will be used soon

tests/e2e-playwright/pages/browser/components/BulkActionsPanel.ts
```diff
@@ -0,0 +1,161 @@
...
+    this.uploadStatusCompleted = page.getByTestId('bulk-status-completed');
```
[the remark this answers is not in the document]
KrumTy: explained in an earlier comment

PR #5464, opened 2026-02-09
RI-7909: Index List

dantovska:
# What

- Create`IndexList` component, which is used to list all indexes using the redis-ui `Table` component
- This component is the equivalent of the deprecated `ManageIndexesList` component
- Includes columns with the index data
- Includes `Query` button by providing callback
- Includes actions (with a menu) which are Label + callback array

image

# Testing

It can be tested using storybook:

`yarn build-sb`

`yarn sb`

Navigate to `vector-search` -> `IndexesList`

---

> [!NOTE]
> **Low Risk**
> Primarily new UI components, tests, and shared formatting helpers; minimal impact on runtime behavior outside the vector-search UI.
> 
> **Overview**
> Introduces a new `IndexList` component (and supporting cells/config/constants/types) that renders vector-search indexes in a redis-ui `Table` with sortable numeric columns, header tooltips, and an actions column that can show a **Query** button and a menu of custom callbacks.
> 
> Adds Storybook stories and comprehensive component tests, plus a new mock `indexListRowFactory`/sample data for repeatable UI testing. Prefix formatting is centralized into `pages/vector-search/utils/formatPrefixes` and `IndexInfo` is updated to use it; the deprecated `ManageIndexesList` is annotated to point reviewers to the new component.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f50e18c63fc1cd552367db36d7922602b30d6d4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/components/index-list/components/NumericCell/NumericCell.tsx
+++ b/redisinsight/ui/src/pages/vector-search/components/index-list/components/NumericCell/NumericCell.tsx
@@ -1,14 +1,10 @@
-import React, { useMemo } from 'react'
+import React from 'react'
 
 import { Text } from 'uiSrc/components/base/text'
 import { NumericCellProps } from './NumericCell.types'
 
-export const NumericCell = ({ value, testId }: NumericCellProps) => {
-  const displayValue = useMemo(() => value, [value])
-
-  return (
-    <Text size="s" data-testid={testId}>
-      {displayValue}
-    </Text>
-  )
-}
+export const NumericCell = ({ value, testId }: NumericCellProps) => (
+  <Text size="s" data-testid={testId}>
+    {value}
+  </Text>
+)


redisinsight/ui/src/pages/vector-search/components/indexes-list/IndexesList.tsx
```diff
@@ -0,0 +1,31 @@
...
+    () => getIndexesListColumns({ onQueryClick, actions }),
+    [onQueryClick, actions],
+  )
+
+  return (
+    <Table
+      data={data}
+      columns={columns}
+      stripedRows
+      emptyState={emptyMessage}
```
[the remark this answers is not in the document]
dantovska: this is irrelevant, since one of emptyState allowed types is string

redisinsight/ui/src/mocks/factories/vector-search/indexList.factory.ts
```diff
@@ -0,0 +1,128 @@
...
+      prefixes: faker.helpers.arrayElements(
+        [
+          `${faker.word.noun()}:`,
+          `${faker.word.noun()}:`,
+          `${faker.word.noun()}:`,
+        ],
+        faker.number.int({ min: 0, max: 3 }),
+      ),
+      fieldTypes,
+      numDocs: faker.number.int({ min: 0, max: 1000000 }),
```
valkirilov, line 35: nit: I would suggest creating non-overlapping ranges, because it may produce flakiness in the unit tests, whenever we check the table for the actual rendered values. 

Here, we have big numbers, and the chance is low, but generally, I have used this practice, because sometimes we end up with the same number rendered in two separate columns ( for example, from `numDocs` and `numTerms`), and it causes the tests to fail. 

This simple tweak is the workaround I usually do.

```suggestion
      numDocs: faker.number.int({ min: 0, max: 1000000 }),
      numRecords: faker.number.int({ min: 1000001, max: 2000000 }),
      numTerms: faker.number.int({ min: 2000001, max: 500000 }),
```

redisinsight/ui/src/mocks/factories/vector-search/indexList.factory.ts
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Factory for IndexListRow type.
+ * Used for component tests and Storybook.
+ */
+export const indexListRowFactory = Factory.define<IndexListRow>(
+  ({ sequence }) => {
+    const numFields = faker.number.int({ min: 1, max: 8 })
+    const fieldTypes = faker.helpers.arrayElements(
+      FIELD_TYPES,
```
valkirilov: nit: I think you don't need to explicitly redefine the types in the array with `FIELD_TYPES`, you can give it a try:

```suggestion
      Object.values(FieldTypes),
```

redisinsight/ui/src/mocks/factories/vector-search/indexList.factory.ts
dantovska: oh nice! good catch - I didn't think of that.

PR #5465, opened 2026-02-09
RI-7911: Add automatic refresh of Azure Entra ID tokens

pawelangelow:
# What

Implements automatic token refresh for Azure Entra ID authenticated Redis databases using an event-driven architecture.

How it works:

- Token acquired →  AzureAuthService emits event →  AzureTokenRefreshManager schedules timer + re-authenticates clients
- Timer fires (5 min before expiry) → checks for active clients → requests new token → event emitted → back to first flow ☝️ 
- Re-authentication → filters clients by  tokenExpiresOn → sends AUTH command → updates token expiry

# Testing

1. Connect to an Azure Entra ID authenticated Redis database
2. Verify token refresh is scheduled (check logs for "Scheduled token refresh" message)
3. Wait for token to approach expiration - refresh should happen automatically (you can find some good read in the meantime, if you're out of options, try [this](https://eslint.org/blog/2026/02/eslint-v10.0.0-released/) one)
4. Verify connection remains active after refresh

---

> [!NOTE]
> **Medium Risk**
> Introduces background timers and automatic `AUTH` re-authentication based on token expiry, which can affect connection stability if scheduling or client filtering is wrong. Scope is contained to Azure Entra ID flows but touches shared Redis client construction/storage paths.
> 
> **Overview**
> Adds automatic refresh for Azure Entra ID Redis auth by emitting `AzureRedisTokenEvents.Acquired` from `AzureAuthService.getRedisTokenByAccountId()` and introducing `AzureTokenRefreshManager` to schedule pre-expiry refreshes and re-`AUTH` active clients.
> 
> To support this, Redis client instances now retain a small slice of database metadata (`providerDetails`, including new `tokenExpiresOn`) and `RedisClientStorage` can query clients by nested database fields; connection strategies and mocks were updated to pass database metadata through, and tests were added/extended to cover event emission, timer/race-condition behavior, and client re-auth filtering.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 37833eaa0dab3995d4930dfcd3981b1f07be0ea7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
+++ b/redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
@@ -47,8 +47,14 @@ export class AzureTokenRefreshManager implements OnModuleDestroy {
     accountId: string;
     tokenResult: AzureTokenResult;
   }): Promise<void> {
-    this.scheduleRefresh(accountId, tokenResult.expiresOn);
-    await this.reAuthenticateClients(accountId, tokenResult);
+    try {
+      this.scheduleRefresh(accountId, tokenResult.expiresOn);
+      await this.reAuthenticateClients(accountId, tokenResult);
+    } catch (error) {
+      this.logger.error(
+        `Failed to handle token acquired event for account ${accountId}: ${error.message}`,
+      );
+    }
   }
 
   scheduleRefresh(azureAccountId: string, expiresOn: Date): void {


redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+      accountId: DEFAULT_ACCOUNT_ID,
+    };
+
+    try {
+      const tokenResult =
+        await this.azureAuthService.getRedisTokenByAccountId(azureAccountId);
+
+      if (!tokenResult) {
+        this.logger.warn(
+          `Failed to refresh token for database ${databaseId} - re-authentication may be needed`,
```
KrumTy: Is it worth retrying this at least once, in case of network/azure issue?

redisinsight/api/src/modules/azure/auth/utils/jwt.utils.ts
```diff
@@ -0,0 +1,17 @@
...
+import { decode } from 'jsonwebtoken';
+
+export const getJwtExpiration = (token: string): Date | null => {
+  const payload = decode(token);
+  if (!payload || typeof payload === 'string' || !payload.exp) {
+    return null;
+  }
+
+  // exp is in seconds since epoch
+  return new Date(payload.exp * 1000);
```
ArtemHoruzhenko: is it better to wrap it by try-catch block and return null in case of an error?

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    const refreshAt =
+      expiration.getTime() - TOKEN_REFRESH_BUFFER_SECONDS * 1000;
+    const now = Date.now();
+    // Enforce minimum delay to prevent infinite loops if token expires within buffer
+    const delayMs = Math.max(
+      MIN_TOKEN_REFRESH_DELAY_SECONDS * 1000,
+      refreshAt - now,
+    );
```
ArtemHoruzhenko: what will happen if `refreshAt - now` < `MIN_TOKEN_REFRESH_DELAY_SECONDS * 1000` ?

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+      databaseId,
+      token,
+      providerDetails,
+      clientIds,
+    );
+
+    if (!scheduled) {
+      this.refreshTimers.delete(databaseId);
+    }
+  }
```
ArtemHoruzhenko: seems like redundant function. Why not to cover `if (!scheduled)` right inside `createRefreshTimer` method?

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+    const reAuthPromises = [...clientIds].map(async (clientId) => {
+      try {
+        const client = await this.redisClientStorage.get(clientId);
+        if (!client || !client.isConnected()) {
+          // Remove stale client IDs to prevent keeping timers alive indefinitely
+          clientIds.delete(clientId);
+          this.logger.debug(`Removed stale client ${clientId}`);
+          return;
+        }
+        await client.sendCommand(['AUTH', username, newToken]);
```
ArtemHoruzhenko: not sure if azure supports oss cluster, but if so will 'AUTH' command will be redirected to a single node from or for all the nodes of the cluster or should we add 'auth' method to redis.client?

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+    newToken: string,
+  ): Promise<void> {
+    this.logger.debug(`Re-authenticating ${clientIds.size} clients`);
+
+    const reAuthPromises = [...clientIds].map(async (clientId) => {
+      try {
+        const client = await this.redisClientStorage.get(clientId);
+        if (!client || !client.isConnected()) {
+          // Remove stale client IDs to prevent keeping timers alive indefinitely
+          clientIds.delete(clientId);
```
ArtemHoruzhenko: just a question: does it mean that we just stop tracking auto refresh for that particular client? but it might be in the client pool and fail at some point?

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+      const latestTimer = this.refreshTimers.get(databaseId);
+      const clientIds = latestTimer?.clientIds || new Set<string>();
+
+      if (clientIds.size === 0) {
+        this.logger.debug(
+          `No active clients for database ${databaseId} after refresh, not scheduling next refresh`,
+        );
+        this.refreshTimers.delete(databaseId);
+        return;
+      }
```
ArtemHoruzhenko: does it make sense to have have only one such check that we already have right below `reAuthenticateClients` to simplify code?

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+        password: tokenResult.token,
+      });
+
+      this.logger.debug(
+        `Token refreshed for database ${databaseId}, expires at ${tokenResult.expiresOn.toISOString()}`,
+      );
+
+      // Get fresh reference to clientIds in case it changed during async operations
+      const latestTimer = this.refreshTimers.get(databaseId);
+      const clientIds = latestTimer?.clientIds || new Set<string>();
```
ArtemHoruzhenko: shouldn't `this.refreshTimers.get` return empty Set to avoid `latestTimer?.clientIds || new Set<string>();` in all places where it is used and follow single responsibility principle?

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+      const database = await this.databaseRepository.get(
+        sessionMetadata,
+        databaseId,
+      );
+      // Use localAccountId (Azure OID) for AUTH command, not azureAccountId (MSAL identifier)
+      const username = database.username || tokenResult.account.localAccountId;
+
+      await this.databaseRepository.update(sessionMetadata, databaseId, {
+        password: tokenResult.token,
+      });
```
ArtemHoruzhenko: why do we need this?

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+      // Re-check timer state after async operations to handle race conditions.
+      // handleClientRemoved may have deleted the timer while we were refreshing.
+      const currentTimer = this.refreshTimers.get(databaseId);
+      if (!currentTimer || currentTimer.clientIds.size === 0) {
+        this.logger.debug(
+          `No active clients for database ${databaseId}, skipping token update`,
+        );
+        this.refreshTimers.delete(databaseId);
+        return;
+      }
```
ArtemHoruzhenko: again. we can remove this check to simplify everything. we will remove async operation below and make  `reAuthenticateClients` handle empty Set.

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+      return;
+    }
+
+    this.logger.debug(`Refreshing token for database ${databaseId}...`);
+
+    const sessionMetadata = {
+      sessionId: DEFAULT_SESSION_ID,
+      userId: DEFAULT_USER_ID,
+      accountId: DEFAULT_ACCOUNT_ID,
+    };
```
ArtemHoruzhenko: no-no. We can't do this. let's pass sessionMetadata with other data on event emit.

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+  ): Promise<void> {
+    const { azureAccountId } = providerDetails;
+
+    if (!azureAccountId) {
+      this.logger.warn(
+        `Database ${databaseId} missing azureAccountId, cannot refresh token`,
+      );
+      this.refreshTimers.delete(databaseId);
+      return;
+    }
```
ArtemHoruzhenko: I don't understand why we are checking it here. Shouldn't we assume that this particular service will have only azure clients?

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+    return this.refreshTimers.size;
+  }
+
+  @OnEvent(RedisClientEvents.ClientStored)
+  async handleClientStored(event: RedisClientLifecycleEvent): Promise<void> {
+    const { clientId, databaseId } = event;
+
+    try {
+      // Check if we already have a refresh timer for this database
+      const existingTimer = this.refreshTimers.get(databaseId);
```
ArtemHoruzhenko: relying on databaseId is not multi-tenant approach in our case. I understand that by design we support single user for desktop apps but this makes the service not scalable.

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,362 @@
...
+      }
+
+      this.logger.debug(
+        `Azure Entra ID client stored: ${clientId} (database: ${database.name})`,
+      );
+
+      await this.scheduleTokenRefresh(
+        clientId,
+        databaseId,
+        database.password,
```
ArtemHoruzhenko: ok. I was reviewing from end to start. Now I see why we need to store this, but I feel like it creates bunch of race conditions cases.

redisinsight/api/src/common/models/client-metadata.ts
```diff
@@ -49,6 +50,9 @@ export class ClientMetadata {
   @Max(2147483647)
   db?: number;
 
+  @IsOptional()
+  providerDetails?: ProviderDetails;
+
```
ArtemHoruzhenko: why do we want to change ClientMetadata? I thought we agreed to add entire database object into client itself

redisinsight/api/src/modules/database/providers/database.client.factory.ts
```diff
@@ -141,6 +141,10 @@ export class DatabaseClientFactory {
 
     database = await this.credentialProvider.resolve(database);
 
+    if (database.providerDetails) {
+      clientMetadata.providerDetails = database.providerDetails;
+    }
+
```
ArtemHoruzhenko: below you can see that we are passing entire database object into .createClient method

redisinsight/api/src/modules/redis/redis.client.storage.ts
```diff
@@ -212,4 +212,20 @@ export class RedisClientStorage {
...
+   * @param fieldPath - dot-notation path to the field in clientMetadata
+   * @param value - value to match
+   */
+  public getClientsByMetadataField(
+    fieldPath: string,
+    value: unknown,
+  ): RedisClient[] {
+    return [...this.clients.values()].filter(
+      (client) =>
+        get(client.clientMetadata, fieldPath) === value && client.isConnected(),
```
ArtemHoruzhenko: and here we can find it by `get(client.database, fieldPath)`

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,195 @@
...
+  private clearAllTimers(): void {
+    this.timers.forEach((timerInfo, accountId) => {
+      clearTimeout(timerInfo.timer);
+      this.logger.debug(`Cleared timer for account ${accountId}`);
+    });
+    this.timers.clear();
+  }
+
+  @OnEvent(RedisClientEvents.ClientStored)
+  async handleClientStored(event: ClientStoredEvent): Promise<void> {
```
ArtemHoruzhenko: why do we need to listen on `RedisClientEvents.ClientStored` event?

redisinsight/api/src/modules/azure/azure.module.ts
```diff
@@ -15,8 +16,14 @@ import { DatabaseModule } from '../database/database.module';
     AzureAuthAnalytics,
     AzureAutodiscoveryService,
     AzureAutodiscoveryAnalytics,
+    AzureAutodiscoveryService,
+    AzureTokenRefreshManager,
```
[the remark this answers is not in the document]
pawelangelow: oh, rebasing issues :D

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,145 @@
...
+    const timer = setTimeout(() => {
+      this.refreshTokenAndReauth(azureAccountId).catch((error) => {
+        this.logger.error(
+          `Token refresh failed for account ${azureAccountId}: ${error.message}`,
+        );
+      });
+    }, delay);
+
+    this.timers.set(azureAccountId, timer);
+  }
```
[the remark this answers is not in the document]
pawelangelow: It's a valid comment - to remove the timer, we need to listen to clients' lifecycle events, but the logic gets too complicated. The worst thing here is that we'll have a timer for the refresh period, and then it will do nothing. At this point, I think we can live with this simpler approach.

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,173 @@
...
+      this.logger.debug(
+        `No active clients for account ${azureAccountId}, stopping refresh cycle`,
+      );
+      return;
+    }
+
+    const tokenResult =
+      await this.azureAuthService.getRedisTokenByAccountId(azureAccountId);
+
+    if (!tokenResult) {
```
ArtemHoruzhenko: nit: why we need additional check here? shouldn't it be part of `getRedisTokenByAccountId` function?

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  clearAllTimers(): void {
+    this.timers.forEach(({ timeout }) => clearTimeout(timeout));
+    this.timers.clear();
+  }
+
+  private async refreshToken(azureAccountId: string): Promise<void> {
+    this.logger.debug(`Refreshing token for account ${azureAccountId}`);
+
+    this.timers.delete(azureAccountId);
```
ArtemHoruzhenko: why we need delete timer here? 
timer should be deleted from a single place when you do not have active clients otherwise you are opening box for race condition

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,173 @@
...
+    const clients = this.redisClientStorage.getClientsByDatabaseField(
+      'providerDetails.azureAccountId',
+      azureAccountId,
+    );
+
+    if (clients.length === 0) {
+      this.logger.debug(
+        `No active clients for account ${azureAccountId}, stopping refresh cycle`,
+      );
+      return;
```
ArtemHoruzhenko, line 136: we can remove timer here

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,173 @@
...
+    tokenResult,
+  }: {
+    accountId: string;
+    tokenResult: AzureTokenResult;
+  }): Promise<void> {
+    this.scheduleRefresh(accountId, tokenResult.expiresOn);
+    await this.reAuthenticateClients(accountId, tokenResult);
+  }
+
+  scheduleRefresh(azureAccountId: string, expiresOn: Date): void {
```
ArtemHoruzhenko, line 60: nit: seems like private method. Also I would prefer to to have this code inside `handleTokenAcquired` to not have possibility occasionally invoke it from other place. with current architecture there is only one scenario when schedule refresh should be triggered

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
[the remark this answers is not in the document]
ArtemHoruzhenko: The only difference between current approach and listening for clients removals is when timer will be deleted. In current approach we will remove timer when find out that there are no active clients. It is more simple and robust solution in our case.

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,173 @@
...
+    const existing = this.timers.get(azureAccountId);
+
+    // Skip if already scheduled for the same expiry time (race condition protection)
+    if (existing?.expiresOn?.getTime() === expiresOn.getTime()) {
+      return;
+    }
+
+    this.clearTimer(azureAccountId);
```
ArtemHoruzhenko: we can't do this. this methods removes timer from the map and opens door for race condition. we must clearTimeout only on existing timer

redisinsight/api/src/modules/azure/azure-token-refresh.manager.ts
```diff
@@ -0,0 +1,173 @@
...
+        try {
+          await client.call([
+            'AUTH',
+            tokenResult.account.localAccountId,
+            tokenResult.token,
+          ]);
+          // Update the client's token expiry after successful re-auth
+          // eslint-disable-next-line no-param-reassign
+          client.database.providerDetails.tokenExpiresOn =
+            tokenResult.expiresOn;
```
ArtemHoruzhenko, line 184: Nit: No need to do anything here but looking on this code I have more strong feeling that it must be part of the client

pawelangelow: let's leave it as is for now

PR #5487, opened 2026-02-12
RI-8085: Add switch account functionality for Azure OAuth login

pawelangelow:
# What

Added the ability to switch Azure accounts during the OAuth login flow by passing the `prompt=select_account` parameter to MSAL's authorization URL. This forces the Azure account picker to appear, allowing users to sign in with a different account.

---

> [!NOTE]
> **Medium Risk**
> Touches the Azure OAuth login initiation path end-to-end (UI thunk/hook + API controller/service), so regressions could break sign-in or change login UX, though the change is additive and gated by an optional parameter.
> 
> **Overview**
> Adds support for an optional `prompt` query param on the Azure OAuth login endpoint and forwards it through MSAL URL generation, enabling flows like **account switching** via `prompt=select_account`.
> 
> Updates the UI Azure auth hook and thunk to accept options (including `prompt`) and introduces a `switchAccount` action used from the subscriptions page with telemetry; tests are added/updated and a small `AzureAccountFactory` is introduced to standardize test fixtures.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 29fbdf221b31b78da56dcded0e14fa5c615d5594. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/pages/autodiscover-azure/azure-databases/AzureDatabasesPage.spec.tsx
+++ b/redisinsight/ui/src/pages/autodiscover-azure/azure-databases/AzureDatabasesPage.spec.tsx
@@ -17,6 +17,7 @@ import {
 } from 'uiSrc/slices/instances/azure'
 import { azureAuthAccountSelector } from 'uiSrc/slices/oauth/azure'
 import { AzureRedisDatabase, AzureRedisType } from 'uiSrc/slices/interfaces'
+import { AzureAccountFactory } from 'uiSrc/mocks/factories/cloud/AzureAccount.factory'
 
 import AzureDatabasesPage from './AzureDatabasesPage'
 
@@ -54,11 +55,7 @@ const mockSubscription = {
   state: 'Enabled',
 }
 
-const mockAccount = {
-  id: faker.string.uuid(),
-  username: faker.internet.email(),
-  name: faker.person.fullName(),
-}
+const mockAccount = AzureAccountFactory.build()
 
 const defaultAzureState = {
   loading: false,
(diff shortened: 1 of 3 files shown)


redisinsight/ui/src/components/hooks/useAzureAuth.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+  }),
+  initiateAzureLoginAction: jest.fn().mockReturnValue({ type: 'mock-action' }),
+}))
+
+const mockedAzureAuthSelector = azureAuthSelector as jest.Mock
+const mockedInitiateAzureLoginAction = initiateAzureLoginAction as jest.Mock
+
+let store: typeof mockedStore
+
+const mockAccount = {
```
valkirilov: nit: It's very small, but still it can became a reusable factory, in case you think we might need it on few more places?

redisinsight/ui/src/slices/oauth/azure.ts
```diff
@@ -115,12 +115,17 @@ export const azureAuthSourceSelector = (state: RootState) =>
 // The reducer
 export default azureAuthSlice.reducer
 
+export interface InitiateAzureLoginOptions {
+  source: AzureLoginSource
+  prompt?: 'select_account' | 'login' | 'consent'
```
valkirilov, line 120: I was heading to ask why not using `AzureOAuthPrompt`, and then I saw it's living inside a DTO on the back-end 😀

valkirilov (dismissed):

redisinsight/ui/src/slices/oauth/azure.ts
pawelangelow: Yeah, not very easy to reuse BE types. Since it's a single place usage, I didn't define a specified type.

PR #5525, opened 2026-02-20
RI-7934: Show confirmation modal when deleting index

dantovska:
# What

When clicking on the three dots actions -> Delete, delete confirmation modal opens and the index gets deleted only when Delete index is clicked, and the modal is closed when Keep index / close button are clicked. 

# Testing

1. Navigate to Search tab
2. Create index sample data
3. Visualize the Index List table (change navigation to <database-id>/vector-search)
4. Click on the three dots -> Delete
5. Click Delete index / Keep index 

https://github.com/user-attachments/assets/81e6cfb0-e867-4384-89fe-a9c9d54e4c6a

---

> [!NOTE]
> **Low Risk**
> UI-level flow change gating an existing delete action plus test updates; minimal surface area, with low risk of regressions beyond the delete interaction.
> 
> **Overview**
> Vector Search index deletion now opens a confirmation modal and only dispatches `deleteRedisearchIndexAction` after the user confirms; cancel/close dismisses the modal without deleting.
> 
> `ListContent` tracks a pending index to delete, triggers telemetry on successful delete, and `ActionsCell` adds a stable `data-testid` on the actions menu trigger to support tests. Tests were updated to cover confirm vs cancel behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 76e8c806b17af84ce87ac733d35d80bf59d243c7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/pages/VectorSearchListPage/components/list-content/ListContent.spec.tsx
+++ b/redisinsight/ui/src/pages/vector-search/pages/VectorSearchListPage/components/list-content/ListContent.spec.tsx
@@ -101,7 +101,7 @@ describe('ListContent', () => {
     expect(emptyMessage).toBeInTheDocument()
   })
 
-  it('should open delete confirmation modal when delete action is clicked', async () => {
+  it('should show confirmation modal and dispatch delete after confirming', async () => {
     const user = userEvent.setup({ pointerEventsCheck: 0 })
     renderComponent()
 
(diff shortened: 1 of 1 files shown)


redisinsight/ui/src/pages/vector-search/pages/VectorSearchListPage/components/list-content/ListContent.spec.tsx
```diff
@@ -102,20 +101,49 @@ describe('ListContent', () => {
...
+    await user.click(screen.getByText('Delete'))
 
-    const deleteBtn = screen.getByTestId('index-actions-delete-btn-test-index')
-    await userEvent.click(deleteBtn)
+    await user.click(screen.getByRole('button', { name: 'Delete index' }))
 
     expect(deleteRedisearchIndexAction).toHaveBeenCalled()
   })
 
+  it('should not dispatch deleteRedisearchIndexAction when cancel is clicked', async () => {
```
valkirilov, line 126: nit: Also, we can verify that the modal disappears after clicking this button.

redisinsight/ui/src/pages/vector-search/pages/VectorSearchListPage/components/list-content/ListContent.spec.tsx
```diff
@@ -102,20 +101,49 @@ describe('ListContent', () => {
...
+    )
+    await user.click(screen.getByText('Delete'))
+
+    expect(
+      screen.getByText('Are you sure you want to delete this index?'),
+    ).toBeInTheDocument()
+    expect(deleteRedisearchIndexAction).not.toHaveBeenCalled()
+  })
+
+  it('should dispatch deleteRedisearchIndexAction after confirmation', async () => {
```
valkirilov: nit: I would rather combine this test with the above one, where we actually click the button. 

And you can keep this check, but I think it will be more valuable to verify that the toast notification appeared on the screen, because this is the actual feedback message to the users. Otherwise, this dispatch of the Redux action is an implementation detail that users don't care about.

PR #5593, opened 2026-03-04
RI-8039 [E2E] 7.3 Settings -> Workbench Settings

pd-redis:
# What

Add Playwright E2E tests for the Workbench Settings section (7.3) of the Settings page.

Tests verify that the editor cleanup switch and pipeline commands setting are displayed when expanding the Workbench section. The "Configure command timeout" test case is N/A (per-database setting, not on the Settings page).

# Testing

Run: `npx playwright test tests/main/settings/workbench-settings`

---

Closes #RI-8039

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes shared `SettingsPage` locators/expanded-state detection, which could affect other Settings E2E tests if the new selectors don’t match all UI variants; otherwise changes are test-only.
> 
> **Overview**
> Adds new Playwright E2E coverage for **Settings → Workbench**: verifies the editor cleanup switch and pipeline command batch size control are visible, persist after navigation, and that editor cleanup behavior actually affects Workbench command execution.
> 
> Updates the `SettingsPage` object to better match the accordion UI (heading-based toggles, `data-state` checks), and adds scoped locators plus helper methods for editing/applying the pipeline commands value. Marks the corresponding Workbench settings cases as ✅ in `TEST_PLAN.md`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b4bd14650fc18b7e8118fb288ce705c460b40b0c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tests/e2e-playwright/tests/main/settings/workbench-settings/workbench-settings.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Note: "Configure command timeout" is N/A -- it's a per-database setting, not on the Settings page.
+ */
+test.describe('Workbench Settings', () => {
+  test.beforeEach(async ({ settingsPage }) => {
+    await settingsPage.goto();
+    await settingsPage.expandWorkbench();
+  });
+
+  test('should show editor cleanup switch', async ({ settingsPage }) => {
+    await expect(settingsPage.editorCleanupSwitch).toBeVisible();
```
pawelangelow, line 22: haven't tested it, but if it's easy, it makes sense to add a test that executes something in the workbench so we can validate the switch logic

tests/e2e-playwright/tests/main/settings/workbench-settings/workbench-settings.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  test('should show pipeline commands setting', async ({ settingsPage }) => {
+    await expect(settingsPage.pipelineCommandsText).toBeVisible();
```
pawelangelow, line 27: and for that one too (I imagine we can set it to 2, and validate just a single command is visible)

pd-redis: I don't think these should be tested here. The test must verify that you can change and save settings. How they affect the related feature, that should be tested as part of that feautre's tests

tests/e2e-playwright/tests/main/settings/workbench-settings/workbench-settings.spec.ts
pd-redis: added

PR #5638, opened 2026-03-13
RI-7944: Handle edge cases when making keys searchable from Browser Page

dantovska:
# What

Handle two edge cases for the "Make keys searchable from Browser" feature:
- **Redis < 6.0**: The `SCAN ... TYPE` filter is not supported. Added a proactive version check (`RedisFeature.ScanTypeFilter`) so the endpoint returns empty results immediately without issuing SCAN commands or producing error logs.

# Testing

Create this docker-compose file and run this redis instance with redisearch version 1.4.20 which is redis 5"

```
version: '3.8'

services:
  redis5-redisearch:
    image: redislabs/redisearch:1.4.20
    container_name: redis5-redisearch
    hostname: redis5-redisearch
    ports:
      - '6378:6379'
    volumes:
      - redis5-data:/data
    command:
      [
        'redis-server',
        '--appendonly',
        'yes',
        '--protected-mode',
        'no',
        '--loadmodule',
        '/usr/lib/redis/modules/redisearch.so',
      ]
    networks:
      redis5-net:
        ipv4_address: 172.22.0.2

networks:
  redis5-net:
    driver: bridge
    ipam:
      config:
        - subnet: 172.22.0.0/16

volumes:
  redis5-data:
```

### DIFF
--- a/redisinsight/api/src/modules/browser/keys/keys.service.spec.ts
+++ b/redisinsight/api/src/modules/browser/keys/keys.service.spec.ts
@@ -471,6 +471,7 @@ describe('KeysService', () => {
 
     beforeEach(() => {
       mockStandaloneRedisClient.sendPipeline.mockReset();
+      mockStandaloneRedisClient.isFeatureSupported.mockResolvedValue(true);
     });
 
     it('should return searchable key when hash key found', async () => {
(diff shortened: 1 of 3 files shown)


dantovska:
Solution is out of scope. Will be handled using different approach in a separate PR

PR #5648, opened 2026-03-17
RI-8138: Adjust landing page spacing

dantovska:
# What

Increased the spacing between the content and the outer edges on the Vector Search landing pages (WelcomeScreen and RqeNotAvailable) using responsive `clamp()` values that scale with the viewport while staying within comfortable bounds.
For RqeNotAvailable, also refactored the layout to follow the same Container/ScrollArea pattern used in WelcomeScreen — an outer container with `overflow: hidden` and an inner scrollable area that holds the padding. This fixes content clipping on smaller viewports.

| Before | After |
| --- | --- | 
| image | image |
| image | image |
| image | image |
| image | image |

# Testing

1. Connect to a database **with** RediSearch — navigate to the Search tab with no indexes and verify the WelcomeScreen has increased spacing from the edges
2. Connect to a database **without** RediSearch — navigate to the Search tab and verify the RqeNotAvailable page has matching spacing
3. Resize the window to minimum size (900x652) — content should remain fully accessible via scrolling, no clipping

---

> [!NOTE]
> **Low Risk**
> Low risk UI-only layout changes; main risk is unintended padding/scroll behavior differences on small viewports.
> 
> **Overview**
> **Adjusts Vector Search landing page spacing and scrolling behavior.** `WelcomeScreen` now uses responsive `clamp()` padding for larger, viewport-scaled gutters.
> 
> `RqeNotAvailable` is refactored to match the same *outer container + inner scroll area* pattern: the card becomes `overflow: hidden`, and a new scrollable wrapper owns the responsive `clamp()` padding to prevent content clipping on smaller screens.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2e9d23e08c71c79acb271bb3c55f5ae44c29c7b0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/pages/vector-search/components/rqe-not-available/RqeNotAvailable.tsx
+++ b/redisinsight/ui/src/pages/vector-search/components/rqe-not-available/RqeNotAvailable.tsx
@@ -30,91 +30,93 @@ export const RqeNotAvailable = () => {
 
   return (
     <S.StyledCard data-testid="rqe-not-available">
-      <S.StyledCardBody>
-        <S.ContentSection>
-          <Title
-            color="primary"
-            size="XL"
-            data-testid="rqe-not-available-title"
-          >
-            {rqeContent?.title?.[0]}
-          </Title>
+      <S.ScrollArea>
+        <S.StyledCardBody>
+          <S.ContentSection>
+            <Title
+              color="primary"
+              size="XL"
+              data-testid="rqe-not-available-title"
+            >
+              {rqeContent?.title?.[0]}
+            </Title>
 
-          <Spacer size="m" />
+            <Spacer size="m" />
 
-          <ColorText color="primary" variant="semiBold">
-            {rqeContent?.text?.[0]}
-          </ColorText>
+            <ColorText color="primary" variant="semiBold">
+              {rqeContent?.text?.[0]}
+            </ColorText>
 
-          <S.FeatureList data-testid="rqe-feature-list">
-            {rqeContent?.improvements?.map((improvement: string) => (
-              <S.FeatureListItem
-                key={improvement}
-                iconType="ToastCheckIcon"
-                color="primary"
-                label={<ColorText color="primary">{improvement}</ColorText>}
-              />
-            ))}
-          </S.FeatureList>
+            <S.FeatureList data-testid="rqe-feature-list">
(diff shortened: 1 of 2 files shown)


redisinsight/ui/src/pages/vector-search/components/welcome-screen/WelcomeScreen.styles.ts
```diff
@@ -14,7 +14,7 @@ export const ScrollArea = styled(Col)`
   position: relative;
   z-index: 1;
   overflow-y: auto;
-  padding: ${({ theme }) => theme.core.space.space300};
+  padding: clamp(2rem, 8vh, 9rem) clamp(7rem, 8vw, 14rem);
```
DimoHG, line 17: It means we are running away from the theme defaults and introducing custom value, right?

redisinsight/ui/src/pages/vector-search/components/rqe-not-available/RqeNotAvailable.tsx
```diff
@@ -30,91 +30,93 @@ export const RqeNotAvailable = () => {
 
   return (
     <S.StyledCard data-testid="rqe-not-available">
-      <S.StyledCardBody>
```
DimoHG, line 33: Seems like mostly reformatting stuff. Do we have a formatter set for the project :?

redisinsight/ui/src/pages/vector-search/components/welcome-screen/WelcomeScreen.styles.ts
dantovska: In this case I decided to do so, yes. We do not have spacing as big as 9 or 14 rem and I would have used it only for the 2rem, which would just be confusing IMO - it seems clearer this way. As for 9 and 14 rem, they are the closes to what is in the designs

redisinsight/ui/src/pages/vector-search/components/rqe-not-available/RqeNotAvailable.tsx
dantovska: seems like a lot of changes but it is actually only 2 rows - I have wrapped the whole section in another container and everything inside adds tabs.

PR #5678, opened 2026-03-20
RI-000: Add copy button to command results

pawelangelow:
# What

Adds a copy button to Workbench command results, allowing users to quickly copy output to the clipboard.

## Before

https://github.com/user-attachments/assets/04c8e657-3004-44f2-b998-0fd756bf3ed5

## After

https://github.com/user-attachments/assets/6226e937-5a63-4895-abd2-a467ce2e02ad

---

> [!NOTE]
> **Low Risk**
> Low risk UI-only change, but it introduces new result-to-text formatting (including group/db prefixes) that could affect what users copy from the workbench.
> 
> **Overview**
> Adds a **Copy result** button to the Workbench command results area (shown on hover) and adjusts spacing to accommodate it.
> 
> Extracts result formatting into a new `getResultText` helper used to generate the clipboard text for both single and grouped command outputs (including optional DB prefixing), and expands tests to cover the new formatting behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 282d785e85bafca9a9d47d13c4c7cfd7182e75c9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/components/query/query-card/QueryCardCliResultWrapper/QueryCardCliResultWrapper.spec.tsx
+++ b/redisinsight/ui/src/components/query/query-card/QueryCardCliResultWrapper/QueryCardCliResultWrapper.spec.tsx
@@ -242,7 +242,7 @@ describe('getResultText', () => {
     expect(text).toBe('"(nil)"')
   })
 
-  it('should format group results with command prefix', () => {
+  it('should format group results with command prefix (no db prefix for db 0)', () => {
     const result = [
       {
         response: [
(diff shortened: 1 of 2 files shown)


pawelangelow:
> Cool, for sure it will be handy 🙌🍕
> 
> Tip: I'm not sure what the correct approach is, but in the **Query Library,** we have a similar interface to copy the command, and there, we have the copy button in the top right corner.
> 
> So, in the sake of consistency - consider it as an option for **Command Results** as well 🤔
> 


I did, but for short commands, where your focus is more on the left, you kind of don't see the button. That's why I adjusted the space to have enough room for the button. Not perfect for sure, but something.

PR #5696, opened 2026-03-25
RI-8121: Add support for Vector Set key type

pawelangelow:
# What

Adds full-stack support for the new Redis **Vector Set** (`VSET`) key type, introduced in Redis 8.0. This includes backend key-info strategy, frontend type details component, and feature flag gating.

---

> [!NOTE]
> **Medium Risk**
> Adds a new Redis key type with new Redis command plumbing (`VCARD`) and key-info logic, which could affect key browsing and Redis client behavior if the command support differs across server versions. UI is gated behind a dev feature flag, reducing exposure but requiring validation of backend compatibility and size/length handling.
> 
> **Overview**
> Adds **feature-flagged support for the Redis 8.0 Vector Set** (`vectorset`) key type across API and UI.
> 
> Backend introduces `RedisDataType.VectorSet`, registers the `VCARD` command with the ioredis client, and adds a `VectorSetKeyInfoStrategy` wired into `KeyInfoProvider/KeysModule` (with tests) to fetch TTL/length and optionally memory usage.
> 
> Frontend adds `KeyTypes.VectorSet` display/color constants and a new `VectorSetDetails` panel, rendered only when the new `dev-vectorSet` flag is enabled (including a new selector and default flag config on both server and UI).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f8030b2769924264b987972655ce8100f9569bf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisinsight/api/src/modules/feature/constants/index.ts
+++ b/redisinsight/api/src/modules/feature/constants/index.ts
@@ -35,6 +35,7 @@ export enum KnownFeatures {
   AzureEntraId = 'azureEntraId',
   DevAzureEntraId = 'dev-azureEntraId',
   DevBrowser = 'dev-browser',
+  DevVectorSet = 'dev-vectorSet',
 }
 
 export interface IFeatureFlag {
--- a/redisinsight/api/src/modules/feature/constants/known-features.ts
+++ b/redisinsight/api/src/modules/feature/constants/known-features.ts
@@ -75,4 +75,8 @@ export const knownFeatures: Record<KnownFeatures, IFeatureFlag> = {
     name: KnownFeatures.DevBrowser,
     storage: FeatureStorage.Database,
   },
+  [KnownFeatures.DevVectorSet]: {
+    name: KnownFeatures.DevVectorSet,
+    storage: FeatureStorage.Database,
+  },
 };
--- a/redisinsight/api/src/modules/feature/providers/feature-flag/feature-flag.provider.ts
+++ b/redisinsight/api/src/modules/feature/providers/feature-flag/feature-flag.provider.ts
@@ -99,6 +99,10 @@ export class FeatureFlagProvider {
       KnownFeatures.DevBrowser,
       new CommonFlagStrategy(this.featuresConfigService, this.settingsService),
     );
+    this.strategies.set(
+      KnownFeatures.DevVectorSet,
+      new CommonFlagStrategy(this.featuresConfigService, this.settingsService),
+    );
   }
 
   getStrategy(name: string): FeatureFlagStrategy {


redisinsight/ui/src/styles/themes/light_theme/_theme_color.scss
```diff
@@ -114,6 +114,7 @@ $typeReJSONColor: #b8c5db;
 $typeStreamColor: #c7cea8;
 $typeGraphColor: #acccd7;
 $typeTimeSeriesColor: #c7c7c7;
+$typeVectorSetColor: #C7C7C7;
```
valkirilov, line 117: I see it's the same as the color that we already have for the Time Series, but to be honest, I'm not sure whether I like it on the light theme 😀

_PS: No need to make any changes to it_

Image

redisinsight/api/src/modules/browser/keys/key-info/strategies/vector-set.key-info.strategy.ts
```diff
@@ -0,0 +1,60 @@
...
+      };
+    }
+
+    const [[, ttl = null], [, length = null]] = (await client.sendPipeline([
+      [BrowserToolKeysCommands.Ttl, key],
+      [BrowserToolVectorSetCommands.VCard, key],
+    ])) as [any, number][];
+
+    let size = -1;
+    if (length < 50_000) {
```
valkirilov, line 45: nit: I see we already have this in 5 files, for the rest of the types (that's why it's replaced here as well, following the pattern we already have).

But I find this magic number to be easily exported in a shared constant, like `MAX_KEY_SIZE`. 
However, I'm no longer confident that this still makes sense in the age of AI 😀

redisinsight/api/src/modules/browser/keys/key-info/strategies/vector-set.key-info.strategy.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+  BrowserToolVectorSetCommands,
+} from 'src/modules/browser/constants/browser-tool-commands';
+import { ReplyError } from 'src/models';
+import {
+  GetKeyInfoResponse,
+  RedisDataType,
+} from 'src/modules/browser/keys/dto';
+import { VectorSetKeyInfoStrategy } from 'src/modules/browser/keys/key-info/strategies/vector-set.key-info.strategy';
+
+const getKeyInfoResponse: GetKeyInfoResponse = {
```
valkirilov, line 15: nit: Very minor, but it looks like a good candidate for a reusable factory (I feel we might need that in multiple tests later).

redisinsight/ui/src/styles/themes/light_theme/_theme_color.scss
dantovska: noted. looking at the new designs, we'll might have a black text which will affect all colors, so this will be revisited. thanks!

redisinsight/api/src/modules/browser/keys/key-info/strategies/vector-set.key-info.strategy.ts
dantovska: thanks, will be resolved in the next pr

redisinsight/api/src/modules/browser/keys/key-info/strategies/vector-set.key-info.strategy.spec.ts
dantovska: handling it in the next pr

PR #5816, opened 2026-04-22
Reconnect Redis clients on network interface change

valkirilov:
# What

Fixes a bug where the UI spun on a loading indicator after the host switched network interfaces (e.g. Ethernet → Wi-Fi). Clicking Refresh immediately after unplugging the cable would hang for the full HTTP request timeout before any error surfaced; subsequent clicks worked.

### Before

https://github.com/user-attachments/assets/a4c15914-a998-4a3c-a7ef-65bf051383d2

### After

https://github.com/user-attachments/assets/50869ed4-805c-478f-95fe-8c2a55a2cb14

## Root cause

Cached Redis clients kept TCP sockets that became half-open the moment the OS tore down the old interface. Writes on those sockets are silently queued by the kernel — ioredis/node-redis never see an error, so queued commands sit until something else times out.

## Fix

Three layers, backend-only:

1. **TCP keep-alive** — `redis_clients.keepAlive: 5000` is now passed to both ioredis and node-redis connection strategies. Gives the OS a chance to surface a dead socket on its own.
2. **NetworkChangeMonitor** (`redisinsight/api/src/modules/redis/network-change.monitor.ts`) — polls `os.networkInterfaces()` every 2 s, computes a stable signature (excluding loopback + IPv6 link-local), and when the interface set changes, calls `RedisClientStorage.removeAll()`. The next request opens a fresh socket on the currently-active interface. Cross-platform, no native deps, tunable via `RI_CLIENTS_NETWORK_WATCHER*` env vars.
3. **424 error mapping** — `catchAclError` now converts runtime driver errors (`EADDRNOTAVAIL`, `ECONNRESET`, `Socket closed unexpectedly`, reconnect backoff exhausted, …) into `RedisConnection*` exceptions (HTTP 424), which the existing UI `connectivityErrorsInterceptor` already handles to show the "Connection Lost" banner instead of a generic 500.

Existing HTTP `requestTimeout` (25 s) remains the single per-request cap.

## Testing

1. Connect to a Redis DB via wired connection, browse keys → data loads.
2. Unplug the cable, switch to Wi-Fi (with the Redis host reachable on the new network).
3. Immediately click Refresh → should return data without a hang (previously spun for ~25 s).
4. Same with keep-alive disabled (`RI_CLIENTS_KEEP_ALIVE=0`) → still works because the monitor handles it.
5. Disable the monitor (`RI_CLIENTS_NETWORK_WATCHER=false`) → reverts to old behavior.

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Medium Risk**
> Touches Redis connection setup and runtime error handling, which can affect connectivity and user-facing error behavior if misclassified; also introduces a background polling monitor that clears all cached clients on interface churn.
> 
> **Overview**
> Prevents hangs after host network switches by **enabling TCP keepalive** for both ioredis and node-redis clients (new `redis_clients.keepAlive`, default `5000`, configurable via `RI_CLIENTS_KEEP_ALIVE`).
> 
> Adds a **`NetworkChangeMonitor`** (configurable via `RI_CLIENTS_NETWORK_WATCHER*`) that polls `os.networkInterfaces()` and calls `RedisClientStorage.removeAll()` when the routable interface set changes, forcing fresh sockets on next use.
> 
> Updates error handling so **runtime driver/socket disconnects** are recognized and re-thrown as `RedisConnection*` (HTTP 424) via `wrapRedisRuntimeConnectionError`, improving UI handling; includes new/expanded unit tests for keepalive passthrough, `removeAll`, interface-change monitoring, and runtime error mapping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 555da668f3e447e8e1bb54e652b3b7fde8b9cdcd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisinsight/api/config/default.ts
+++ b/redisinsight/api/config/default.ts
@@ -178,10 +178,26 @@ export default {
     maxRedirections: parseInt(process.env.RI_CLIENTS_MAX_REDIRECTIONS, 10) || 3,
     slotsRefreshTimeout:
       parseInt(process.env.RI_CLIENTS_SLOTS_REQUEST_TIMEOUT, 10) || 5000,
+    // TCP keepalive initial idle delay in ms. Enables SO_KEEPALIVE so that
+    // half-open sockets (e.g. after a network interface switch) are eventually
+    // detected and closed by the OS instead of hanging indefinitely.
+    // Set to 0 to disable.
+    keepAlive:
+      process.env.RI_CLIENTS_KEEP_ALIVE !== undefined
+        ? parseInt(process.env.RI_CLIENTS_KEEP_ALIVE, 10)
+        : 5_000,
     maxStringSize: parseInt(process.env.RI_CLIENTS_MAX_STRING_SIZE, 10),
     truncatedStringPrefix:
       process.env.RI_CLIENTS_TRUNCATED_STRING_PREFIX ||
       '[Truncated due to length]',
+    // When enabled, the backend watches the host's network interfaces and
+    // disconnects every cached Redis client whenever the interface set
+    // changes (e.g. Ethernet -> Wi-Fi). The next command then opens a fresh
+    // socket on the currently-active interface instead of reusing a
+    // half-open one and hanging until the HTTP requestTimeout fires.
+    networkWatcher: process.env.RI_CLIENTS_NETWORK_WATCHER !== 'false',
+    networkWatcherInterval:
+      parseInt(process.env.RI_CLIENTS_NETWORK_WATCHER_INTERVAL, 10) || 2_000,
   },
   redis_scan: {
     countDefault: parseInt(process.env.RI_SCAN_COUNT_DEFAULT, 10) || 200,
--- a/redisinsight/api/src/modules/redis/connection/ioredis.redis.connection.strategy.spec.ts
+++ b/redisinsight/api/src/modules/redis/connection/ioredis.redis.connection.strategy.spec.ts
@@ -230,6 +230,20 @@ describe('IoredisRedisConnectionStrategy', () => {
 
       process.nextTick(() => mockIoredisNativeClient.emit('ready'));
     });
+    it('should pass keepAlive to detect half-open sockets after network switch', (done) => {
+      service
+        .createStandaloneClient(mockClientMetadata, mockDatabase, {})
+        .then(() => {
+          expect(spyRedis).toHaveBeenCalledWith(
+            expect.objectContaining({
+              keepAlive: REDIS_CLIENTS_CONFIG.keepAlive,
+            }),
+          );
+          done();
+        });
+
+      process.nextTick(() => mockIoredisNativeClient.emit('ready'));
+    });
   });
 
   describe('createClusterClient', () => {
--- a/redisinsight/api/src/modules/redis/connection/ioredis.redis.connection.strategy.ts
+++ b/redisinsight/api/src/modules/redis/connection/ioredis.redis.connection.strategy.ts
@@ -54,6 +54,11 @@ export class IoredisRedisConnectionStrategy extends RedisConnectionStrategy {
       password,
       family: 0, // Enable dual-stack IPv4/IPv6 (auto-detect)
       connectTimeout: timeout,
+      // TCP keepalive. Required to detect half-open sockets after the OS
+      // silently drops a network interface (e.g. Ethernet -> Wi-Fi hand-off).
+      // Without this, ioredis would never surface an error on the pending
+      // command and requests would hang forever.
+      keepAlive: REDIS_CLIENTS_CONFIG.keepAlive,
       db: isNumber(clientMetadata.db) ? clientMetadata.db : db,
       connectionName:
         options?.connectionName ||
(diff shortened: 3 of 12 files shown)


ArtemHoruzhenko:
Not sure this is a proper fix. I'm not sure we must monitor for network interface change. It looks like workaround. I feel like proper solution is `connection lost` + `retry`. Meaning that we must better (earlier) understand that connection was lost and then retry. I think all this could be configured in the ioredis conf. Not sure. But if it is possible to go this way we might cover all the cases with connection lost, not only when interface changed.

valkirilov:
> Not sure this is a proper fix. I'm not sure we must monitor for network interface change. It looks like workaround. I feel like proper solution is `connection lost` + `retry`. Meaning that we must better (earlier) understand that connection was lost and then retry. I think all this could be configured in the ioredis conf. Not sure. But if it is possible to go this way we might cover all the cases with connection lost, not only when interface changed.

Good push-back @ArtemHoruzhenko :) 

The short answer - there's no `ioredis` setting that detects this. When the OS drops the old interface, the TCP socket stays in `ESTABLISHED` state, writes succeed into the kernel buffer, and `ioredis` never sees an error - so there's nothing for it to time out against. 

TCP keep-alive (which this PR enables) is the closest built-in signal, but Node only exposes the initial delay, not the probe interval (default ~75 s between probes), so detection takes over a minute. 

So the real options are:

1. What this PR does - watch `os.networkInterfaces()` (a ~0.1 ms in-memory kernel read every 2s) and drop cached clients when the interface set changes. Fixes the reported bug exactly, doesn't help if the server dies with no interface change.

2. Application-level heartbeat PING on idle connections, force reconnect on failure. Covers all "connection lost" cases (what you're asking for) but adds ping traffic, tuning, and false-positive reconnects on transient server blips. It's a bigger change and deserves its own PR.

They're complementary - (1) is a cheap fast-path, (2) is the general safety net.

dantovska (dismissed):

ArtemHoruzhenko:
> > Not sure this is a proper fix. I'm not sure we must monitor for network interface change. It looks like workaround. I feel like proper solution is `connection lost` + `retry`. Meaning that we must better (earlier) understand that connection was lost and then retry. I think all this could be configured in the ioredis conf. Not sure. But if it is possible to go this way we might cover all the cases with connection lost, not only when interface changed.
> 
> Good push-back @ArtemHoruzhenko :) 
> 
> The short answer - there's no `ioredis` setting that detects this. When the OS drops the old interface, the TCP socket stays in `ESTABLISHED` state, writes succeed into the kernel buffer, and `ioredis` never sees an error - so there's nothing for it to time out against. 
> 
> TCP keep-alive (which this PR enables) is the closest built-in signal, but Node only exposes the initial delay, not the probe interval (default ~75 s between probes), so detection takes over a minute. 
> 
> So the real options are:
> 
> 1. What this PR does - watch `os.networkInterfaces()` (a ~0.1 ms in-memory kernel read every 2s) and drop cached clients when the interface set changes. Fixes the reported bug exactly, doesn't help if the server dies with no interface change.
> 
> 2. Application-level heartbeat PING on idle connections, force reconnect on failure. Covers all "connection lost" cases (what you're asking for) but adds ping traffic, tuning, and false-positive reconnects on transient server blips. It's a bigger change and deserves its own PR.
> 
> They're complementary - (1) is a cheap fast-path, (2) is the general safety net.

Can't user actions be a ping-like activity? We don't need to introduce another mechanism.

ArtemHoruzhenko (dismissed):

valkirilov:
> Can't user actions be a ping-like activity? We don't need to introduce another mechanism.

Yup, I hear what you say, and in theory it makes sense. However, the problem is that in this bug, the user action is the "ping" - and that's exactly what hangs for 25 s.

Walking through the failure:

1. User unplugs **Ethernet**.
2. User clicks **Refresh** → `ioredis` writes the command to the cached socket.
3. The socket still looks `ESTABLISHED` to the kernel, so the write succeeds into the send buffer. No error is returned.
4. The packet is queued for retransmission on a dead route and never acked.
5. `ioredis` has no way to time this out, so it waits. The HTTP `requestTimeout` fires at 25 s.
6. Now `ioredis` knows the socket is bad, reconnects, and the next click works.

So "use the user's action as the ping" is literally the current behavior - and it's what produces the 25 s hang. It only works as a detection mechanism if the request itself returns an error quickly, but on a half-open socket it doesn't - the kernel swallows the write silently.

For user actions to be a viable signal, we'd need either:
- A short per-command timeout (`commandTimeout` ~2–3 s) so a hung write surfaces quickly - but that breaks long-running commands in CLI / Workbench / Profiler, and I don't think it's OK
- or something proactive that runs before the user clicks, so the stale client is already gone. That's what `NetworkChangeMonitor` does (via a free kernel signal) and what a heartbeat PING would do (via a Redis round-trip).

I was also not 100% confident with this change, so we can postpone it and discuss more, or rework it 🙂

valkirilov:
Closing this one in favor of a minimal change https://github.com/redis/RedisInsight/pull/5848.

Instead of adding this "real-time" polling, let's leave it to `ioredis`/`node-redis` to handle the lifecycle of the TCP connection.

The trade-off is that after unplugging the cable, the user has to wait up to ~10 s before the first **Refresh** click succeeds (vs sub-second with the monitor approach). 

But I think, that is acceptable for our case because nobody realistically unplugs the cable and instantly hits Refresh. By the time they reach for the mouse the socket has already failed and the next click opens a fresh connection.

Issue #5637, opened 2026-03-13
[Feature Request]: Make linux arm releases available


rad-pat:
I'm currently struggling to install on Ubuntu ARM64 (VM on MacBook). No snaps are available for my architecture and I can't see anything in the release assets either. Is it easy to add this to the set of items built?

I tried to build from source also, but ran into problems with that.

DimoHG:
Hey @rad-pat good point. We'll take it with the team. Expect updates soon

rad-pat:
Hi @DimoHG , I see a PR was merged for this and a version updated to 3.4.0, but no release yet. Is one to be expected soon? I would love to try it out.

DimoHG:
Yep, should be available sometime next week :)

valkirilov:
We have released a new version of Redis Insight [v3.4.1](https://github.com/redis/RedisInsight/releases/tag/3.4.1), and the issue should be gone now. You can give it a try, and in case you find any other issues, feel free to reopen it and let us know 👍 

Thanks again for reporting it, and sorry for the delay 🙏

rad-pat:
Hi @valkirilov, I installed v3.4.2 - it seems to install, but not GUI appears. What logs can be provided to help understand the problem?

valkirilov:
Sorry to hear that @rad-pat 😞

Let's try to debug it together:
1. What OS do you use?
2. Which package of Redis Insight - `.deb`, `.snap`, `.AppImage`?
3. You can just start the app through your Terminal and see the logs, which will help us identify the issue

_PS: If you use Ubuntu + Wayland, there is a known Electron [issue](https://github.com/electron-userland/electron-builder/issues/9452), and you can temporarily bypass it by passing `--ozone-platform=x11` when you start the app._

rad-pat:
Great, thanks.

I'm trying to run on Ubuntu 26.04 in a VM, on arm64
I'm installing via the .deb because no snap is available for arm64

These are the logs from the terminal, but no GUI appears:
```console
me@my-vm:~$ redisinsight
08:49:59.203 › [Server] Launching API server
(node:4927) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `redisinsight --trace-deprecation ...` to show where the warning was created)
08:49:59.211 › [Server] Starting production server with port: 5530
08:49:59.212 › [Server] Environment: production

```

rad-pat:
Also, from syslog, if any use:
[...]

PR #5836, opened 2026-04-28
Fix Linux ARM native binary packaging

valkirilov:
# What

- Refactored Linux packaging into a single matrix job (x64, arm64) in `pipeline-build-linux.yml`, removing duplicated job logic while keeping arch-specific behavior in matrix fields.
- Added matrix target resolution (resolve-targets) to decide per-arch execution (default for all/linux, custom for `build_linux_*`) and avoid false build/verify runs.
- Kept ARM-specific RPM handling in the ARM matrix leg (`USE_SYSTEM_FPM` + system fpm) to avoid electron-builder x86 FPM exec-format failures on ARM.
- Added deterministic native addon checks by running `verify-native-modules-linux.js` only when an arch actually built packages.
- Preserved artifact compatibility: matrix legs upload linux-builds-{arch} / vendor-plugins-{arch}, then publish-artifacts merges and republishes legacy artifact names (linux-builds, vendor-plugins) used by downstream workflows.
- Kept expected Linux filename variants in uploads (x86_64/amd64 and arm64/aarch64) so release/upload pipelines continue finding the same deliverables.

# Why

We split Linux packaging into separate x64 and arm64 jobs because our ARM desktop artifacts are currently unreliable - the app starts, but DB initialization fails at runtime with:

```
better_sqlite3.node: cannot open shared object file: No such file or directory
```

## Why this happens
After migrating from `sqlite3` to `better-sqlite3`, we now depend on native `.node` binaries being present inside `app.asar.unpacked`.

When Linux artifacts for multiple architectures are built together on an x64 runner, native addon resolution/rebuild/packaging can produce incomplete ARM outputs, and the required ARM native binary may be missing from the final package.

## Why splitting by architecture helps
- x64 artifacts are built on x64 runner
- arm64 artifacts are built on native ARM runner

This aligns build environment with target architecture and reduces cross-arch native module packaging issues.

## Additional safety in this PR
We also added deterministic post-package verification that fails CI if required native binaries are missing from packaged output (at least `better-sqlite3` and `keytar`).

That moves failure from customer runtime to CI time.

## Important non-goal
This change does not alter published artifact names/consumption contracts.
Final linux-builds artifact content remains compatible with existing release/upload/manual pipelines and external links.

# Testing

Verified via a manual build on both x86 and ARM

https://github.com/redis/RedisInsight/actions/runs/25048225683

---

References  #5637

---

> [!NOTE]
> **Medium Risk**
> Changes CI packaging and artifact publication for Linux, which can break releases if matrix target resolution or artifact merging is incorrect. No production runtime code changes, but build outputs and caches are affected across architectures.
> 
> **Overview**
> Fixes unreliable Linux ARM packaging by splitting the Linux build workflow into an `x64`/`arm64` matrix that runs ARM builds on native ARM runners, with arch-specific target selection, artifact naming, and an extra `publish-artifacts` job to merge outputs back into the existing `linux-builds`/`vendor-plugins` artifacts.
> 
> Adds deterministic post-build validation (`scripts/verify-native-modules-linux.js` + `scripts/native-modules.config.json`) to fail CI if required native `.node` binaries (e.g., `better-sqlite3`, `keytar`) are missing from `app.asar.unpacked`, and updates the dependency cache key to include `runner.arch` to avoid cross-arch `node_modules` reuse.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1ca81bff3cb36ae6ea2f13fcbc4bc78381fb4424. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/actions/install-all-build-libs/action.yml
+++ b/.github/actions/install-all-build-libs/action.yml
@@ -34,9 +34,9 @@ runs:
           node_modules
           redisinsight/node_modules
           redisinsight/api/node_modules
-        key: node-modules-${{ runner.os }}-${{ hashFiles('yarn.lock', 'redisinsight/yarn.lock', 'redisinsight/api/yarn.lock') }}
+        key: node-modules-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('yarn.lock', 'redisinsight/yarn.lock', 'redisinsight/api/yarn.lock') }}
         restore-keys: |
-          node-modules-${{ runner.os }}-
+          node-modules-${{ runner.os }}-${{ runner.arch }}-
 
     - name: Setup Python
       uses: actions/setup-python@v5
--- a/.github/workflows/pipeline-build-linux.yml
+++ b/.github/workflows/pipeline-build-linux.yml
@@ -25,136 +25,49 @@ on:
         type: boolean
 
 jobs:
-  build-x64:
-    name: Build linux x64
+  build-linux:
+    name: Build linux ${{ matrix.arch }}
     if: |
       inputs.target == 'all' ||
       inputs.target == 'linux' ||
-      contains(inputs.target, 'build_linux_appimage_x64') ||
-      contains(inputs.target, 'build_linux_deb_x64') ||
-      contains(inputs.target, 'build_linux_rpm_x64') ||
-      contains(inputs.target, 'build_linux_snap_x64')
-    runs-on: ubuntu-24.04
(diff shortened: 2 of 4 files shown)


ArtemHoruzhenko (commented):
good

ArtemHoruzhenko (approved):
lgtm

PR #5867, opened 2026-05-08
RI-8159: [BE] vector set similarity

dantovska:
# What

Add two backend endpoints for the vector-set similarity-search feature:
- `POST /vector-set/similarity-search` — runs `VSIM` against the supplied key and returns the matches (`name` + `score` + optional `attributes`).
- `POST /vector-set/similarity-search/preview` — returns the same `VSIM` invocation as a CLI-friendly string so the FE can show users the command they're about to run.

Both endpoints accept a single `SimilaritySearchDto` and pick the query mode (`ELE` / `VALUES` / `FP32`) at runtime — exactly one of `elementName`, `vectorValues`, or `vectorFp32` must be supplied. `WITHSCORES` and `WITHATTRIBS` are always appended so the response shape is stable; `COUNT` / `FILTER` are forwarded only when present.
The command builder, preview formatter, reply parser, and shared `VsimTokenWriter` strategy live in `vector-set.utils.ts` so the executable command and its preview can't drift on clause order or rendering. `VSIM` is also registered as a built-in ioredis command.

# Testing

---

> [!NOTE]
> **Medium Risk**
> Adds new API endpoints that execute Redis `VSIM` and parse its flat reply format, so correctness depends on command construction and reply parsing across RESP2/RESP3 types. Risk is mitigated by extensive unit tests but touches runtime Redis command execution.
> 
> **Overview**
> Adds vector-set similarity search support via `POST /vector-set/similarity-search` (executes `VSIM` with mandatory `WITHSCORES WITHATTRIBS` and optional `COUNT`/`FILTER`) and `POST /vector-set/similarity-search/preview` (returns a CLI-safe command string preview).
> 
> Introduces new DTOs/responses for search results and preview, registers `VSIM` as a built-in ioredis command, and centralizes VSIM command building/preview formatting/reply parsing in `vector-set.utils.ts` with tests covering query-mode validation (`ELE`/`VALUES`/`FP32`), quoting/escaping, and score/attributes decoding.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da7c9718f9f41da05038d3b657cfe091c030e970. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisinsight/api/bruno/RedisInsight/Vector Set/Similarity Search Preview.bru
+++ b/redisinsight/api/bruno/RedisInsight/Vector Set/Similarity Search Preview.bru
@@ -27,20 +27,18 @@ docs {
 
   ## Request Body
 
-  Same shape as `POST /vector-set/similarity-search`. The only difference is that all query fields are optional — when none is supplied the endpoint returns an empty preview instead of a `400`.
+  Same shape as `POST /vector-set/similarity-search` — exactly one of `elementName`, `vectorValues`, `vectorFp32` must be supplied.
 
   | Field | Type | Description |
   |-------|------|-------------|
   | `keyName` | string | The Vector Set key name. Required (`@IsDefined` on the DTO). |
-  | `elementName` | string | (Optional) Existing element's vector as the query |
-  | `vectorValues` | number[] | (Optional) Query vector as numeric values |
-  | `vectorFp32` | string | (Optional) Query vector as a base64-encoded little-endian FP32 blob |
+  | `elementName` | string | (One-of) Existing element's vector as the query |
+  | `vectorValues` | number[] | (One-of) Query vector as numeric values |
+  | `vectorFp32` | string | (One-of) Query vector as a base64-encoded little-endian FP32 blob |
   | `count` | number | Maximum number of results (omitted from preview when undefined) |
   | `filter` | string | Optional filter expression evaluated against element attributes |
 
-  At most one of `elementName`, `vectorValues`, `vectorFp32` may be present. Supplying more than one returns `400`.
-
-  When **none** of `elementName` / `vectorValues` / `vectorFp32` is supplied the response `preview` is an empty string — the FE treats that as "nothing to show yet" and hides the preview bar.
+  Exactly one of `elementName`, `vectorValues`, `vectorFp32` must be present — zero or more than one returns `400` (same rule as the search endpoint).
 
   ## Response
 
@@ -54,24 +52,9 @@ docs {
 
   | Status | Condition |
   |--------|-----------|
-  | `400` | More than one of `elementName` / `vectorValues` / `vectorFp32` supplied, invalid `count`, or invalid `vectorFp32` base64 |
+  | `400` | Zero or more than one of `elementName` / `vectorValues` / `vectorFp32` supplied, invalid `count`, or invalid `vectorFp32` base64 |
 
   ## Examples
-
-  ### Empty preview when no query payload is supplied
-
-  ```json
-  {
-    "keyName": "vset:small_3d",
-    "count": 10
-  }
-  ```
-
-  Returns:
-
-  ```json
-  { "preview": "" }
-  ```
 
   ### Preview for FP32 input
 
@@ -92,7 +75,7 @@ docs {
   ## Notes
 
   - `WITHSCORES` and `WITHATTRIBS` are always rendered (matches the search endpoint).
-  - When **no** query field is supplied, the response `preview` is an empty string instead of a placeholder command.
+  - The FE is expected to gate the call so the preview endpoint is only hit once the form has resolved to exactly one query mode.
 }
 
 settings {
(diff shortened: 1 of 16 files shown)


redisinsight/api/src/modules/browser/vector-set/__tests__/vector-set.factory.ts
```diff
@@ -132,3 +136,143 @@ export const getVectorSetElementsResponseFactory =
...
+export const searchVectorSetByElementDtoFactory =
+  Factory.define<SimilaritySearchDto>(() => ({
+    keyName: Buffer.from(`vset:${faker.string.alphanumeric(6)}`),
+    elementName: Buffer.from(faker.string.alphanumeric(8)),
+    vectorValues: undefined,
+    vectorFp32: undefined,
+    count: faker.number.int({ min: 1, max: 100 }),
+  }));
+
+export const searchVectorSetByValuesDtoFactory =
```
valkirilov: I don't think we need to define 3 separate factories here - it's really a single data type with variations on its data (whether it carries `vectorValues`, `vectorFp32`, or something else).

I know it sounds a bit more complex to define a single factory and toggle which fields get prefilled internally (via transient params), but the entity itself is a single one - a vector set - and if it changes at some point, we'd only need to update one factory instead of keeping all the variants in sync.

Here's a quick sketch of what I mean:

```ts
interface VectorSetTransientParams {
  variant: VectorVariant;
}

export const vectorSetFactory = Factory.define<VectorSet, VectorSetTransientParams>(
  ({ transientParams }) => {
    const { variant = 'none' } = transientParams;

    return {
      keyName: Buffer.from(`vset:${faker.string.alphanumeric(6)}`),
      elementName: Buffer.from(faker.string.alphanumeric(8)),
      vectorValues: variant === 'values' ? [/* fill defaults */] : undefined,
      vectorFp32:   variant === 'fp32'   ? Buffer.from(/* ... */) : undefined,
      count: faker.number.int({ min: 1, max: 100 }),
    };
  },
);

// Usage:
vectorSetFactory.build({}, { transient: { variant: 'values' } });  // values variant
vectorSetFactory.build({}, { transient: { variant: 'fp32' } });    // fp32 variant
```

And if a specific test needs custom values, regular params still take precedence over the transient defaults:

```ts
vectorSetFactory.build(
  { vectorFp32: Buffer.from([/* ... */]) },
  { transient: { variant: 'fp32' } },
);
```

Docs for reference:
- Fishery README — Transient params section: https://github.com/thoughtbot/fishery#transient-params

Happy to pair on it if it'd be easier - totally fine to keep things as-is too if you'd rather not expand scope on this PR!

dantovska: thanks! not at all, seems reasonable and it is in scope anyway

redisinsight/api/bruno/RedisInsight/Vector Set/Similarity Search Preview.bru
```diff
@@ -0,0 +1,100 @@
...
+  | Field | Type | Description |
+  |-------|------|-------------|
+  | `keyName` | string | The Vector Set key name. Required (`@IsDefined` on the DTO). |
+  | `elementName` | string | (Optional) Existing element's vector as the query |
+  | `vectorValues` | number[] | (Optional) Query vector as numeric values |
+  | `vectorFp32` | string | (Optional) Query vector as a base64-encoded little-endian FP32 blob |
+  | `count` | number | Maximum number of results (omitted from preview when undefined) |
+  | `filter` | string | Optional filter expression evaluated against element attributes |
+
+  At most one of `elementName`, `vectorValues`, `vectorFp32` may be present. Supplying more than one returns `400`.
```
valkirilov: Note: As we discussed offline, you can add examples for the various ways of using this API endpoint, but it can happen in a separate PR :)

redisinsight/api/src/modules/browser/vector-set/vector-set.controller.ts
```diff
@@ -189,4 +192,52 @@ export class VectorSetController extends BrowserBaseController {
...
+  ): Promise<SearchVectorSetResponse> {
+    return await this.vectorSetService.similaritySearch(clientMetadata, dto);
+  }
+
+  @Post('/similarity-search/preview')
+  @ApiRedisInstanceOperation({
+    description:
+      'Build a human-readable preview of the VSIM command that the similarity-search endpoint would execute for the supplied DTO. ' +
+      'Reuses the same internal command builder as the search endpoint so the preview cannot drift from what is actually executed. ' +
+      'Returns an empty `preview` string when no query payload (`elementName` / `vectorValues` / `vectorFp32`) is supplied.',
```
valkirilov: nit: Why don't we enforce the same validation rules as for the other endpoints - to have at least one correct param passed, since the other validation (when you pass more than a single param) is already in place?

valkirilov (dismissed):

redisinsight/api/bruno/RedisInsight/Vector Set/Similarity Search Preview.bru
dantovska: good note. I'll tackle this in a separate PR

dantovska: thanks!

valkirilov (approved):

PR #5863, opened 2026-05-07
RI-7682 Strengthen API public swagger docs and setup a postinstall script to autogenerate types for clients

KrumTy:
# What

Phase 1 of decoupling the UI from the API codebase (RI-7682).

- Scaffolds an OpenAPI / `typescript-axios` client generation pipeline (`yarn generate:api-client`), regenerated on `postinstall` and gitignored.
- Decorates all backend modules currently consumed by the UI with `@nestjs/swagger` so the generated client emits real types instead of `any`/`void`. Strict spec validation is enabled.
- Generated client lives at `redisinsight/api-client/` and is exposed to UI / Storybook / Jest via the `apiClient` path alias.
- Adds a warn-only ESLint rule discouraging new `apiSrc/*` imports from the UI, pointing devs at `apiClient/*` (RI-7682).

No UI imports are migrated in this PR — that follows per-domain in Phase 2.

# Testing

- `yarn install` (or `yarn generate:api-client`) produces `redisinsight/api-client/` with no validation errors.

---

> [!NOTE]
> **Medium Risk**
> Adds a new OpenAPI spec dump + client generation step that runs on API `postinstall`, which can affect install/build workflows and CI caching. Broad Swagger annotation changes across many endpoints may subtly alter the published API schema and generated client types.
> 
> **Overview**
> Adds a **generated OpenAPI TypeScript client pipeline**: the API now dumps `openapi.json` via a lightweight Nest bootstrap (`dump-openapi.ts`) and runs `@hey-api/openapi-ts` to output `redisinsight/api-client/` (gitignored), exposed via the new `apiClient/*` alias in TS, Vite/Storybook, and Jest.
> 
> Strengthens **Swagger/OpenAPI typing** across multiple API modules by standardizing Redis string schemas (`ApiRedisString`), fixing/adding response metadata (`@ApiOkResponse`, correct `isArray`/`type`), and improving union/discriminator schemas (e.g., REJSON, RDI statistics). Also adds a warn-only ESLint `no-restricted-imports` rule to discourage new UI imports from `apiSrc/*` in favor of `apiClient/*`, and updates test Docker installs to skip client generation when needed (`SKIP_API_CLIENT_GEN=1`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f087339fbf31be236d56400977c571283aca122. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #5884, opened 2026-05-11
RI-7682 Consume auto-generated apiClient types in the UI

KrumTy:
# What

Phase 2 of RI-7682 — replace every UI import from `apiSrc/*` / `src/*` (BE) with the auto-generated `apiClient` types produced by the postinstall script that landed in #5863. Flip the `no-restricted-imports` lint rule for those patterns from `warn` to `error` so new BE -> FE coupling fails CI.

Where the OpenAPI surface doesn't cover the type — socket.io payloads (`IMonitorData`, `PubSubMessage`, `SubscriptionDto`), the `IBulkActionOverview` family that the bulk-import controller declares as `type: Object` in swagger, and the Azure OAuth callback values that ship via `localStorage` rather than HTTP — define FE-local equivalents inline in their existing `slices/interfaces/*` files. Two stale imports that pointed at BE paths that no longer exist got fixed in passing: `AddSentinelMasterResponse` → `CreateSentinelDatabaseResponse`, `GetCloudAccountShortInfoResponse` → `CloudAccountInfo`.

Three of the FE-local enum mirrors *are* on the REST surface and were straightforward to promote — `ListElementDestination`, `NodeRole`, `HealthStatus` now come from `apiClient`. To make that work, added `enumName: 'X'` on those 4 `@ApiProperty({ enum })` decorator sites (Nest emits the enum as a `components/schemas/X` reference instead of inlining the union) and configured `@hey-api/typescript` to emit `enums: 'typescript'` with `case: 'PascalCase'`, so call sites keep reading `NodeRole.Primary` rather than the default `SCREAMING_SNAKE_CASE`.

# Testing

- `grep` confirms zero remaining `apiSrc/*` / `src/*` imports under `redisinsight/ui`. A probe import errors out, confirming the guard.
- `yarn lint` passes.
- No CI gate runs `tsc` on the UI (verified — `lint.yml` runs eslint only; tests use `babel-jest` which strips types; the `type-check:ui` script is never invoked from a workflow). Build/test behavior is unchanged.
- `tsc -p tsconfig.json` for context: main has 2498 errors, this branch has 2160 (−338 net). Of those, 795 main-only errors went away, 457 are new. The 457 are real BE↔FE shape-drift findings now visible after the type swap (e.g. `RdiInstance.lastConnection` is required in swagger but optional in BE class; `CommandExecutionResult.sizeLimitExceeded` similar; `GetUserAgreementsResponse` is missing fields that exist at runtime; factories use `faker.date.past()` where the wire format is `string`). None of these is a functional regression — type-only — and a follow-up PR will tighten the swagger decorators / fixtures to address them.

---

Refs RI-7682


KrumTy:
Reopening with a branch name that matches the project's GitHub Actions enforce-branch-name-rules convention (`feature/<ticket>/<title>`). Continued at the new PR.

PR #5911, opened 2026-05-14
RI-8179 Add ACL CAT dangerous commands service

KrumTy:
# What

New `GET /databases/:id/dangerous-commands` endpoint that returns the list of dangerous commands for the connected Redis instance, sourced from `ACL CAT dangerous`.

- Results are upper-cased and de-duplicated for FE consumption.
- Cached per database id; invalidated when the database instance is deleted (via `AppRedisInstanceEvents.Deleted`).
- Degrades to an empty list on Redis versions that do not support ACL — never throws.

Part of epic [RI-6594](https://redislabs.atlassian.net/browse/RI-6594) (Prod vs non-prod modes). Provides the authoritative dangerous-commands list that FE prod-mode guards will consume.

# Testing

- Unit tests for the provider (happy path, dedupe, empty reply, ACL unsupported, non-array reply, cache hit, invalidate, on-delete handler).
- Updated `database-info.service` spec to cover the new method.
- Integration test asserting the endpoint returns an array of upper-case strings and 404s for unknown ids.

Manual:
1. Start the API against a Redis 6.0+ instance, hit `GET /databases/{id}/dangerous-commands` — expect upper-cased command names.
2. Hit it against Redis < 6.0 — expect `200 []`.
3. Delete the database and call again on a new instance — expect a fresh fetch (no stale cache).

---

Closes RI-8179

[RI-6594]: https://redislabs.atlassian.net/browse/RI-6594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Adds a new database API endpoint that issues `ACL CAT dangerous` against Redis and caches results per database id; correctness depends on Redis error parsing and cache invalidation on instance deletion.
> 
> **Overview**
> Introduces a new `GET /databases/:id/dangerous-commands` endpoint that returns an upper-cased, de-duplicated list of dangerous Redis commands sourced from `ACL CAT dangerous`.
> 
> Implements a `DangerousCommandsProvider` with per-`databaseId` caching, cache invalidation on `AppRedisInstanceEvents.Deleted`, and error handling that caches permanent failures (ACL unsupported / `NOPERM`) as `[]` while treating other failures as transient.
> 
> Wires the provider through `DatabaseInfoService`/`DatabaseModule` and adds unit + integration tests covering fetch/normalization, caching/invalidation, and the new endpoint’s response/404 behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a8ff01976983ffe107fa525f0aa1132f31f95279. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisinsight/api/src/modules/database/providers/dangerous-commands.provider.ts
+++ b/redisinsight/api/src/modules/database/providers/dangerous-commands.provider.ts
@@ -1,0 +1,69 @@
+import { Injectable, Logger } from '@nestjs/common';
+import { OnEvent } from '@nestjs/event-emitter';
+import { AppRedisInstanceEvents } from 'src/constants';
+import { RedisClient } from 'src/modules/redis/client';
+
+@Injectable()
+export class DangerousCommandsProvider {
+  private logger = new Logger('DangerousCommandsProvider');
+
+  private cache: Map<string, string[]> = new Map();
+
+  /**
+   * Get cached dangerous commands list for the connection, or fetch and cache it.
+   */
+  async getDangerousCommands(client: RedisClient): Promise<string[]> {
+    const cached = this.cache.get(client.clientMetadata.databaseId);
+    if (cached) {
+      return cached;
+    }
+
+    const commands = await this.fetchDangerousCommands(client);
+    this.cache.set(client.clientMetadata.databaseId, commands);
+    return commands;
+  }
+
+  /**
+   * Invalidate the cached entry for a given database id.
+   * Should be called when the connection for this database is closed.
+   */
+  invalidate(databaseId: string): void {
+    this.cache.delete(databaseId);
+  }
+
+  @OnEvent(AppRedisInstanceEvents.Deleted)
+  handleInstanceDeletedEvent(databaseId: string): void {
+    this.invalidate(databaseId);
+  }
+
+  /**
+   * Run `ACL CAT dangerous` against the connected Redis instance.
+   * Returns an upper-cased, deduplicated list of command names.
+   * Returns an empty list if ACL is not supported (e.g. Redis < 6.0).
+   */
+  private async fetchDangerousCommands(client: RedisClient): Promise<string[]> {
+    try {
+      const reply = (await client.call(['acl', 'cat', 'dangerous'], {
+        replyEncoding: 'utf8',
+      })) as string[];
+
+      if (!Array.isArray(reply)) {
+        return [];
+      }
+
+      return Array.from(
+        new Set(
+          reply
+            .filter((cmd): cmd is string => typeof cmd === 'string')
+            .map((cmd) => cmd.toUpperCase()),
+        ),
+      );
+    } catch (e) {
+      this.logger.debug(
+        'ACL CAT dangerous is not available on this Redis instance',
+        e?.message,
+      );
+      return [];
+    }
+  }
+}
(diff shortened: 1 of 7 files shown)


redisinsight/api/src/modules/database/providers/dangerous-commands.provider.ts
```diff
@@ -0,0 +1,69 @@
...
+    return commands;
+  }
+
+  /**
+   * Invalidate the cached entry for a given database id.
+   * Should be called when the connection for this database is closed.
+   */
+  invalidate(databaseId: string): void {
+    this.cache.delete(databaseId);
+  }
```
ArtemHoruzhenko, line 47: I'm not sure how you are handling this but keep in mind that single database can have multiple connections and any of it might be closed for some reason. Ideally it is better to understand that there is no more opened connections for particular database. Anyway, since it is cache - it is good enough

redisinsight/api/src/modules/database/providers/dangerous-commands.provider.ts
```diff
@@ -0,0 +1,69 @@
...
+            .filter((cmd): cmd is string => typeof cmd === 'string')
+            .map((cmd) => cmd.toUpperCase()),
+        ),
+      );
+    } catch (e) {
+      this.logger.debug(
+        'ACL CAT dangerous is not available on this Redis instance',
+        e?.message,
+      );
+      return [];
```
ArtemHoruzhenko: I think we might want to check if there is NOPERM error and cache result. There is no need to run command which will always fail

ArtemHoruzhenko (commented):
overall looks good

ArtemHoruzhenko (dismissed):

redisinsight/api/src/modules/database/providers/dangerous-commands.provider.ts
KrumTy: Good point — to be clear: invalidation here is wired to AppRedisInstanceEvents.Deleted (whole instance removed), not to individual connection close, since the dangerous-commands list is a server property and stays valid across reconnects. I've also tightened the invalidate() JSDoc to remove the misleading 'on connection close' wording (baa5d13b3).

redisinsight/api/src/modules/database/providers/dangerous-commands.provider.ts
KrumTy: The current code does cache the empty result (the catch returns [] and getDangerousCommands sets it unconditionally), so NOPERM/ACL-unsupported calls are only issued once per database. The hit check happened to rely on [] being truthy though — I've switched it to Map.has() in baa5d13b3 to make that explicit.

ArtemHoruzhenko: right. then it is even worst. because we might cache [] due to network issue in some particular call

ArtemHoruzhenko (dismissed):

redisinsight/api/src/modules/database/providers/dangerous-commands.provider.ts
KrumTy: Good catch — split the error handling: NOPERM and 'unknown command' (Redis < 6) are now classified as permanent and cached as [], while transient errors (ECONNRESET, ETIMEDOUT, etc.) return [] without caching so a retry next call is still possible. Pushed in a8ff01976 with covering tests.

ArtemHoruzhenko (approved):
lgtm

PR #5914, opened 2026-05-14
hide all vector set for version < 8.0 and when ff is off 

dantovska:
# What

Hide Vector Set key type from the Browser when it's not usable:
- On Redis < 8.0 (version doesn't support Vector Sets).
- When the `devVectorSet` feature flag is off.
Applies to both the key type filter dropdown and the New Key dropdown.

# Testing

1. On Redis < 8.0: "Vector Set" is absent from the filter and New Key dropdowns.
2. On Redis >= 8.0 with `devVectorSet` off: still hidden.
3. On Redis >= 8.0 with `devVectorSet` on: visible in both dropdowns.

---

> [!NOTE]
> **Low Risk**
> UI-only gating logic to hide/show a key type based on Redis version and a feature flag; low risk but could affect available dropdown options if the version/flag detection is wrong.
> 
> **Overview**
> Hides the **Vector Set** key type in the Browser unless *both* Redis is `>= 8.0` and the `devVectorSet` feature flag is enabled, applying to both the **New Key** type dropdown and the **key type filter** dropdown.
> 
> Refactors both dropdowns to derive option availability from option metadata (`minVersion` + optional `isEnabledSelector` evaluated against Redux state), adds typed option definitions, and extends tests to cover the new version/flag gating behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4e3f3091ccc9d55281bfb7fbb5e19a85949000b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/pages/browser/components/add-key/AddKey.tsx
+++ b/redisinsight/ui/src/pages/browser/components/add-key/AddKey.tsx
@@ -10,13 +10,21 @@ import {
   resetAddKey,
   keysSelector,
 } from 'uiSrc/slices/browser/keys'
-import { connectedInstanceSelector } from 'uiSrc/slices/instances/instances'
+import {
+  connectedInstanceOverviewSelector,
+  connectedInstanceSelector,
+} from 'uiSrc/slices/instances/instances'
 import {
   sendEventTelemetry,
   TelemetryEvent,
   getBasedOnViewTypeEvent,
 } from 'uiSrc/telemetry'
-import { isContainJSONModule, Maybe, stringToBuffer } from 'uiSrc/utils'
+import {
+  isContainJSONModule,
+  isVersionHigherOrEquals,
+  Maybe,
+  stringToBuffer,
+} from 'uiSrc/utils'
 import { RedisResponseBuffer } from 'uiSrc/slices/interfaces'
 
 import { Col, FlexItem, Row } from 'uiSrc/components/base/layout/flex'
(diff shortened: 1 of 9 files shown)


redisinsight/ui/src/pages/browser/components/add-key/AddKey.tsx
```diff
@@ -64,7 +78,21 @@ const AddKey = (props: Props) => {
...
 
-  const options = ADD_KEY_TYPE_OPTIONS.map((item) => {
+  const options = ADD_KEY_TYPE_OPTIONS.filter(
+    ({ minVersion, typeFeatureFlag }) => {
+      if (minVersion && !isVersionHigherOrEquals(version, minVersion)) {
+        return false
+      }
+      if (
+        typeFeatureFlag &&
+        !isDevelopment &&
```
ArtemHoruzhenko: Not sure that having feature enabled for dev mode makes sense tbh

ArtemHoruzhenko (dismissed):
looks good

redisinsight/ui/src/pages/browser/components/add-key/AddKey.tsx
dantovska: i changed it so this check is only in one place. anyway, we decided that while in development, it would be easier if we enable the flag by default so we don't switch it constantly.

pawelangelow (approved):

PR #5929, opened 2026-05-15
Docs: update vector set bruno docs with better examples

dantovska:
# What

Update vector set bruno docs with better examples, using the basic straightforward compass as a vector set key name and the sides of the world as vector set elements 

# Testing

1. Open bruno and choose the workspace
2. Vector Set collection should load
3. Start from the top to the bottom - first create, get, delete ...


valkirilov (approved):
Lovely 👌

PR #5930, opened 2026-05-15
RI-8196 Enrich BE analytics events with environment

KrumTy:
# What

Enriches existing backend analytics events with the connection's `environment` classification so we can answer the PRD's analytics questions ("did users run dangerous commands on prod?", "did Profiler start on prod?", "did bulk-delete confirm on prod?") without adding parallel FE `PROD_CONFIRMATION_SHOWN` events. Part of epic [RI-6594](https://redislabs.atlassian.net/browse/RI-6594) — Prod vs non-prod modes. Ticket: [RI-8196](https://redislabs.atlassian.net/browse/RI-8196).

Builds on [RI-8198](https://redislabs.atlassian.net/browse/RI-8198) (the `environment` enum: `unspecified | production | development`, schema convention from `database.analytics.ts`) and [RI-8179](https://redislabs.atlassian.net/browse/RI-8179) (`DangerousCommandsProvider`, an ACL-CAT-backed dangerous-command cache).

### Events enriched

| Module | Events | New fields |
| --- | --- | --- |
| CLI | `CLI_COMMAND_EXECUTED`, `CLI_COMMAND_ERROR_RECEIVED`, `CLI_CLUSTER_NODE_COMMAND_EXECUTED` | `environment`, `isDangerous` |
| Workbench | `WORKBENCH_COMMAND_EXECUTED`, `WORKBENCH_COMMAND_ERROR_RECEIVED`, `SEARCH_*` variants, `WORKBENCH_INDEX_INFO_SUBMITTED` | `environment`, `isDangerous` (index info: `environment` only) |
| Bulk actions | `BULK_ACTIONS_STARTED`, `BULK_ACTIONS_STOPPED`, `BULK_ACTIONS_SUCCEED`, `BULK_ACTIONS_FAILED`, `IMPORT_SAMPLES_UPLOADED` | `environment`, `confirmedThrough` (Bulk doesn't carry `isDangerous` — `action: overview.type` already conveys Delete/Unlink vs Upload) |
| Profiler | `PROFILER_LOG_DOWNLOADED`, `PROFILER_LOG_DELETED`, **new** `PROFILER_STARTED` | `environment` |

### Implementation notes

- **Analytics services are pure event constructors.** They no longer inject `DatabaseRepository` or `DangerousCommandsProvider` — all telemetry context comes from the caller. CLI and Workbench analytics methods accept `environment: Environment` and `isDangerous: 'true' | 'false'` as positional primitives. Bulk-actions and profiler analytics accept the `Database` object directly (only `.id` and `.environment` are read).
- **Consumer-side resolution.** `CliBusinessService.sendCommand` and `WorkbenchCommandsExecutor.sendCommand` resolve `database.environment` and `isDangerous(client, command)` once per request (as `const`s inside the existing try block) and pass them through to every analytics call. On the error path the analytics emit uses literal defaults (`Environment.Unspecified`, `'false'`) — matches master's error-path semantics, which never carried these fields.
- **Bulk-actions and profiler** consumers (`BulkActionsService.create`, `BulkImportService.import`, `ProfilerService.addListenerForInstance`) fetch the `Database` once and thread it through. `BulkAction` and `LogFile` models hold the `Database` for their lifetime and pass it to deferred analytics emits.
- **`DangerousCommandsProvider`** is now exported from `DatabaseModule`. Its `isDangerous(client, command)` method centralizes the null-guard + uppercase + `Set.has` logic.
- **New optional `confirmedThrough: 'standard' | 'type-to-confirm'`** on `CreateBulkActionDto`, threaded through `BulkAction` → `IBulkActionOverview` → analytics. Defaults to `null` for non-confirmation paths.
- **`additionalData`** on the CLI/Workbench command-event analytics methods is typed as `CliCommandEventData` / `WorkbenchCommandEventData` — no inline casts; consumers comply at the type system level.
- **New `PROFILER_STARTED`** event fires from `ProfilerService.addListenerForInstance` after the listener attaches.
- **Pre-existing `sessionMetadata`-plumbing bugs fixed:**
  - `BulkAction.sendOverview` is invoked via a debounce wrapper with no args; `sessionMetadata` is now stored on the instance from the constructor.
  - `LogFile` analytics events were invoked with 2 args missing `sessionMetadata`; threaded through `LogFileProvider.getOrCreate` from `ProfilerClientProvider`.
- **Internal `resolveEnvironment` util removed.** Earlier iteration wrapped `databaseRepository.get(...)` inside the analytics layer; the consumer-lift refactor made it obsolete.

### Out of scope (deferred)

Browser write events (key rename / delete / TTL change / value mutations, hash/list/set/zset/stream writes). The ticket's #5 scope item assumes existing BE analytics to enrich, but the codebase has **none** for those services — they're FE-tracked today. Adding them is a net-new module-wide telemetry effort and exceeds the "enrich existing emits" framing. [RI-8184](https://redislabs.atlassian.net/browse/RI-8184) keeps its FE confirmation event or gets a follow-up BE ticket.

### Schema note

The original ticket targeted an `isProduction: boolean` field. [RI-8198](https://redislabs.atlassian.net/browse/RI-8198) (merged 2026-05-20) replaced that with a tri-state `environment` enum and dropped the `skipConfirmationsForNonProduction` user setting. This PR is rebased on top of RI-8198 and emits the enum directly (no stringification), matching the convention RI-8198 established in `database.analytics.ts`.

# Testing

- **Unit tests**: cli + workbench + bulk-actions + profiler suites pass (~146 tests across the touched analytics surface). New assertions cover:
  - `environment === Environment.Production` when the database is production-flagged
  - `isDangerous='true'` when the caller marks the command as dangerous (CLI/Workbench)
  - `confirmedThrough` round-trips through the bulk-action DTO
  - `BulkAction.sendOverview` invoked through the debounce wrapper resolves the stored sessionMetadata
  - `ProfilerLog{Downloaded,Deleted}` receive the correct 3-arg call shape
- **Lint**: clean on all touched files.
- **Type check**: no new `src/` errors; touched analytics services emit/accept strictly typed `additionalData`.
- **Manual smoke** (still to do before un-drafting): with a production and unspecified database connected, run `GET foo` and `FLUSHDB` from CLI/Workbench and confirm Segment payloads carry the correct `environment` and `isDangerous`; trigger a bulk-delete with the FE confirmation dialog and confirm `confirmedThrough='type-to-confirm'`; open Profiler and confirm `PROFILER_STARTED` fires once.

[RI-6594]: https://redislabs.atlassian.net/browse/RI-6594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RI-8196]: https://redislabs.atlassian.net/browse/RI-8196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RI-8198]: https://redislabs.atlassian.net/browse/RI-8198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RI-8179]: https://redislabs.atlassian.net/browse/RI-8179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RI-8184]: https://redislabs.atlassian.net/browse/RI-8184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Touches multiple request paths (CLI, Workbench, bulk actions, profiler) to thread database context into analytics emits; mistakes could break telemetry or introduce runtime errors when context is missing, but changes are limited to event construction and supporting plumbing.
> 
> **Overview**
> Adds **environment-aware backend telemetry** across CLI, Workbench/Search, bulk actions, and Profiler so emitted events include the connection `environment` (and, for command execution, an `isDangerous` flag).
> 
> Bulk actions now accept an optional `confirmedThrough` value on create, persist it on the `BulkAction`/overview, and include it plus `environment` in all bulk-action related events; services/providers were updated to fetch the `Database` once and pass it through.
> 
> CLI and Workbench execution paths now resolve `database.environment` and compute `isDangerous` via a new `DangerousCommandsProvider.isDangerous()` helper (provider exported from `DatabaseModule`), and include these fields in success/error telemetry payloads.
> 
> Profiler telemetry now includes `environment` for log download/delete and introduces a new `PROFILER_STARTED` event emitted when a listener is attached; profiler log files carry `sessionMetadata`/`database` so deferred analytics calls have the right context.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c178d40d29650a9db2f4fe68a3ef26f857840f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redisinsight/api/src/modules/bulk-actions/bulk-actions.analytics.ts
```diff
@@ -32,16 +58,23 @@ export class BulkActionsAnalytics extends TelemetryBaseService {
             BULK_ACTIONS_BREAKPOINTS,
           ),
         },
+        isProduction: await resolveIsProduction(
+          this.databaseRepository,
+          sessionMetadata,
+          overview.databaseId,
+        ),
+        dangerous: this.isDangerousAction(overview.type),
```
valkirilov: nit: why don't we prefix this with `isDangerous` as well, since it's a boolean check.

redisinsight/api/src/modules/bulk-actions/bulk-actions.analytics.ts
```diff
@@ -32,16 +58,23 @@ export class BulkActionsAnalytics extends TelemetryBaseService {
...
+        confirmedThrough: this.resolveConfirmedThrough(overview),
```
valkirilov: nit: Here, I think we should also prefix it with `is*`, but to be honest, I'm not sure whether the current `confirmedThrough` name is good enough to provide context for what this is used for. 

It sounds a bit vague to me, but I'm not sure what else to use that will be more descriptive.

redisinsight/api/src/modules/cli/services/cli-analytics/cli-analytics.service.ts
```diff
@@ -118,12 +132,23 @@ export class CliAnalyticsService extends CommandTelemetryBaseService {
...
-        ...(await this.getCommandAdditionalInfo(additionalData['command'])),
-        ...additionalData,
+        ...(await this.getCommandAdditionalInfo(rest['command'])),
+        ...rest,
+        isProduction: await resolveIsProduction(
+          this.databaseRepository,
+          sessionMetadata,
+          databaseId,
+        ),
+        dangerous: dangerous ? 'true' : 'false',
```
valkirilov: nit: Why do we need string booleans here?

redisinsight/api/src/modules/bulk-actions/bulk-actions.analytics.ts
KrumTy: it's actually an enum - standard | type-to-confirm

redisinsight/api/src/modules/bulk-actions/bulk-actions.analytics.ts
KrumTy: yup, was thinking the same, waiting for PO to confirm

redisinsight/api/src/modules/cli/services/cli-analytics/cli-analytics.service.ts
KrumTy: looks like an established convention in the repo, analytics booleans are stringified

redisinsight/api/src/modules/bulk-actions/models/bulk-action.ts
```diff
@@ -316,18 +323,22 @@ export class BulkAction implements IBulkAction {
 
   /**
    * Send overview to a client
-   * @param sessionMetadata
+   * @param sessionMetadata defaults to the value stored at construction time
    */
-  sendOverview(sessionMetadata: SessionMetadata) {
+  sendOverview(
+    sessionMetadata: SessionMetadata | undefined = this.sessionMetadata,
```
ArtemHoruzhenko: sessionMetadata should never be undefined.

redisinsight/api/src/modules/bulk-actions/bulk-actions.analytics.ts
```diff
@@ -1,16 +1,26 @@
...
 import { SessionMetadata } from 'src/common/models';
+import { DatabaseRepository } from 'src/modules/database/repositories/database.repository';
+import { resolveEnvironment } from 'src/modules/database/utils/resolve-environment';
 
 @Injectable()
 export class BulkActionsAnalytics extends TelemetryBaseService {
-  sendActionStarted(
+  constructor(
+    protected eventEmitter: EventEmitter2,
+    private readonly databaseRepository: DatabaseRepository,
```
ArtemHoruzhenko: I don't think this is good design choice. if telemetry event depends on database we need to pass database here from top layer

redisinsight/api/src/modules/bulk-actions/bulk-actions.analytics.ts
```diff
@@ -32,16 +42,22 @@ export class BulkActionsAnalytics extends TelemetryBaseService {
             BULK_ACTIONS_BREAKPOINTS,
           ),
         },
+        environment: await resolveEnvironment(
+          this.databaseRepository,
+          sessionMetadata,
+          overview.databaseId,
+        ),
```
ArtemHoruzhenko: instead of calculating it here

redisinsight/api/src/modules/database/utils/resolve-environment.ts
```diff
@@ -0,0 +1,20 @@
...
+  sessionMetadata: SessionMetadata,
+  databaseId: string,
+): Promise<Environment> {
+  try {
+    const database = await databaseRepository.get(sessionMetadata, databaseId);
+    return database?.environment ?? Environment.Unspecified;
+  } catch (e) {
+    return Environment.Unspecified;
+  }
+}
```
ArtemHoruzhenko: I think this function should receive `Database` and return `Environment`

redisinsight/api/src/modules/cli/services/cli-analytics/cli-analytics.service.ts
```diff
@@ -98,13 +110,26 @@ export class CliAnalyticsService extends CommandTelemetryBaseService {
...
         ...additionalData,
+        environment: await resolveEnvironment(
+          this.databaseRepository,
+          sessionMetadata,
+          databaseId,
+        ),
+        isDangerous: (await this.dangerousCommandsProvider.isDangerous(
+          client,
+          command,
+        ))
```
ArtemHoruzhenko: I understand why we are doing this but can't say I like it. There is non-zero chance we will execute 'ACL' command each time we are sending such telemetry events. We need to think carefully wether we need this flag at all or not.

ArtemHoruzhenko (changes requested):

redisinsight/api/src/modules/database/utils/resolve-environment.ts
KrumTy: that makes the function obsolete (1 liner) and moves the fetching to consumer

redisinsight/api/src/modules/cli/services/cli-analytics/cli-analytics.service.ts
KrumTy: > There is non-zero chance we will execute 'ACL' command each time we are sending such telemetry events.

only if the user has no permissions; we do cache the response otherwise ([source](https://github.com/redis/RedisInsight/blob/be/feature/RI-8196/enrich-analytics-with-is-production/redisinsight/api/src/modules/database/providers/dangerous-commands.provider.ts#L31))

> We need to think carefully wether we need this flag at all or not.

for example we want to answer "did users run dangerous commands on prod?". Without isDangerous analytics can't differentiate dangerous vs safe

redisinsight/api/src/modules/bulk-actions/models/bulk-action.ts
KrumTy: adjusted here https://github.com/redis/RedisInsight/pull/5930/changes/99e747e067481864ec9c5ccc38eb66e9345d4c51

redisinsight/api/src/modules/bulk-actions/models/bulk-action.ts
```diff
@@ -316,9 +323,9 @@ export class BulkAction implements IBulkAction {
 
   /**
    * Send overview to a client
-   * @param sessionMetadata
+   * @param sessionMetadata defaults to the value stored at construction time
    */
-  sendOverview(sessionMetadata: SessionMetadata) {
+  sendOverview(sessionMetadata: SessionMetadata = this.sessionMetadata) {
```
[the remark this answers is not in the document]
KrumTy: Outdated — addressed when I reverted the async-analytics design. `sendActionSucceed` / `sendActionStopped` / `sendActionFailed` were briefly `async` in the exploration commit `6193adf28`, but the follow-up moved the `Database` lookup out of the analytics layer entirely and back to the caller. All bulk analytics methods are synchronous again as of commit `30c377907`, so there are no floating promises from `sendOverview`.

redisinsight/api/src/modules/bulk-actions/bulk-actions.analytics.ts
```diff
@@ -32,16 +49,22 @@ export class BulkActionsAnalytics extends TelemetryBaseService {
...
+        confirmedThrough: this.resolveConfirmedThrough(overview),
```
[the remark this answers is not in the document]
KrumTy: Intentional — see commit `4b01a98a6` ("refactor(analytics): move isDangerous derivation into the analytics services"). `isDangerous` was dropped from bulk-action events because `action` (= `overview.type`) already carries the same information: `delete` / `unlink` map 1:1 to dangerous, `upload` to not-dangerous. Downstream dashboards can derive the boolean from `action` without adding a redundant field.

For CLI and Workbench events the situation is different — `command` ranges over the full Redis command set so the dangerous classification is not derivable from a single field. Those events do carry `isDangerous` explicitly.

I'll update the PR description to remove the bulk-action `isDangerous` mention to avoid further confusion.

redisinsight/api/src/modules/cli/services/cli-analytics/cli-analytics.service.ts
```diff
@@ -115,15 +140,28 @@ export class CliAnalyticsService extends CommandTelemetryBaseService {
...
+          ? 'true'
+          : 'false',
```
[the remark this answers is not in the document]
KrumTy: Addressed in commit `5b81b6efc` by lifting `isDangerous` resolution out of the analytics services entirely. The consumers (`cli-business.service.sendCommand` and `workbench-commands.executor.sendCommand`) now resolve `isDangerous` against the same parsed `command` they pass in `additionalData`, then thread the result through to all analytics calls as a positional arg. `DangerousCommandsProvider` is no longer injected into either analytics service.

The divergence window you flagged is closed: the emit's `command` field and the `isDangerous` flag now share a single source (the consumer's parsed command). They can no longer disagree.

redisinsight/api/src/modules/cli/services/cli-analytics/cli-analytics.service.ts
```diff
@@ -98,13 +110,26 @@ export class CliAnalyticsService extends CommandTelemetryBaseService {
...
       this.sendEvent(sessionMetadata, TelemetryEvents.CliCommandExecuted, {
         databaseId,
-        ...(await this.getCommandAdditionalInfo(additionalData['command'])),
+        ...(await this.getCommandAdditionalInfo(command)),
         ...additionalData,
+        environment: await resolveEnvironment(
+          this.databaseRepository,
+          sessionMetadata,
+          databaseId,
+        ),
```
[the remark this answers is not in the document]
KrumTy: Already addressed by commit `ed0716c69` (lift Database fetch out of analytics) and finalized in `e629ff5de`. The CLI analytics service no longer performs any DB lookup — `cli-business.service.sendCommand` resolves the Database once per request and threads it through, matching the bulk-actions / profiler pattern. `resolveEnvironment` is deleted.

redisinsight/api/src/modules/cli/services/cli-business/cli-business.service.ts
```diff
@@ -188,11 +191,25 @@ export class CliBusinessService {
...
+      // doing its own repository lookup. Failure is non-fatal — analytics
+      // calls check for undefined and skip the emit.
+      try {
+        database = await this.databaseService.get(
+          clientMetadata.sessionMetadata,
+          clientMetadata.databaseId,
+        );
+      } catch (e) {
+        // proceed without database — environment will be unknown
+      }
```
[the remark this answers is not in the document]
KrumTy: Good catch — fixed in `e629ff5de`. The `if (database)` guards are gone. `database` is now pre-seeded with a minimal stub (`{ id, environment: Unspecified }` cast as `Database`) so analytics always has something to emit. The real Database overwrites the stub on lookup success; on failure (or if execution never reaches the lookup) the stub stays in place and analytics emits with `environment: 'unspecified'`. The stub only carries the two fields the analytics services actually read.

redisinsight/api/src/modules/cli/services/cli-business/cli-business.service.ts
```diff
@@ -252,6 +271,8 @@ export class CliBusinessService {
           clientMetadata.sessionMetadata,
           clientMetadata.databaseId,
           error,
+          Environment.Unspecified,
+          'false',
```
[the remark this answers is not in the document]
KrumTy: compromise for error case

redisinsight/api/src/modules/bulk-actions/models/bulk-action.ts
```diff
@@ -310,24 +318,41 @@ export class BulkAction implements IBulkAction {
...
     this.debounce();
   }
 
   /**
    * Send overview to a client
-   * @param sessionMetadata
+   * @param sessionMetadata defaults to the value stored at construction time
    */
-  sendOverview(sessionMetadata: SessionMetadata) {
+  sendOverview(sessionMetadata: SessionMetadata = this.sessionMetadata) {
```
ArtemHoruzhenko: If we already set sessionMetadata in constructor, let's use this.sessionMetadata instead of passing it here?

redisinsight/api/src/modules/bulk-actions/bulk-import.service.ts
```diff
@@ -90,10 +91,13 @@ export class BulkImportService {
   /**
    * @param clientMetadata
    * @param fileStream
+   * @param prefetchedDatabase Optional pre-fetched Database to avoid a redundant
+   *   repository lookup when the caller already has it (e.g. `importDefaultData`).
    */
   public async import(
     clientMetadata: ClientMetadata,
     fileStream: Readable,
+    prefetchedDatabase?: Database,
```
ArtemHoruzhenko: nit: do we need this? can't say that we will get database in this service or have requirement to pass this database from caller?

redisinsight/api/src/modules/bulk-actions/bulk-import.service.ts
```diff
@@ -111,15 +115,28 @@ export class BulkImportService {
...
+          clientMetadata.databaseId,
+        ));
+
+      this.analytics.sendActionStarted(
+        clientMetadata.sessionMetadata,
+        result,
+        database,
+      );
+
       client = await this.databaseClientFactory.createClient(clientMetadata);
```
ArtemHoruzhenko, line 135: actually previous command is not nit. we must rely on clientMetadata to fetch database in this service. it is the only way

redisinsight/api/src/modules/cli/services/cli-analytics/cli-analytics.service.ts
```diff
@@ -98,13 +105,17 @@ export class CliAnalyticsService extends CommandTelemetryBaseService {
   public async sendCommandExecutedEvent(
     sessionMetadata: SessionMetadata,
     databaseId: string,
-    additionalData: object = {},
+    environment: Environment,
+    isDangerous: 'true' | 'false',
```
ArtemHoruzhenko: nit: cant `environment and` `isDangerous` be part of `additionalData`?

redisinsight/api/src/modules/cli/services/cli-business/cli-business.service.ts
```diff
@@ -255,6 +274,8 @@ export class CliBusinessService {
           {
             command,
             outputFormat,
+            environment: Environment.Unspecified,
+            isDangerous: 'false',
```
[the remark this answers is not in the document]
KrumTy: this is error state

ArtemHoruzhenko (approved):
lgtm

PR #5962, opened 2026-05-22
RI-8186 Gate dangerous CLI commands with type-to-confirm modal

KrumTy:
# What

Gates dangerous Redis commands (e.g. `FLUSHDB`, `FLUSHALL`, `DEBUG`) in the CLI behind a type-to-confirm modal on production-environment connections. Two pieces of wiring:

1. **On-connect fetch** — `fetchConnectedInstanceAction` now dispatches `fetchConnectedInstanceDangerousCommandsAction` right after `setConnectedInstanceSuccess`, populating the dangerous-commands list in Redux for both CLI and Workbench. Fire-and-forget; failure doesn't block the connection.
2. **CLI gating** — `CliBodyWrapper` checks `useDatabaseEnvironment().isDangerousCommand(cmd)` before dispatching. When true, the command is held in local state and the modal opens; the user types the database name to confirm. Cancel resets the input and dispatches nothing. The hook already short-circuits to `false` outside `Environment.Production` (and when the `devProdMode` flag is off), so no extra environment guard is needed at the call site.

No FE telemetry — per RI-8196 the BE-side `CLI_COMMAND_EXECUTED` event will carry `environment` + `dangerous` flags (separate PR #5930).

Screenshot 2026-05-22 at 10 14 21
(copy is subject to change) 

# Testing

- [ ] `yarn test` passes for `CliBodyWrapper.spec.tsx` and `slices/tests/instances/instances.spec.ts`
- [ ] Connect to a database marked as Production (requires `devProdMode` flag on). Open the CLI.
  - [ ] `FLUSHDB` → modal opens with the DB name; cancel → command not dispatched; submit again → type the DB name → confirm → command runs
  - [ ] `PING` → runs immediately, no modal
- [ ] Connect to a Development-marked database → `FLUSHDB` runs without a modal
- [ ] With `devProdMode` flag off → `FLUSHDB` runs without a modal regardless of environment

---

References: #RI-8186

Depends on: #RI-8198 (hook + slice action), #RI-8179 (ACL CAT endpoint)
Sibling: #RI-8187 (Workbench gating — same hook, reuses the same on-connect fetch)

---

> [!NOTE]
> **Medium Risk**
> Changes CLI command submission flow and adds an extra on-connect fetch, which could impact command dispatch behavior and connection sequencing if the new gating or async fetch behaves unexpectedly.
> 
> **Overview**
> Adds *dangerous command* protection in the CLI: when `useDatabaseEnvironment().isDangerousCommand()` flags a command, `CliBodyWrapper` holds it and opens a `TypeToConfirmModal` requiring the user to type the database identifier before dispatching; cancel clears input and appends a new `cliTexts.DANGEROUS_COMMAND_CANCELLED` output.
> 
> Updates `fetchConnectedInstanceAction` to also trigger `fetchConnectedInstanceDangerousCommandsAction` after a successful connect (fire-and-forget) so the per-instance dangerous-commands list is populated in Redux, and extends unit tests to cover the new connect behavior and the CLI modal confirm/cancel paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf83016cc1b9a3f3c169f7ab4ac0fc3b0e6f001e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/components/cli/components/cli-body/CliBodyWrapper.spec.tsx
+++ b/redisinsight/ui/src/components/cli/components/cli-body/CliBodyWrapper.spec.tsx
@@ -18,6 +18,8 @@ import { processCliClient } from 'uiSrc/slices/cli/cli-settings'
 import { connectedInstanceSelector } from 'uiSrc/slices/instances/instances'
 import { processUnsupportedCommand } from 'uiSrc/utils/cliOutputActions'
 import { useDatabaseEnvironment } from 'uiSrc/components/hooks/useDatabaseEnvironment'
+import { DBInstanceFactory } from 'uiSrc/mocks/factories/database/DBInstance.factory'
+import { ConnectionType } from 'uiSrc/slices/interfaces'
 
 import CliBodyWrapper from './CliBodyWrapper'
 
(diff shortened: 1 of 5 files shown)


redisinsight/ui/src/components/cli/components/cli-body/CliBodyWrapper.spec.tsx
```diff
@@ -114,6 +132,96 @@ describe('CliBodyWrapper', () => {
...
+        host: 'h',
+        port: 6379,
+        name: 'prod-db',
+      }))
+      ;(getCommandRepeat as jest.Mock).mockReturnValue(['FLUSHDB', 1])
+    })
+
+    it('does not dispatch the command and shows the modal when the command is dangerous', () => {
+      const sendCliCommandActionMock = jest.fn()
+      ;(sendCliCommandAction as jest.Mock).mockImplementation(
```
valkirilov: nit: let's avoid the leading `;` at the start of these lines. The agents make this mistake very often, and we tend to fix it via the following:

```
const mockedConnectedInstanceSelector = connectedInstanceSelector as jest.Mock
mockedConnectedInstanceSelector.mockImplementation(() => ({ ... }))

const mockedGetCommandRepeat = getCommandRepeat as jest.Mock
mockedGetCommandRepeat.mockReturnValue(['FLUSHDB', 1])
```

redisinsight/ui/src/components/cli/components/cli-body/CliBodyWrapper.spec.tsx
```diff
@@ -114,6 +132,96 @@ describe('CliBodyWrapper', () => {
     expect(processUnsupportedCommandMock).toBeCalled()
   })
 
+  describe('dangerous-command gating', () => {
+    const { getCommandRepeat } = jest.requireMock('uiSrc/utils')
+
+    beforeEach(() => {
+      ;(connectedInstanceSelector as jest.Mock).mockImplementation(() => ({
```
valkirilov: nit: Let's use the existing factory instead of hardcoding the instance shape here. 

We already have `DBInstanceFactory` which produces a valid Instance and stays in sync with the interface - please use it and override only the fields the test actually cares about 🙂

valkirilov (approved):

PR #5975, opened 2026-05-28
RI-8200 Skip Type-to-confirm modal for command for session

KrumTy:
# What

In `TypeToConfirmModal` (shown for production-environment databases) the user now sees an optional **"Don't ask again for this command in this session"** checkbox. When ticked, subsequent requests sharing the same `commandId` bypass the modal until the connected database changes or the app is reopened.

**Refactors** `CliBodyWrapper` and `WBViewWrapper` to go through `ProductionWriteConfirmationProvider` (same path the browser key-details modules already used) instead of mounting `TypeToConfirmModal` directly. The provider is hoisted into `InstancePageTemplate` so all three consumers share a single session-scoped skip set per database. RDI template is untouched.

**CommandId namespacing**:
- `browser:*` — key-details ops (`browser:edit-value`, `browser:rename-key`, `browser:change-ttl`, …)
- `cmd:*` — Redis verbs (shared between CLI and Workbench, so opting out of `FLUSHDB` in the CLI also skips it for a subsequent Workbench `FLUSHDB`)

The provider accepts `commandId: string | string[]`; Workbench passes the array of dangerous verbs in a batch, and the modal is skipped only when *every* id is in the skip set.

Screenshot 2026-05-28 at 11 36 20

# Testing

1. Connect to a database flagged as **Production** (env-aware feature flag enabled).
2. **Browser**: edit a string value → modal appears → type DB name, tick the checkbox, confirm → edit another string value → modal no longer appears. Switch to a different (or non-production) database and edit again → modal returns.
3. **CLI**: run `FLUSHDB` → modal appears → tick the checkbox, confirm → run `FLUSHDB` again → no modal. Run `DEBUG SLEEP 1` → modal still appears (different verb).
4. **Workbench**: submit `FLUSHALL\nFLUSHDB` → modal lists both → tick + confirm → submit `FLUSHALL\nFLUSHDB` again → no modal. Submit `FLUSHALL\nDEBUG SLEEP 1` → modal returns (DEBUG not in skip set).
5. Cross-feature consistency: opt out of `FLUSHDB` in CLI → run `FLUSHDB` from Workbench → no modal.
6. Switch databases or restart the app → skip set resets; first dangerous command prompts again.

- [ ] yarn lint
- [ ] yarn test
- [ ] manual flows above

---

Closes RI-8200

---

> [!NOTE]
> **Medium Risk**
> Users can bypass repeat confirmations for dangerous Redis commands and production writes after one typed confirm; scope is limited to session and per-database, but it still weakens the guardrail for destructive operations.
> 
> **Overview**
> Adds an optional **"Don't ask again for this command in this session"** checkbox to production type-to-confirm flows. When checked, `ProductionWriteConfirmationProvider` remembers `commandId`(s) for the rest of the session and skips the modal on repeat requests until the connected database changes.
> 
> **CLI** and **Workbench** no longer mount `TypeToConfirmModal` locally; they use the shared provider with `cmd:*` ids (e.g. `cmd:FLUSHDB`) so opting out in one surface applies to the other. **Browser** key-details actions get stable `browser:*` ids; the provider moves from `KeyDetails` up to **`InstancePageTemplate`** so browser, CLI, and workbench share one skip set per database.
> 
> Workbench passes an array of dangerous verb ids for multi-command batches; skip only applies when every id in that batch was previously opted out.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33198e609870751b481d15f44a0abc75bdc879c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/components/production-write-confirmation/ProductionWriteConfirmationProvider.spec.tsx
+++ b/redisinsight/ui/src/components/production-write-confirmation/ProductionWriteConfirmationProvider.spec.tsx
@@ -3,6 +3,7 @@ import { Environment } from 'apiClient'
 import { fireEvent, render, screen } from 'uiSrc/utils/test-utils'
 import { connectedInstanceSelector } from 'uiSrc/slices/instances/instances'
 import * as useDatabaseEnvironmentModule from 'uiSrc/components/hooks/useDatabaseEnvironment'
+import { DBInstanceFactory } from 'uiSrc/mocks/factories/database/DBInstance.factory'
 
 import {
   ProductionWriteConfirmationProvider,
@@ -10,14 +11,11 @@ import {
   useProductionWriteConfirmation,
 } from '.'
 
+const defaultInstance = DBInstanceFactory.build({ name: 'prod-cache' })
+
 jest.mock('uiSrc/slices/instances/instances', () => ({
   ...jest.requireActual('uiSrc/slices/instances/instances'),
-  connectedInstanceSelector: jest.fn().mockReturnValue({
-    id: 'instance-1',
-    name: 'prod-cache',
-    host: 'localhost',
-    port: 6379,
-  }),
+  connectedInstanceSelector: jest.fn(),
 }))
 
 const mockedConnectedInstanceSelector = connectedInstanceSelector as jest.Mock
@@ -71,12 +69,7 @@ const renderWithProvider = (ui: React.ReactElement) =>
 describe('ProductionWriteConfirmationProvider', () => {
   beforeEach(() => {
     jest.clearAllMocks()
-    mockedConnectedInstanceSelector.mockReturnValue({
-      id: 'instance-1',
-      name: 'prod-cache',
-      host: 'localhost',
-      port: 6379,
-    })
+    mockedConnectedInstanceSelector.mockReturnValue(defaultInstance)
     mockEnvironment(Environment.Unspecified)
   })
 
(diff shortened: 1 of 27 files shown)


redisinsight/ui/src/components/production-write-confirmation/ProductionWriteConfirmationProvider.spec.tsx
```diff
@@ -13,6 +13,7 @@ import {
 jest.mock('uiSrc/slices/instances/instances', () => ({
   ...jest.requireActual('uiSrc/slices/instances/instances'),
   connectedInstanceSelector: jest.fn().mockReturnValue({
+    id: 'instance-1',
```
valkirilov: nit: If we were using a factory instead, we'll have to touch a single place and rely on the correct interface for this object, instead of amending each line separately.

redisinsight/ui/src/pages/browser/modules/key-details/components/hash-details/add-hash-fields/AddHashFields.tsx
```diff
@@ -168,6 +168,7 @@ const AddHashFields = (props: Props) => {
         </>
       ),
       confirmButtonText: 'Add fields',
+      commandId: 'browser:add-hash-fields',
```
valkirilov: nit: I think it will be handy to put these in an enum, or constants, so we have a list of them all, in a single place.

valkirilov (dismissed):

valkirilov (approved):

PR #5984, opened 2026-05-29
RI-8190 E2E coverage for prod vs non-prod modes

KrumTy:
# What

Adds Playwright E2E coverage for prod vs non-prod modes. All scenarios run with `dev-prodMode` enabled via `test.use({ featureFlags: ... })`.

Specs are co-located with the feature area each scenario exercises:

- **`databases/list/environment-badge.spec.ts`** — Production / Development / Unspecified DBs render the right badge (or no badge) in the list and the instance header.
- **`browser/key-details/environment-gating.spec.ts`** — Browser writes on a Production DB open the modal (no typing needed); writes on a Development DB bypass it entirely (per-connection gating).
- **`browser/bulk-actions/environment-gating.spec.ts`** — Bulk delete on a Production DB keeps the typed-input variant; mistyped name keeps Confirm disabled.
- **`cli/dangerous-commands.spec.ts`** — `FLUSHDB` on a Production DB opens the modal; cancel preserves keys, confirm runs (verified against the real Redis backend).
- **`workbench/dangerous-commands.spec.ts`** — Workbench batch containing a dangerous command opens the modal mid-batch; same cancel/confirm contract.
- **`workbench/profiler.spec.ts`** — Starting the Profiler on a Production DB requires a confirmation popover.
- **`insights/tutorials.spec.ts`** — Tutorial **Run** buttons are disabled on a Production DB.

Modal copy, per-component wiring, and per-action confirmation strings are covered by unit tests and were deliberately not duplicated here.

## Infrastructure

- `TypeToConfirmModal` page object + fixture exposing both `confirm()` (input variant) and `confirmWithoutInput()` (Browser variant).
- `Environment` enum in `e2eSrc/types`; optional `environment` flows through `ApiHelper.createDatabase`.
- `selectEnvironment()` helper on `AddDatabaseDialog`; `fillForm` honours the new field.
- `DatabaseList.getRow` switched from a cell-level locator to the inner name `<p>` so the environment badge in the same cell doesn't break the lookup.
- TEST_PLAN entries distributed across sections 1.2, 2.4, 2.11, 3.1, 3.3, 3.4 and 4.1 (no dedicated section — coverage lives next to its peers).

# Testing

- [ ] CI: full Playwright run (`@e2e-tests` label).
- [ ] Manual: `yarn dev:api` + `yarn dev:ui`, then `cd tests/e2e-playwright && npx playwright test --project=chromium tests/main/browser/bulk-actions tests/main/browser/key-details tests/main/cli tests/main/workbench tests/main/insights/tutorials.spec.ts tests/main/databases/list/environment-badge.spec.ts` against the docker-compose RTE.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to E2E tests, fixtures, and TEST_PLAN; no application runtime logic is modified in this diff.
> 
> **Overview**
> Adds Playwright E2E coverage for **prod vs non-prod** behavior with `dev-prodMode` enabled, plus shared test infrastructure so scenarios can set **Production / Development / Unspecified** on connections and drive the UI consistently.
> 
> New specs exercise **PROD/DEV badges** in the database list and instance header, **type-to-confirm** on production for browser writes (rename vs dev bypass on edit), bulk delete (typed DB name), CLI and Workbench dangerous commands (`FLUSHDB`, cancel/confirm against Redis), **profiler start** confirmation, and **disabled tutorial Run** on production.
> 
> Supporting changes: `TypeToConfirmModal` and `ProfilerPanel` page objects wired through fixtures; `Environment` on API/UI helpers; `DatabaseList.getRow` targets the name cell’s inner `<p>` so badges don’t break row lookup; **TEST_PLAN** marks the new cases as implemented.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f713593e3651578bd51cf37bc98d74a44d7d23f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/e2e-playwright/pages/components/TypeToConfirmModal.ts
```diff
@@ -0,0 +1,98 @@
...
+  readonly input: Locator;
+  readonly confirmButton: Locator;
+  readonly cancelButton: Locator;
+  readonly closeButton: Locator;
+  readonly skipForSessionCheckbox: Locator;
+  readonly tip: Locator;
+
+  constructor(page: Page) {
+    this.page = page;
+    this.title = page.getByTestId('type-to-confirm-modal-title');
```
KrumTy, line 25: it's a generic, reusable modal, that's used with various props based on case; hence it's easier to manage it via test ids in e2e

tests/e2e-playwright/TEST_PLAN.md
```diff
@@ -355,6 +360,7 @@ The test plan is organized by feature area. Tests are grouped for parallel execu
 | 🔲 | main | Confirm deletion failures surfaced in summary log |
 | 🔲 | main | Confirm performance when deleting thousands of keys |
 | 🔲 | main | Confirm performance when bulk uploading large datasets (>10K keys) |
+| ✅ | main | Production DB > should require typing the database name to bulk-delete |
```
valkirilov, line 363: note: Make sure to run this test in an isolated database instance, so it doesn’t conflict with other tests that depend on the same data, which will be deleted.

tests/e2e-playwright/tests/main/workbench/profiler.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    test.afterAll(async ({ apiHelper }) => {
+      await apiHelper.deleteDatabase(database.id).catch(() => {});
+    });
+
+    test('should require confirmation to start the profiler', async ({ browserPage, page }) => {
+      await browserPage.goto(database.id);
+      await browserPage.openProfiler();
+
+      await page.getByTestId('start-monitor').click();
```
valkirilov: nit: I think you can hide this inside a page object for the Profiler component

example: `tests/e2e-playwright/pages/browser/components/AddKeyDialog.ts`

valkirilov: nit: is there a way to use a better visual selector, instead of the IDs

tests/e2e-playwright/tests/main/workbench/profiler.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+      await expect(page.getByTestId('profiler-start-confirm')).toBeVisible();
+      await expect(page.getByTestId('profiler-start-cancel')).toBeVisible();
+
```
valkirilov, line 32: suggestion: Can we verify visually that the profiler is running and outputting stuff, to make these e2e test meaningful? Otherwise, the same check can be performed via a simple jest test, for much less cost.

tests/e2e-playwright/TEST_PLAN.md
```diff
@@ -447,6 +453,7 @@ The test plan is organized by feature area. Tests are grouped for parallel execu
 | 🔲 | main | History limited to 30 commands (oldest replaced by newest) |
 | 🔲 | main | Quick-access to command history with Up Arrow |
 | 🔲 | main | Use Non-Redis Editor with Shift+Space |
+| ✅ | main | Production DB > should require type-to-confirm modal for dangerous workbench batches |
```
valkirilov, line 456: note: same as https://github.com/redis/RedisInsight/pull/5984/changes#r3324636891, since the test contains `FLUSHALL` in the list of dangerous commands we execute.

tests/e2e-playwright/tests/main/insights/tutorials.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    test.beforeAll(async ({ apiHelper }) => {
+      database = await apiHelper.createDatabase(StandaloneConfigFactory.build({ environment: Environment.Production }));
+    });
+
+    test.afterAll(async ({ apiHelper }) => {
+      await apiHelper.deleteDatabase(database.id).catch(() => {});
+    });
+
+    test('should disable the tutorial Run button', async ({ browserPage, insightsPanel }) => {
```
valkirilov, line 25: nit: update the test title

> should disable the tutorial Run button when database is flagged as Production

tests/e2e-playwright/pages/components/TypeToConfirmModal.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.description = page.getByTestId('type-to-confirm-modal-description');
+    this.input = page.getByTestId('type-to-confirm-modal-input');
+    this.confirmButton = page.getByTestId('type-to-confirm-modal-confirm-btn');
```
valkirilov, line 28: note: **Title** and **Description** are dynamic, I get that. But for the static labels, I still think we can use real visual selectors, instead of their IDs.

tests/e2e-playwright/tests/main/insights/tutorials.spec.ts
KrumTy: it's contained in the description above
> test.describe('Production DB', () => {

valkirilov: My bad 👍 

Do we need this describe block then, since it's the only test case in the suite, but is 3 level deep already?

tests/e2e-playwright/TEST_PLAN.md
KrumTy: https://github.com/redis/RedisInsight/blob/8cb8d1a1047b5e935ce7ac7b82c2ac29ce965518/tests/e2e-playwright/tests/main/browser/bulk-actions/environment-gating.spec.ts#L19-L21

it creates keys with a keyPrefix, then bulk deletes only them

lets see if this ever collides with another test first?

If we have a test that deliberately wants to read/mutate all data in the database and does not want to rely on predixes, then maybe we can spin an instance for this test alone

tests/e2e-playwright/tests/main/insights/tutorials.spec.ts
KrumTy: unlikely to add more here but I wanted to keep the test itself starting with "should" but also have a prefix in the logs in front that it's for prod environment 

alternative is to push it to the end - "should disable the tutorial Run button for Production environment", which works too if this bothers you

tests/e2e-playwright/pages/components/TypeToConfirmModal.ts
KrumTy: it's all dynamic actually; some may use the default values but still..

I could drop the ids and rely on specific text but I thought this can be left for unit tests and we are testing functionality only here; eventually we can do visual/snapshots and check texts as well

either way, no hard preference for me; if you want I can switch to text

tests/e2e-playwright/tests/main/insights/tutorials.spec.ts
KrumTy: I think I will leave it in describe as it also handles before/after hooks (creating a prod database)

tests/e2e-playwright/pages/components/TypeToConfirmModal.ts
KrumTy: if we go with text level selector we will need a different page object for every modal variant; I'd rather keep it like that; But I've added some expect calls for the dynamic description - verifying it contains dangerous command itself

tests/e2e-playwright/TEST_PLAN.md
valkirilov: I think I saw the `FLUSHALL` command, and it can cause flakiness with other tests (because they run in parallel). At least, it was happening with some of the test suites previously, but if it's not the case, then okay 👌

tests/e2e-playwright/pages/components/TypeToConfirmModal.ts
valkirilov: As I mentioned, I understand that we have dynamic content rendered, and I'm good with that. My comment was targeting the static labels (like the Cancel/Confirm buttons in the modal, etc), because the agent defaulted to test IDs in all the cases.

And it's not about my preferences, it's just what we agreed to follow as a guideline, as a good community standard:
- https://playwright.dev/docs/locators
- https://testing-library.com/docs/queries/about/#priority

>We recommend prioritizing role locators to locate elements, as it is the closest way to how users and assistive technology perceive the page.

tests/e2e-playwright/tests/main/insights/tutorials.spec.ts
valkirilov: Ok, np 👍

tests/e2e-playwright/pages/components/TypeToConfirmModal.ts
KrumTy: yeah not talking about personal preferences here, just picking best tool for the job and this modal seems like an exception to this general rule

you already mentioned that you are fine with id selectors for dynamic content, which title and description are

then I said the rest of the stuff are also dynamic. Even the confirm button is sometimes different - [source](https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/components/cli/components/cli-body/CliBodyWrapper.tsx#L212)

I see two solutions
1. current, treating it as a special case and relying on test ids because it's generic and used with different texts (I still kept assertion for dynamic description that's supposed to show dangerous command for example)
2. make 1 TTC page modal object per variant (I think 14)

a mix of these two is not ideal imo but feel free to suggest something else

tests/e2e-playwright/TEST_PLAN.md
KrumTy: fixed in https://github.com/redis/RedisInsight/pull/5984/changes/8f713593e3651578bd51cf37bc98d74a44d7d23f

valkirilov (approved):

PR #5989, opened 2026-06-01
Split e2e tests into parallel/ and serial/ folders

KrumTy:
# What

Restructure the Playwright test layout: replace the single `tests/main/` directory + `@serial` tag with two sibling folders, and give each browser platform a paired parallel/serial project.

**Folder split**
- `tests/parallel/` — safe to run with multiple workers
- `tests/serial/` — must run sequentially (shared DB state, dangerous commands, vector-index ops)

**Playwright projects** — keyed off `testDir`, no more `grep`/`grepInvert`:
- `chromium-parallel` + `chromium-serial`
- `electron-parallel` + `electron-serial`

**Execution order**

For each platform, the serial project depends on the parallel project (`dependencies: ['<platform>-parallel']`), so the order is always:

1. `<platform>-parallel` runs first (with up to N workers)
2. `<platform>-serial` runs after, one worker at a time

**Why sequential, not concurrent on a shared RTE**

Serial tests perform destructive operations on the shared RTE Redis (`FLUSHDB`, broad `deleteAllIndexes`, dangerous commands). Letting them race a parallel test means the parallel test can observe a wiped database between its own steps. Electron projects can't run concurrently anyway because the desktop app binds its embedded API on a fixed port (5530).

### Trade-offs considered

| Option | Pros | Cons |
|---|---|---|
| **Sequence serial after parallel (chosen)** | Simple mental model: folder location maps to position in the run. No hidden per-test contracts. Same model for chromium and electron. | Adds ~serial-suite-duration to wall-clock per platform. |
| Run both concurrently on a shared RTE, isolate by Redis instance | Faster wall-clock. | Hidden contract — every serial test must use the right factory or it silently races parallel tests. Requires a dedicated container in the compose. Easy to break later. *(Tried and reverted.)* |
| Run both concurrently on separate CI jobs/machines | Faster wall-clock and full isolation. | Doubles RTE bring-up + install cost; for chromium-serial today the savings ≈ duplicated setup. |

If the serial suite grows large enough that the sequential ordering becomes a CI bottleneck, the next step is to give serial tests a dedicated Redis instance (or split into a separate CI job) — not to revert the ordering.

**Bug fix side effect**

The Docker CI workflow previously ran only `--project=chromium`, which silently skipped every `@serial`-tagged test. It now runs both `chromium-parallel` and `chromium-serial`.

# Testing

- [ ] Chromium workflow runs `chromium-parallel`, then `chromium-serial`
- [ ] Docker workflow runs `chromium-parallel`, then `chromium-serial` (previously skipped serial)
- [ ] Electron workflow runs `electron-parallel`, then `electron-serial`
- [ ] Local: `yarn --cwd tests/e2e-playwright test:chromium` runs both
- [ ] Local: `yarn --cwd tests/e2e-playwright test:chromium:serial` runs only the serial set

---

> [!NOTE]
> **Low Risk**
> Changes are limited to E2E layout, Playwright config, and CI commands; no application runtime code is modified.
> 
> **Overview**
> Replaces the single **`tests/main/`** tree plus **`@serial`** grep with **`tests/parallel/`** and **`tests/serial/`**, and wires Playwright to **four projects** (`chromium-parallel` / `chromium-serial`, `electron-parallel` / `electron-serial`) via **`testDir`** instead of tags.
> 
> **`chromium-serial`** and **`electron-serial`** now **depend on** their parallel project so serial suites (shared Redis, `FLUSHDB`, vector index cleanup, dangerous commands) run **after** parallel work finishes, not alongside it. Specs under **`tests/serial/`** drop **`{ tag: '@serial' }`**; Electron-only checks use **`electron-parallel`** instead of **`electron`**.
> 
> CI and docs follow the new names: Chromium, Docker, and Electron workflows run both parallel and serial projects; **`package.json`** adds split scripts (e.g. **`test:chromium:serial`** with **`--no-deps`**). **Docker CI** previously ran only **`--project=chromium`**, which skipped serial tests—now it runs **`chromium-parallel`** and **`chromium-serial`**. **`TEST_PLAN.md`** and AI e2e docs point spec paths at **`tests/serial/`** where applicable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e376cf357717abfa0a9f59780d71acd947debd6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/e2e-playwright/package.json
```diff
@@ -6,12 +6,16 @@
   "license": "MIT",
   "scripts": {
     "test": "playwright test",
-    "test:chromium": "playwright test --project=chromium",
-    "test:chromium:headed": "playwright test --project=chromium --headed",
+    "test:chromium": "playwright test --project=chromium --project=chromium-serial",
```
valkirilov: nit: I know that `project=chromium` being the default one is the one that runs the tests in parallel, but I would rather mark it explicitly, so it's clear if you don't have that context in mind
```suggestion
    "test:chromium": "playwright test --project=chromium-parallel --project=chromium-serial",
```

valkirilov (dismissed):

tests/e2e-playwright/package.json
KrumTy: adjusted here https://github.com/redis/RedisInsight/pull/5989/changes/8929f15d318671d696cb97e7fec3891e59ac2133

valkirilov (dismissed):

valkirilov (dismissed):

valkirilov (approved):

PR #5994, opened 2026-06-02
RI-8208: Bump Redux stack to RTK 2 / react-redux 9 / redux 5 / redux-thunk 3

pawelangelow:
# What

Upgrades the Redux dependency stack and applies the v5 / v9 breaking-
-change fixes:

- `@reduxjs/toolkit` 1.6 → 2.12
- `react-redux` 7 → 9
- `redux-thunk` added as direct dep at 3.1 (no longer bundled in RTK)
- removed `@types/react-redux` (react-redux 9 ships its own types)
- removed `@types/redux-mock-store` (pulled in `redux@4`); replaced
  with a 5-method local shim in
  `redisinsight/ui/src/types/redux-mock-store.d.ts`

Source-level adjustments:

- `slices/store.ts` — explicit `AppDispatch = ThunkDispatch & typeof
  store.dispatch` so the thunk overload survives the inference path,
  and a typed cast on the middleware return that defuses the TS2719
  collision between RTK 2's `redux@5`-rooted `Middleware` and the
  `redux@4` `Middleware` hoisted by `@elastic/eui` → `react-beautiful-
  dnd` → `@types/react-redux@7`. See the comment in `store.ts` for
  the full background and why the cast is preferred over a yarn
  `resolutions` entry.
- `utils/test-utils.tsx` — named `thunk` import, `Options.store`
  widened to `EnhancedStore<any>` so specs can pass focused stores
  without casting.
- 10 spec fixes: stable `getState` references (react-redux 9's
  `useSyncExternalStore` rejects unstable snapshots), `subscribe`
  must return an unsubscriber, `{ type: undefined }` → `{ type:
  '@@INIT' }` (redux 5 rejects undefined action types).
- TS error baseline refreshed (`.tscheck.rec.json`, 1825 → 1823).

# Testing

- `yarn lint`, `yarn type-check`, `yarn test` all pass.
- Manual smoke of the higher-risk flows for the runtime semantics
  that shifted with RTK 2's `autoBatchEnhancer` and react-redux 9's
  `useSyncExternalStore`:
  - app startup + database list (no flicker, no duplicate fetches)
  - connect to instance + open Browser
  - Bulk Delete (with mid-run cancel)
  - CLI after a backgrounded tab (~1 min)
  - Pub/Sub

---

> [!NOTE]
> **Medium Risk**
> Touches global state dependencies and subscription semantics (react-redux 9 / RTK 2) across the whole UI; risk is mitigated by focused test fixes but warrants smoke on Browser, Pub/Sub, CLI, and OAuth flows.
> 
> **Overview**
> This PR **bumps the UI Redux stack** to `@reduxjs/toolkit` 2.12, `react-redux` 9, `redux` 5, and direct `redux-thunk` 3.1, and drops `@types/react-redux` and `@types/redux-mock-store` in favor of built-in types and a small local `redux-mock-store` shim typed against `redux@5`.
> 
> **Runtime / typing follow-ups:** `slices/store.ts` exports a typed `dispatch` as `AppDispatch`; `cliOutputActions` and test helpers use that instead of destructuring from `store`. Browser virtual-tree `Node` now uses a `NodeProps` type (optional `index` for react-vtree) and guards row striping with `(index ?? 0)`.
> 
> **Tests and tooling:** `test-utils` switches to named `thunk` import, widens `store` to `EnhancedStore | MockStoreEnhanced`, and extends default test state (`azure`, `vectorSet`). Specs adapt to Redux 5 (`UnknownAction` payloads, `@@INIT` instead of `undefined` type), react-redux 9 (stable `getState` snapshots, `subscribe` returns unsubscriber), and middleware casts where RTK 2 + thunk 3 collide with legacy `redux@4` types. Some vector-search specs use `mockStore` + `initialStateDefault` instead of mini real stores. `.tscheck.rec.json` drops `test-utils.tsx` baseline entries.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9c76049c97530e1370620d38fc5e193a7fcc6d02. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redisinsight/ui/.tscheck.rec.json
```diff
@@ -1163,9 +1163,6 @@
   "src/utils/streamUtils.ts": {
     "TS2345": 2
   },
-  "src/utils/test-utils.tsx": {
-    "TS2322": 2
-  },
```
pawelangelow, line 1168: side effect - these are fixed 😄

valkirilov (dismissed):

redisinsight/ui/src/types/redux-mock-store.d.ts
```diff
@@ -0,0 +1,38 @@
...
+// that imports test-utils and surfaces as TS2719 "two different types with
+// this name exist, but they are unrelated" errors against RTK 2's
+// `Middlewares` / `EnhancedStore` types (which are derived from `redux@5`).
+//
+// `redux-mock-store` itself ships no types, is in maintenance mode, and is
+// expected to be retired in favor of real `<Provider>` + real store testing.
+// Until that migration happens this shim is the smallest possible surface
+// we own that keeps the existing test infrastructure compiling against
+// `redux@5` types.
+declare module 'redux-mock-store' {
```
KrumTy, line 13: something off here; in the 1st PR you create a wrapper over react-redux methods, while here you override types (two patterns)

would package resolution help in this case? If we need to lock to a certain redux-mock-store version

redisinsight/ui/src/utils/test-store.ts
```diff
@@ -22,14 +22,14 @@ const getState: ReduxStore['getState'] = () => {
   return storeRef.getState()
 }
 
-export const dispatch: ReduxStore['dispatch'] = (action: any) => {
+export const dispatch = ((action: any) => {
```
KrumTy: maybe you can drop the `any` and it will be typed?

redisinsight/ui/src/utils/test-utils.tsx
```diff
@@ -79,7 +82,13 @@ import { setStoreRef } from './test-store'
...
 interface Options {
   initialState?: RootState
-  store?: ReduxStore
+  // Accept either the real RTK store, a `redux-mock-store` instance, or any
+  // narrowly-typed `EnhancedStore` produced by `configureStore` in a spec.
+  // The narrow `EnhancedStore<any>` member is the proper-typed escape hatch
+  // for specs that build a focused store with only the slices they read
+  // from (see vector-search specs) — without it, every such spec needs an
+  // `as any` cast at the call site.
+  store?: ReduxStore | MockStoreEnhanced<RootState> | EnhancedStore<any>
```
KrumTy: this is quite odd, why accept anything other than ReduxStore?

redisinsight/ui/src/utils/cliOutputActions.ts
```diff
@@ -12,7 +12,7 @@ import {
 import { CommandExecutionStatus } from 'uiSrc/slices/interfaces/cli'
 import { concatToOutput } from 'uiSrc/slices/cli/cli-output'
 
-import { store } from 'uiSrc/slices/store'
+import { dispatch, store } from 'uiSrc/slices/store'
```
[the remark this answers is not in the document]
pawelangelow: Under Jest, cliOutputActions.ts's import { dispatch } from 'uiSrc/slices/store' resolves to test-store.ts's dispatch wrapper, which delegates to storeRef.dispatch(action). storeRef is set to mockedStore by test-utils.tsx's module body (setStoreRef(mockedStore)), so actions land in mockedStore, not the real singleton. The real slices/store.ts is never loaded during specs. Verified by running the test — it passes and mockedStore.getActions() does record the dispatched actions.

redisinsight/ui/src/types/redux-mock-store.d.ts
pawelangelow: On resolutions - yes, a "resolutions": { "redux": "^5" } entry in package.json would fix this cleanly, but I don't want to add more things into "resolutions" as it makes potential migration from `yarn` to `npm` harder.

redisinsight/ui/src/utils/test-store.ts
pawelangelow: Tried dropping the any - turns out AppDispatch is an intersection of ThunkDispatch overloads, and TypeScript can't infer a single parameter type across that overload set. Without `: any` it falls back to TS7006 (Parameter 'action' implicitly has an 'any' type). Kept the `as AppDispatch` outer cast and added a comment explaining the constraint, so the next reader doesn't try the same simplification.

KrumTy: this is not good, that means AppDispatch is likely messed up

const dispatch = useAppDispatch()
dispatch(123) // this should show a tsc error - only known actions can be dispatched

package.json
```diff
@@ -308,7 +307,7 @@
     "react-hotkeys-hook": "^3.3.1",
     "react-jsx-parser": "^2.4.1",
     "react-monaco-editor": "^0.59.0",
-    "react-redux": "^7.2.2",
+    "react-redux": "^9.2.0",
```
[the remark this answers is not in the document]
pawelangelow: Done!

valkirilov (dismissed):

KrumTy (approved):

PR #6053, opened 2026-06-11
RI-8216 Establish BigInt-as-string index contract

VaskoAtanasovRedis:
# What

Foundation for the Redis array type (epic RI-8174): array indexes are unsigned 64-bit integers and exceed `Number.MAX_SAFE_INTEGER`, so they must travel as numeric strings end-to-end.

- **Mirrored utilities** (this repo has no shared UI↔API code mechanism, so the contract is two modules with identical semantics and identical test tables): `api/src/common/utils/array-index.helper.ts` ↔ `ui/src/utils/arrayIndex.ts` — `parseArrayIndex` (canonical decimal string or `null`), `isValidArrayIndex`, `ARRAY_INDEX_MAX`.
- **`@IsArrayIndex()` DTO decorator** (mirrors the zset-score validator pattern): accepts the canonical form only — validators don't transform, so anything accepted reaches Redis verbatim. Fails with a stable message consumers can match on: `<field> must be an integer string between 0 and 18446744073709551615`. Supports `ValidationOptions` (`{ each: true }` for `string[]` fields).
- **Hardening**: a 20-digit length guard before the `BigInt()` call caps parse cost for request-sized payloads; `BigInt('…')` call form throughout (API tsconfig targets es2019, where BigInt literals are TS2737).
- **Convention documented** in `.ai/skills/backend/SKILL.md` and `.ai/skills/frontend/SKILL.md` so future array verticals (and agents) follow it.

No runtime consumers yet by design — the array feature verticals (view/search/add/delete) build on this.

# Testing

- `yarn --cwd redisinsight/api test --testPathPattern "array-index"` — 40 tests (helper + decorator, incl. the verbatim stable-message and `{ each: true }` cases)
- `node 'node_modules/.bin/jest' 'redisinsight/ui/src/utils/tests/arrayIndex.spec.ts' -c 'jest.config.cjs'` — 25 tests (mirror table, row-for-row identical to the API helper spec)
- `yarn lint` clean; `yarn type-check` unchanged vs main

---

Closes #RI-8216

---

> [!NOTE]
> **Low Risk**
> New shared validation utilities and tests only; no existing routes or behavior are changed and nothing is consumed at runtime yet.
> 
> **Overview**
> Introduces a **shared contract** for Redis array indexes as **canonical numeric strings** (unsigned 64-bit, `0` … `18446744073709551615`), because indexes exceed `Number.MAX_SAFE_INTEGER`.
> 
> **API** adds `parseArrayIndex`, `isValidArrayIndex`, and `ARRAY_INDEX_MAX` in `array-index.helper.ts`, plus `@IsArrayIndex()` / `ArrayIndexValidator` for DTOs (same shape as existing zset-score validation). Validation accepts **only canonical strings**—no trimming or leading-zero normalization on the wire—so values that pass reach Redis unchanged. A **20-digit length guard** limits `BigInt()` parse cost on oversized inputs.
> 
> **UI** mirrors the same helpers in `arrayIndex.ts` with matching test tables. Both modules are exported from their package `utils` barrels.
> 
> No array endpoints or UI flows are wired in this PR; it is foundation-only for upcoming array work.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51d0d3d21ddd55aa31f348d67ee4c62c7bfff772. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/utils/arrayIndex.ts
+++ b/redisinsight/ui/src/utils/arrayIndex.ts
@@ -1,0 +1,44 @@
+/**
+ * Redis array indexes are unsigned 64-bit integers (0 … 2^64−1) and exceed
+ * Number.MAX_SAFE_INTEGER, so they stay numeric strings end-to-end —
+ * never parseInt/Number, no JS-side arithmetic on indexes; Redux stores
+ * them as strings.
+ *
+ * Mirrored in redisinsight/api/src/common/utils/array-index.helper.ts —
+ * keep semantics and tests in sync.
+ */
+// 2^64 - 1; BigInt() call form (not a literal) for parity with the API
+// mirror, whose tsconfig targets es2019 (where BigInt literals are TS2737).
+export const ARRAY_INDEX_MAX = BigInt('18446744073709551615')
+
+const ARRAY_INDEX_REGEX = /^\d+$/
+
+// Max u64 is 20 digits; longer all-digit inputs can't be valid and a length
+// guard keeps BigInt() from parsing arbitrarily large request payloads.
+const ARRAY_INDEX_MAX_LENGTH = 20
+
+/**
+ * Returns the canonical decimal string for a valid index ("007" → "7"),
+ * or null for anything else (empty or whitespace-only, negative,
+ * fractional, exponent, hex, > 2^64−1, non-string input).
+ */
+export const parseArrayIndex = (input: unknown): string | null => {
+  if (typeof input !== 'string') {
+    return null
+  }
+
+  const value = input.trim()
+  if (!ARRAY_INDEX_REGEX.test(value)) {
+    return null
+  }
+
+  if (value.length > ARRAY_INDEX_MAX_LENGTH) {
+    return null
+  }
+
+  const index = BigInt(value)
+  return index > ARRAY_INDEX_MAX ? null : index.toString()
+}
+
+export const isValidArrayIndex = (input: unknown): boolean =>
+  parseArrayIndex(input) !== null
--- a/redisinsight/api/src/common/decorators/array-index/array-index.decorator.spec.ts
+++ b/redisinsight/api/src/common/decorators/array-index/array-index.decorator.spec.ts
@@ -1,0 +1,65 @@
+import { validate } from 'class-validator';
+import { IsArrayIndex } from 'src/common/decorators';
+
+class SingleIndexDto {
+  @IsArrayIndex()
+  index: string;
+}
+
+class MultiIndexDto {
+  @IsArrayIndex({ each: true })
+  indexes: string[];
+}
+
+const validateSingle = async (index: any) => {
+  const dto = new SingleIndexDto();
+  dto.index = index;
+  return validate(dto);
+};
+
+describe('IsArrayIndex', () => {
+  it.each(['0', '42', '18446744073709551615'])(
+    'should pass for valid index %p',
+    async (input) => {
+      expect(await validateSingle(input)).toHaveLength(0);
+    },
+  );
+
+  it.each([
+    '-1',
+    '1.5',
+    'abc',
+    '',
+    '18446744073709551616',
+    42,
+    undefined,
+    '007',
+    ' 42 ',
+  ])('should fail for %p', async (input) => {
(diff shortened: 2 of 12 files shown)


redisinsight/ui/src/utils/arrayIndex.ts
```diff
@@ -0,0 +1,44 @@
...
+// Max u64 is 20 digits; longer all-digit inputs can't be valid and a length
+// guard keeps BigInt() from parsing arbitrarily large request payloads.
+const ARRAY_INDEX_MAX_LENGTH = 20
+
+/**
+ * Returns the canonical decimal string for a valid index ("007" → "7"),
+ * or null for anything else (empty or whitespace-only, negative,
+ * fractional, exponent, hex, > 2^64−1, non-string input).
+ */
+export const parseArrayIndex = (input: unknown): string | null => {
```
KrumTy, line 25: Unfortunate that these need to be kept in sync, makes me wonder if having a shared package makes sense for such utils.

What would be the use case of this? An input in the ui that allows inserting data in a given index? Even then, wouldn't it be better to allow typing any numeric value but fail on api call? The user can type index outside the array's range.

pawelangelow (dismissed):

redisinsight/ui/src/utils/arrayIndex.ts
pawelangelow: That’s the use case: the array type supports adding a member at a user-selected index, where the index must be between `0` and `MAX_VALUE`. Reference: https://redis.io/docs/latest/develop/data-types/arrays/

If the user enters a value outside that range, we should surface an error, though the exact UX still needs to be defined.

Regarding the shared package: I’m fully in favor of it. This is a good example of logic currently duplicated between the API and UI. Since both need to support the same behavior, the main risk is that the implementations drift out of sync over time. We’ve already seen similar issues with other features, so I think it makes sense to have a dedicated place for business logic that is agnostic to both the UI and API.

VaskoAtanasovRedis: I guess two parts answer

1. The agent did indeed suggest attempting a shared package implementation, but given there's no existing mechanism it seemed like an overkill for such a small util; I prompted it to document the contract in the repository skills so future agents working on this are aware of this contract and will take it in consideration. 

2. The UI will indeed allow users to insert data at a given index (supported by the ARSET/ARMSET cli command), as the data type supports `sparse` arrays which means if a user tries to enter a value at a specific index (in the u64 range), the index is just created. And it seemed valid to mirror input validation on UI level so that a request that is deemed invalid doesn't have to travel UI <-> BE. It also canonicalises input ('007' → '7'), which the API requires: the validator intentionally accepts only the canonical form, because validators don't transform and whatever is accepted travels to Redis verbatim. Server-side validation remains the authority either way.

.ai/skills/backend/SKILL.md
```diff
@@ -238,3 +238,24 @@ Extract repeated strings to constants in constants file.
...
 - [ ] Guards for authentication/authorization
 - [ ] Cognitive complexity ≤ 15
+
+## Array index (BigInt-as-string) contract
+
+Redis array indexes are unsigned 64-bit integers (0 … 2⁶⁴−1) and exceed
+`Number.MAX_SAFE_INTEGER`. They travel as **numeric strings end-to-end**:
+
+- DTOs declare index fields as `string` and validate with `@IsArrayIndex()`
+  (`src/common/decorators`); use `{ each: true }` for `string[]` index fields.
```
KrumTy: In backend skill we have overal architecture, folder structure, code style, etc
I'm not sure we should add such specific internal implementation details here

KrumTy (dismissed):

.ai/skills/backend/SKILL.md
VaskoAtanasovRedis: good point - and @pawelangelow also nicely noted that this will fill in context with unnecessary information that will rarely be used in everyday flow that don't touch these utils or this contract - will drop this

pawelangelow (approved):

Issue #2765, opened 2023-11-13
[Bug]: SSH authentication with private key does not work if passphrase is not used


kanevbg:
**Actual behavior** (A short description of what you found)
![image](https://github.com/RedisInsight/RedisInsight/assets/11477130/d50ae234-da6e-48fd-9415-13a335da12e3)

**Expected behavior** (A short description of what you expected to find)
I have no passphrase for that key, so it should work without requiring me to enter such.

ViktarStarastsenka:
@kanevbg , thanks for reporting this.
What RedisInsight version do you use?

kanevbg:
![image](https://github.com/RedisInsight/RedisInsight/assets/11477130/13380287-8e2c-456a-bf4c-3723e8059622)

ViktarStarastsenka:
@kanevbg , please try with 2.38 (released today).

PR #6064, opened 2026-06-15
RI-8219: Add Array read endpoints + key-info strategy

pawelangelow:
# What

Backend slice of the Array data type **read** vertical (RI-8219).

- **`ARGETRANGE` cap**: 1,000,000-element span guard so a `start..end` range can't be used to pull arbitrary memory; `ARSCAN` is exempt because it already skips empty slots.
- **`ARSCAN` reply normalisation**: handles both the flat `[idx, val, idx, val, …]` and nested `[[idx, val], …]` shapes some clients surface, and drops malformed pairs so `JSON.stringify` never reaches the client as `{ index }` with no value.
- **`ArrayKeyInfoStrategy`** for the key-info pipeline: returns the standard TTL / size triple plus `length` (`ARLEN`, total addressable slots including gaps) and `count` (`ARCOUNT`, populated slots only). The two diverge for sparse arrays and the View tab surfaces both.

# Testing

- **Bruno presets** for every endpoint + a `Seed Sample Data` request that uploads `fixtures/readings.txt` through the existing bulk-import endpoint, populating the sparse `readings` array (indexes 0, 1, 5) that the happy-path presets target. `Get Element (empty slot)` documents the gap-preserving semantics (200 OK with `value: null`, not 404).

---

> [!NOTE]
> **Medium Risk**
> Large new browser surface against Redis with a validation breaking change on max index; mitigated by caps, tests, and read-only paths with standard ACL/404 handling.
> 
> **Overview**
> Adds **Browser Array read APIs** (`get-range`, `scan`, `get-length`, `get-count`, `get-next-index`, `get-element`, `get-elements`) wired to Redis `AR*` commands, with DTOs that return **u64 metrics as strings** and **gap slots as `null`** on dense reads.
> 
> **Safety and wire behavior:** `ARGETRANGE` rejects spans over **1M elements**; `ARMGET` / `ARSCAN LIMIT` use the same cap. **`ARSCAN`** normalizes flat vs nested Redis replies and drops incomplete pairs. **`RedisStringToBufferTransformer`** now passes **`null`/`undefined`** through so empty slots serialize correctly.
> 
> **Key browser:** New **`ArrayKeyInfoStrategy`** exposes **`length`** (ARLEN) and **`count`** (ARCOUNT) as strings; **`get-info`** OpenAPI is a **`oneOf`** with **`GetArrayKeyInfoResponse`**. IORedis registers the new array builtins for pipelines.
> 
> **Breaking alignment:** Valid array index max is **`2^64 − 2`** (not full u64) in API/UI validators, Bruno create docs, and tests. Bruno presets plus a **`readings`** bulk-import fixture exercise sparse arrays and empty-slot **`200` + `null`** semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3cdb3f24a74bdb51c6ac703a135a18759c6ef21. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redisinsight/api/src/modules/browser/array/array.service.ts
```diff
@@ -1,21 +1,52 @@
...
+  GetArrayScanResponse,
+} from 'src/modules/browser/array/dto';
+
+// Integer/bulk replies for indexes and counts may arrive as Buffer, string,
+// number, or bigint depending on the client mode. Normalize to a decimal
+// string so the unsigned 64-bit contract is preserved on the wire.
+const toIndexString = (value: unknown): string => {
+  if (typeof value === 'string') return value;
+  if (typeof value === 'bigint') return value.toString();
+  if (typeof value === 'number') return String(value);
```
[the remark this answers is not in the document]
pawelangelow: Right, the wire-format contract (u64-as-decimal-string) is broken upstream of `toIndexString` - RESP :integer replies are decoded as JS number by the client parser before we ever see them, so String(value) only preserves the already-rounded value.

IMO, there's no clean per-command fix available today:

- ioredis 5 (we're on ^5.2.2) - connection-level stringNumbers: true exists ([PR #282](https://github.com/redis/ioredis/pull/282)), but there's no per-command opt-in. A BigInt mode was requested in [#813](https://github.com/redis/ioredis/issues/813) (2019).
- node-redis 4 - string_numbers was removed in v4. The maintainer confirmed this in [#2765](https://github.com/redis/node-redis/issues/2765) and pointed to v5 for a proper typed-reply API.

Honoring the contract end-to-end would require:

- Routing array integer-reply paths through a dedicated ioredis connection with `stringNumbers: true` (isolates the change from `LLEN`/`HLEN`/etc. callers that assume number), AND
- Upgrading node-redis to v5 to get an equivalent on that backend.

redisinsight/api/src/modules/browser/array/array.service.ts
```diff
@@ -52,6 +83,60 @@ export class ArrayService {
...
+        ERROR_MESSAGES.ARRAY_RANGE_TOO_LARGE(ARRAY_RANGE_MAX_ELEMENTS),
+      );
+    }
+  }
+
+  // The DTO accepts start/end in either order so the UI doesn't have to
+  // enforce ordering; swap before forwarding because ARGETRANGE / ARSCAN
+  // require start ≤ end.
+  private normalizeRange(start: string, end: string): [string, string] {
+    return BigInt(start) > BigInt(end) ? [end, start] : [start, end];
```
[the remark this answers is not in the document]
pawelangelow: Actually i'm dropping this.

redisinsight/api/src/modules/browser/keys/key-info/strategies/array.key-info.strategy.ts
```diff
@@ -0,0 +1,67 @@
...
+        [, count = null],
+        [, size = null],
+      ] = (await client.sendPipeline([
+        [BrowserToolKeysCommands.Ttl, key],
+        [BrowserToolArrayCommands.ArLen, key],
+        [BrowserToolArrayCommands.ArCount, key],
+        [BrowserToolKeysCommands.MemoryUsage, key, 'samples', '0'],
+      ])) as [any, any][];
+
+      return { name: key, type, ttl, size, length, count };
```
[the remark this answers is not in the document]
pawelangelow: i dropped this

redisinsight/api/src/modules/browser/keys/dto/get.keys-info.response.ts
```diff
@@ -45,4 +47,12 @@ export class GetKeyInfoResponse {
     description: 'The vector dimensions for vector set keys.',
   })
   vectorDim?: number;
+
+  @ApiPropertyOptional({
+    type: Number,
```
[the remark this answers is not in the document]
pawelangelow: https://github.com/redis/RedisInsight/pull/6064/commits/7ccd26c98e47c3c59d22e4fc1ea961cdb9d8f63e

redisinsight/api/src/modules/browser/array/dto/get.array-range.response.ts
```diff
@@ -0,0 +1,17 @@
...
+import { RedisString } from 'src/common/constants';
+
+export class GetArrayRangeResponse extends KeyResponse {
+  @ApiProperty({
+    description:
+      'Values for each index in the requested range, in order. ' +
+      'Empty slots are returned as null.',
+    type: 'array',
+    items: { oneOf: [{ type: 'string' }, { type: 'null' }] },
+    nullable: true,
```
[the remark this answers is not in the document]
pawelangelow: https://github.com/redis/RedisInsight/pull/6064/commits/0409c614bb8ed567b8327364c6fdc80a1b09a5d3

redisinsight/api/src/modules/redis/client/ioredis/ioredis.client.ts
```diff
@@ -46,9 +46,18 @@ export abstract class IoredisClient extends RedisClient {
     client.addBuiltinCommand(BrowserToolVectorSetCommands.VSetAttr);
     client.addBuiltinCommand(BrowserToolVectorSetCommands.VRem);
     client.addBuiltinCommand(BrowserToolVectorSetCommands.VSim);
-    // Array commands
+    // Array commands — registered so they can be issued via the pipeline
+    // path (which maps commands to prototype methods); sendCommand works
```
VaskoAtanasovRedis: nit: maybe over-explained comment

redisinsight/api/src/modules/browser/keys/dto/get.array-key-info.response.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Key-info response for the Array data type.
+ *
+ * `length` (ARLEN) and `count` (ARCOUNT) are returned as decimal strings
+ * because an Array's index space is unsigned 64-bit and can exceed
+ * Number.MAX_SAFE_INTEGER for sparse keys. Consumers must narrow on the
+ * `array` key type before reading these fields and avoid silent numeric
+ * coercion that would round large values.
+ */
+export class GetArrayKeyInfoResponse {
```
VaskoAtanasovRedis: I wonder if we can derive most fields from GetKeyInfoResponse via OmitType/PickType instead of re-declaring the shared fields?

redisinsight/api/src/modules/browser/array/dto/get.array-scan.dto.ts
```diff
@@ -0,0 +1,39 @@
...
+  start: string;
+
+  @ApiProperty({
+    description:
+      'End index of the range (inclusive). Unsigned 64-bit integer as string. ' +
+      'Must be greater than or equal to start.',
+    type: String,
+    example: '99',
+  })
+  @IsArrayIndex()
```
VaskoAtanasovRedis, line 25: I think we might need to touch the helper that the decorator uses

Because

Image

but this fails

Image

and I can't set a value at that index

Image

redisinsight/api/src/modules/browser/array/array.service.ts
```diff
@@ -92,4 +169,236 @@ export class ArrayService {
...
+      ] as const;
+      // Treat an explicit JSON `null` the same as an omitted limit. @IsOptional()
+      // skips downstream validators for null, so the DTO accepts it; forwarding
+      // `LIMIT null` to Redis would otherwise surface as a 500.
+      const hasLimit = typeof limit === 'number';
+      const reply = (await client.sendCommand(
+        hasLimit ? [...baseArgs, 'LIMIT', limit] : [...baseArgs],
+      )) as unknown[];
+
+      // ARSCAN's wire shape varies by client: some clients surface a flat
```
VaskoAtanasovRedis: is this correct? from the docs - https://redis.io/docs/latest/commands/arscan/ 

it seems to me like only the flat array structure is returned? I might be missing something

redisinsight/api/src/modules/browser/array/array.service.ts
```diff
@@ -52,6 +76,59 @@ export class ArrayService {
     }
   }
 
+  // Inputs are validated as canonical decimal strings ≤ 2^64-1, so BigInt()
+  // is safe. The contract is start ≤ end — matches every Redis range
```
VaskoAtanasovRedis: This is documented to not be the case for this command tho? in https://redis.io/docs/latest/commands/argetrange/

Image

redisinsight/api/src/modules/browser/array/utils.ts
```diff
@@ -0,0 +1,21 @@
+// Normalize an integer/bulk reply to a decimal string so u64 indexes survive
+// the wire. Returns null for nil replies (e.g. ARNEXT when the cursor is
+// exhausted) so callers can distinguish absence from a real value.
+export const toIndexString = (value: unknown): string | null => {
+  if (value === null || value === undefined) return null;
+  if (typeof value === 'string') return value;
+  if (typeof value === 'bigint') return value.toString();
+  if (typeof value === 'number') return String(value);
```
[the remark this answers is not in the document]
pawelangelow: Yeah, but underlying libraries don't support big numbers, so for now this is a known issue :(

redisinsight/api/src/modules/browser/array/array.service.ts
pawelangelow: Yup, I tested it locally and applied the changes.

pawelangelow: Thanks!

redisinsight/api/src/modules/browser/array/array.service.ts
pawelangelow: Good catch - I had to verify empirically because the docs and the wire don't agree. Tested against `redis:8.8-alpine` (same image as CI's `oss-st-8` https://github.com/redis/RedisInsight/pull/6078):

**RESP wire (source of truth)** - `redis-cli` with `--no-raw` so it doesn't collapse nested arrays:

```
$ printf 'ARSCAN k 0 100\r\n' | redis-cli --no-raw
1) 1) (integer) 0
   2) "a"
2) 1) (integer) 1
   2) "b"
3) 1) (integer) 10
   2) "x"
```

That's `*N` of `*2` — nested at the protocol level. Same on RESP2 and RESP3 (`redis-cli -3`). The plain `redis-cli ARSCAN ...` output in the docs (each value on its own line) is just `redis-cli`'s default flat-print rendering — it walks any nested-array reply and prints each leaf on its own line, which makes it look flat in the "Examples" section.

**ioredis** (what our service sees) — `c.sendCommand(new Command('ARSCAN', [key, '0', '100']))`:

```
[ [ 0, <Buffer 61> ], [ 1, <Buffer 62> ], [ 10, <Buffer 78> ] ]
```

Also nested. That matched the CI failure exactly — the flat parser was building `index: '0,20.1'` strings from `String(['0','20.1'])` and `value` was `undefined` on every odd pair, which is why 6 of the scan tests collapsed to 1 element.

The Return-info section of the docs (`Array reply: Flat array of index-value pairs`) is wrong/outdated.

VaskoAtanasovRedis (approved):

Issue #5967, opened 2026-05-23
[Feature Request]:multi-language supported


simidalalala:
**Problem description** (Describe the problem you would like to solve with this feature request or suggested idea).

As the greatest redis view tools on the world，We hope that the RedisInsight could support more languages, such as Chinese.

DimoHG:
Sorry @simidalalala, still not on the roadmap :/

Issue #5052, opened 2025-10-09
[Feature Request]:how to switch UI language in Windows version?


lvxingqi:
Is it possible to add a drop-down box in the general settings to switch the UI language? I see that the installation directory contains language files, but I can't switch the UI language. Or is this a feature that will be released in a future version?

ViktarStarastsenka:
Hi @lvxingqi,
We currently don’t have this on the roadmap.
Any updates regarding support for additional languages will be shared in this ticket - https://github.com/redis/RedisInsight/issues/1317

Issue #4744, opened 2025-07-22
Add 国际化-中文


qjyn1314:
界面的操作和按钮的语言, 是否可以支持语言设置, 比如设置为中文

ViktarStarastsenka:
Hi @qjyn1314 ,
Currently, Redis Insight supports only English.
Here is a ticket to support other languages - https://github.com/redis/RedisInsight/issues/1317.

Issue #4672, opened 2025-07-09
[Feature Request]:So far, is there no support for Chinese?


zhankenson:
A developer friend recommended this software and took a look. A few years ago, there were many requests to support Chinese. So far, does it not support Chinese?

ViktarStarastsenka:
hi @zhankenson , currently, Redis Insight supports only English.
Hi, here is a ticket where the updates will be posted - https://github.com/redis/RedisInsight/issues/1317

PR #6096, opened 2026-06-22
RI-8266 Add i18n foundation (i18next)

valkirilov:
# What

Set up the **i18next foundation** for the UI so user-facing strings can move to `t()` incrementally. No mass migration here — just the shared infrastructure everything else builds on. **Zero behavior change today** (app stays English).

https://github.com/user-attachments/assets/29eb2d00-cdae-4419-9fac-4f2e0e72b59e

# Why

Redis Insight has no i18n - every string is hardcoded English across ~1,580 files, which blocks shipping other languages and turns copy changes into a code hunt. Doing the whole migration at once would be an unreviewable, conflict-prone megadiff. This PR lands the enabler so feature migrations can follow **independently, PR by PR**.

# What's included

- **Deps:** `i18next`, `react-i18next`, `i18next-cli` (dev — key extraction).
- **`redisinsight/ui/src/i18n/`:** one app-level instance, single `translation` namespace, `en` + `bg` locales, flat keys, and **type-safe keys** via `en.json` augmentation (no codegen).
- **App root** wrapped in `<I18nextProvider>`; `renderComponent`/`renderHook` use the **real** instance, so existing `getByText('...')` assertions keep working as strings migrate.
- **`yarn i18n:extract`** (config: `i18next.config.mjs`) syncs `t()` usages into the locale files.
- **Worked example:** the Settings page strings are moved behind `t()` with seeded `en`/`bg` translations.

> The Settings example is illustrative — a `?lang=bg` dev override is included (temporary) to preview locales. A real, persisted language switcher is intentionally **not** part of this PR.

# Usage

```tsx
import { useTranslation } from 'uiSrc/i18n'

const { t } = useTranslation()
return <span>{t('settings.title')}</span>
```

1. Add the key to `locales/en.json` (source of truth) and translate it in `bg.json`.
2. `t()` rejects unknown keys at **compile time**.
3. Run `yarn i18n:extract` to pull new `t()` usages into the locale files.
4. Preview a locale in dev: append `?lang=bg`.

> This is **step 1**. Follow-up PRs migrate strings **feature by feature** (browser, workbench, settings, …), each adding its own keys + translations. Keeping the foundation separate keeps those PRs small, reviewable, and conflict-free.

# Testing

- `yarn type-check`, `yarn lint:ui`, `yarn test` — green (no new TS-baseline errors).
- `yarn dev:ui` — app behaves identically in English.
- `yarn i18n:extract` — runs and updates the locale files.
- Append `?lang=bg` on the Settings page to see Bulgarian.

---

Ref: RI-8266

Related community requests: #605, #1317, #1883, #2040, #2137, #2181, #2213, #2417, #2550, #2590, #2695, #2715, #2716, #2727, #2828, #2912, #2943, #2945, #3081, #3141, #3142, #3319, #3411, #3601, #3625, #3628, #3663, #3671, #3812, #3815, #4150, #4182, #4408, #4664, #4672, #4744, #5052, #5967

---

> [!NOTE]
> **Low Risk**
> Infrastructure and a limited Settings string migration; English remains default with fallback, and tests use the same i18n instance so behavior should match today.
> 
> **Overview**
> Adds **i18next** / **react-i18next** to the UI with a shared `uiSrc/i18n` module (`en` + `bg`, flat keys, compile-time typing from `en.json`), **`yarn i18n:extract`** via `i18next.config.mjs`, and wraps the app and test helpers in **`I18nextProvider`** using the real i18n instance.
> 
> **Settings** is the first consumer: page title, section accordion labels, and the advanced-settings warning now use `t()` with matching locale entries. Default language stays **English**; a temporary **`?lang=bg`** query override lets devs preview Bulgarian before a real language switcher exists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7d114743a7c3367f72971b2fa55cd50b9c92e04f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redisinsight/ui/src/i18n/i18n.ts
```diff
@@ -0,0 +1,38 @@
...
+  supportedLngs: [...SUPPORTED_LANGUAGES],
+  ns: [DEFAULT_NAMESPACE],
+  defaultNS: DEFAULT_NAMESPACE,
+  // Flat keys (dots are literal, not nesting) — matches i18next.config.mjs.
+  keySeparator: false,
+  nsSeparator: false,
+  interpolation: {
+    // React already escapes values, so disable i18next's own escaping.
+    escapeValue: false,
+  },
```
pawelangelow, line 36:
```ts
returnEmptyString: false,
```

By default i18next treats `""` as a valid translation and renders it as-is, so any key that exists in `bg.json` with an empty value will render blank instead of falling back to `en`. This matters here because `yarn i18n:extract` seeds newly found keys into `bg.json` with empty strings - without this flag, every freshly migrated string ships blank on BG until a translator fills it in.

With `returnEmptyString: false`, an empty value counts as missing and `fallbackLng: 'en'` kicks in, so users see English until the BG translation lands.

Ref: [i18next — Fallback › Missing values for existing keys](https://www.i18next.com/principles/fallback#missing-values-for-existing-keys)

redisinsight/ui/src/pages/settings/SettingsPage.tsx
```diff
@@ -115,8 +117,7 @@ const SettingsPage = () => {
       )}
       <CallOut className={styles.warning}>
         <Text size="s" className={styles.smallText}>
-          Advanced settings should only be changed if you understand their
-          impact.
+          {t('settings.advancedWarning')}
         </Text>
```
KrumTy, line 121: what are your thoughts on this approach vs having translation functionality within low level text components and showing the key in case translation value is missing?

So this would look like

```
  <Text size="s" className={styles.smallText}>
   settings.advancedWarning
</Text>
```

or

```
  <Text size="s" className={styles.smallText} tKey="settings.advancedWarning" />
```

Then parent won't need to care about translations (calling t) and at some point the children (or another prop) can be typed as key of i18n translation keys and not just string

redisinsight/ui/src/pages/settings/SettingsPage.tsx
```diff
@@ -128,7 +129,7 @@ const SettingsPage = () => {
       <PageBody component="div">
         <PageHeader>
           <Title size="XXL" className={styles.title}>
-            Settings
+            {t('settings.title')}
```
KrumTy, line 132: is this strictly typed? E.g. can I call t with non existing key?

redisinsight/ui/src/i18n/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Single `translation` namespace.
+- `lng` / `fallbackLng`: `en`. Supported: `en`, `bg`.
+- Locales live in [`locales/`](./locales) — `en.json` is the source of truth.
+- Type-safe keys: [`i18next.d.ts`](./i18next.d.ts) augments i18next with `typeof en.json`. No codegen — add a key to `en.json` and `t()` picks it up.
+
+## Using a string
+
+```tsx
+import { useTranslation } from 'uiSrc/i18n'
```
KrumTy, line 13: I think we might also need a non-hook method for translation - helper utilities, Redux thunks, API services, or event handlers might require it. But this is good enough for now though

redisinsight/ui/src/pages/settings/SettingsPage.tsx
valkirilov: You see a warning, but it's not crashing the app

image

redisinsight/ui/src/pages/settings/SettingsPage.tsx
valkirilov: > showing the key in case translation value is missing?

I think that the safe fallback is to go for the default value, a.k.a. the English variant, instead of showing the key name. Yes, we risk it to miss something to remain forgotten and not translated, but otherwise when we miss something we break the app (visually). It's a debatable choice, but it also contradicts what @pawelangelow suggested [here](https://github.com/redis/RedisInsight/pull/6096#discussion_r3452875548).

_PS: We can think of other ways of catching non-translated stuff, like putting all the translation into an emoji, and if we spot text somewhere in the app, it means we missed translating it. But this is just an example, we can discuss it separately._

> what are your thoughts on this approach vs having translation functionality within low level text components

Regarding your real suggestion, to not use the `t` directly, and rely on inner components to use it - I don't have any hard feelings right now.

We can discuss it with the team and see what are the pros/cons of this approach 👍

redisinsight/ui/src/i18n/i18n.ts
valkirilov: Makes sense, I thought this was the default behavior, thanks for flagging it. Fixed in 422a8bbd8

redisinsight/ui/src/i18n/README.md
valkirilov: Yes, for sure 👍 

Most probably we'll introduce such in the follow-up PRs once we need it.

redisinsight/ui/src/i18n/i18n.ts
```diff
@@ -0,0 +1,39 @@
...
+const queryLanguage = new URLSearchParams(window.location.search).get(
+  'lang',
+) as SupportedLanguage
+
+i18n.use(initReactI18next).init({
+  resources: {
+    en: { translation: en },
+    bg: { translation: bg },
+  },
+  lng: queryLanguage ?? DEFAULT_LANGUAGE,
```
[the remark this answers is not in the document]
valkirilov: Yes, that's true, but it's a temporary way to verify whether the language switch works and renders with the correct version. It will be dropped soon, so we can go with this approach for now.

redisinsight/ui/src/pages/settings/SettingsPage.tsx
KrumTy: yes the english fallback is standard mechanism, I was referring to missing from en.json as well; fallback here usually is the key itself

pawelangelow: Just checking if I have all the context for this - if we are missing translation, so we fallback to the key name - it means that we didn't run the collect script, right? If this is true, we can somehow include the collect script into the building process so we don't forget it, or implement a CI step that checks it.

redisinsight/ui/src/pages/settings/SettingsPage.tsx
```diff
@@ -41,6 +42,7 @@ import { DateTimeFormatter } from './components/general-settings'
 import styles from './styles.module.scss'
 
 const SettingsPage = () => {
+  const { t } = useTranslation()
```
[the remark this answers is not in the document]
valkirilov: Good point, but we're not translating the app yet, it's just an example.

redisinsight/ui/src/pages/settings/SettingsPage.tsx
ArtemHoruzhenko: In my opinion 
```        
<Text size="s" className={styles.smallText}>
  {t('settings.advancedWarning')}
</Text>
```
 is the right way

KrumTy: @ArtemHoruzhenko  with this approach we need to import and use t everywhere, delegating translation to parent and passing it down X times to the child

having a Text component that attempts to translate the given text (key) or displays it outright as a fallback would save us this repetition; and we'd have a centralized way of displaying + formatting + translating text (in most cases) rather then split across the app

ArtemHoruzhenko: @KrumTy My concern is that if we add translation logic to `Text`, the same argument applies to many other components too: `Title`, `Button`, `Tooltip`, labels, menu items, table headers, notifications, etc. Then we either need to add `tKey` everywhere, or we end up with an inconsistent component API.

I think keeping translation at the usage/app layer is cleaner: the place that knows what message should be shown calls `t()`, and low-level components only render what they receive. That also keeps components reusable for literal text, dynamic values, tests, and non-i18n content.

So I’d prefer:

```<Text>{t('settings.advancedWarning')}</Text>```

over making `Text` or other base components responsible for resolving translation keys.

ArtemHoruzhenko: One more concern: `Text` might render user/database data like keys, values, names, command output, etc.

If children are treated as translation keys, user data might be translated. surprise :)

KrumTy: @ArtemHoruzhenko It doesn't have to be embedded in Text, it can be a wrapper over it and used anywhere (e.g. TranslatedText)

What I suggested is that instead of using t directly and passing (presumably translated) strings across multiple components, we can have a wrapper component for translations that accepts valid keys only (ts checked) and be used for rendering text; right now we don't have such a common component for rendering text as you mentioned

valkirilov: There is already such a component provided by the i18n library we use ([link to docs](https://react.i18next.com/latest/trans-component)). But whether we use`t()` or `<Trans></Trans>` - I'm not sure how big the difference is 🤔 

And regarding what @ArtemHoruzhenko said, I think he is right, and we don't have to add an abstraction layer in each presentation component we have (`Text`, `Title`, `Tooltip`, `Popup`, etc.), because we'll end up with mixed convention (on some places it will be embeded in the compnent, but on other we'll still have to use the hook). 

So if there is no clear benefit of adding a separate abstraction layer, maybe we don't have to over-engineer it 🤷‍♂️

dantovska (approved):

redisinsight/ui/src/pages/settings/SettingsPage.tsx
KrumTy: Well I've already listed all the benefits, not sure what to add. 

We have a lot of presentational components that just render something; now they would need to use a hook and yield a big diff in the PR ([example](https://github.com/redis/RedisInsight/blob/main/redisinsight/ui/src/pages/database-analysis/components/summary-per-data/SummaryPerData.tsx#L56)).

A single render-time choke point for translation tooling would allow us to handle 1) missing keys 2) telemetry 3) other logic.

Overall I have no strong feelings. I've used it in projects both ways and I find having all text we translate and display go through a single component easier to manage. But refactoring is cheap now.

KrumTy (approved):

pawelangelow (approved):

PR #6099, opened 2026-06-23
Generate the OpenAPI client during build:statics and guard the api postinstall

dantovska:
# What

Make from-source / consumer builds correctly produce and tolerate the generated OpenAPI client (`redisinsight/api-client/`) that the UI imports:

- `build:statics` (and `build:statics:win`) now run `generate:api-client` **before** the UI/plugin build, so any from-source build generates the client it needs — instead of relying solely on the api `postinstall`, which is skipped in layered/Docker builds where `install` runs before the source is copied.
- The api `postinstall` is guarded with `fs.existsSync`, so it no-ops when `scripts/postinstall-generate-client.js` isn't present (e.g. the api installed as a dependency, or at a Docker cache layer) instead of aborting `yarn install` with `MODULE_NOT_FOUND`. `patch-package` still runs.

Motivated by RedisInsight-Cloud, which consumes the api as a local dependency and builds the UI statics from source; without these, its image build fails (missing `api-client/` in the UI build, and `MODULE_NOT_FOUND` on the postinstall).

# Testing

- `yarn --cwd redisinsight/api install` no longer fails when `scripts/` isn't copied yet (guarded postinstall no-ops; with full source it still generates).
- A from-source statics build (`buildStatics=1`) generates `api-client/` before the UI/plugin build, resolving the prior `[vite:load-fallback] Could not load .../api-client` failure.
- Verified end-to-end via redislabsdev/RedisInsight-Cloud#397, whose Build API Images job now passes with the submodule pointed at this branch.

---

> [!NOTE]
> **Low Risk**
> Build and install-script changes only; no runtime API, auth, or data-path behavior is modified.
> 
> **Overview**
> **From-source statics builds** now run `yarn generate:api-client` as part of `build:statics` and `build:statics:win`, after `build:defaults` and before the UI/plugin statics scripts, so `redisinsight/api-client/` exists when the UI resolves `apiClient`—instead of depending only on api `postinstall`, which may not run (or run too early) in layered Docker or consumer-repo installs.
> 
> The **api `postinstall`** still runs `patch-package`, but only invokes `scripts/postinstall-generate-client.js` when that file is present, avoiding `MODULE_NOT_FOUND` when the api package is installed before its scripts are copied or when used as a dependency without the full tree.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e6272674e08c7d3a8a0fe319fe54885ae7383eb7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dantovska:
Closing — RedisInsight-Cloud will handle the from-source api-client generation in its own Dockerfiles instead of changing the submodule, so these upstream changes are no longer needed. The submodule stays on stock main.

PR #6122, opened 2026-06-27
Fix type-check regression from stricter match() inference

pawelangelow:
# What

`yarn type-check` was failing on main after the typeorm 0.3.29 / socket.io
bumps tightened type inference, surfacing latent strict-null errors where
`String.match()` results (`RegExpMatchArray | null`) were destructured
directly:

> TS2488: Type 'RegExpMatchArray | null' must have a '[Symbol.iterator]()'
> method that returns an iterator.

- Guard the match before destructuring in both affected API files —
  `clusterCursor.ts` (cursor parser) and `cluster.ioredis.client.ts`
  (NAT address parsing). `host`/`port` stay typed as `string`.
- Refresh the api/desktop TS baselines: the same bumps regenerated
  `api-client` and lowered recorded error counts that were never updated.

These were the only two `.match()`-destructure sites in the API.

---

> [!NOTE]
> **Low Risk**
> Small null-safety guards in cluster cursor and NAT parsing; baseline JSON only affects recorded type-check debt, not runtime behavior for valid inputs.
> 
> **Overview**
> Restores **`yarn type-check`** after stricter inference from dependency bumps (e.g. typeorm / socket.io) exposed **TS2488** on direct destructuring of `RegExpMatchArray | null`.
> 
> **Runtime/type fixes:** **`parseClusterCursor`** and the ioredis **cluster NAT map** parser now assign `match` / `natMatch`, return or skip when there is no match, then destructure so `host`/`port` stay correctly narrowed.
> 
> **Baselines:** Updates **`redisinsight/api/.tscheck.rec.json`** (drops the fixed `cluster.ioredis.client` entry) and **`redisinsight/desktop/.tscheck.rec.json`** to reflect lower recorded error counts after regenerated **`api-client`** and related typing changes—not new product logic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 62b777c3f6ba3c5729d68c582b18031e9055ef04. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisinsight/api/.tscheck.rec.json
+++ b/redisinsight/api/.tscheck.rec.json
@@ -1206,9 +1206,6 @@
   "src/modules/redis-sentinel/redis-sentinel.analytics.ts": {
     "TS18048": 1
   },
-  "src/modules/redis/client/ioredis/cluster.ioredis.client.ts": {
-    "TS2488": 1
-  },
   "src/modules/redis/client/ioredis/ioredis.client.ts": {
     "TS2345": 2,
     "TS2532": 1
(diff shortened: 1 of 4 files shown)


valkirilov (approved):
note: Maybe we should enable the type-check job to run on Dependabot PRs, because this is why I missed that regression :/

Issue #4664, opened 2025-07-05
[Feature Request]: Multi-Language Support


01Petard:
Are there plans to support multiple languages in the future? I really hope to support multiple languages, as many people have become accustomed to using software in their native language after using another Redis desktop manager.

ViktarStarastsenka:
Hi, here is a duplicate ticket - https://github.com/RedisInsight/RedisInsight/issues/1317
The updates will be posted there.

Issue #4408, opened 2025-03-06
有计划增加中文支持吗


CrowdLonely:
有计划增加中文支持吗

ViktarStarastsenka:
hi, here is a similar ticket - https://github.com/RedisInsight/RedisInsight/issues/1317

Issue #4182, opened 2024-12-09
[Feature Request]:language and account management


IrisHxr:
It is hoped that two functions can be added in the subsequent versions:
Add a language selection interface in the general settings to facilitate the use of developers from various countries and make the tool more powerful.
Add account management. Different accounts can have different permissions. For example, developers can only have read-only permissions or read and write permissions for individual modules, while operation and maintenance managers can have all permissions. The ks-control access control in the kubesphere-system can be taken as a reference. The tool can not only manage visualization but also conduct security control, which is really excellent.

ViktarStarastsenka:
hi @IrisHxr , thank you for your requests.
1. regarding the language selection - here is the same request - https://github.com/RedisInsight/RedisInsight/issues/1317
2. about the account managemet, the current recommendation is to use:

- standard Redis ACL to limit commands that can be executed - https://redis.io/docs/latest/operate/oss_and_stack/management/security/acl/
- use a reverse proxy to limit users who can access the list of databases. An example is here - https://github.com/RedisInsight/RedisInsight-reverse-proxy

IrisHxr:
> 你好@IrisHxr，谢谢您的请求。
> 
> 1. 关于语言选择 - 这是相同的请求 - [[功能请求]：添加多语言支持](https://github.com/RedisInsight/RedisInsight/issues/1317)[ #1317](https://github.com/RedisInsight/RedisInsight/issues/1317)[](https://github.com/RedisInsight/RedisInsight/issues/1317)
> 2. 关于账户管理，目前建议使用：
> 
> * 标准 Redis ACL 用于限制可以执行的命令 - https://redis.io/docs/latest/operate/oss_and_stack/management/security/acl/
> * 使用反向代理来限制可以访问数据库列表的用户。示例如下 - https://github.com/RedisInsight/RedisInsight-reverse-proxy

@ViktarStarastsenka Thank you for your reply. I hope that the support for multiple languages will be developed in your subsequent versions.
As for account management, I hope that the web interface of RedisInsight can have different account control features. The reason is that some of my Redis instances haven't enabled the ACL (Access Control List) yet. Once it is enabled, many clients will need to be modified and the availability of the service may be affected. I'm looking forward to your reply.

IrisHxr:
> hi @IrisHxr , thank you for your requests.
> 
[...]

Issue #4150, opened 2024-11-28
[Feature Request]:I hope there is a Chinese version


danjiaxin555:
My English is not very good, I hope the author can support internationalization and provide a Chinese version, thank you

ViktarStarastsenka:
this is a duplicate of https://github.com/RedisInsight/RedisInsight/issues/1317

PR #6125, opened 2026-06-29
RI-8268 Translate backend error messages (round 1: error path)

valkirilov:
# What

Round 1 of frontend error-message translation, building on the stable `errorCode` contract added to the API in https://github.com/redis/RedisInsight/pull/6120.

API errors are now rendered as **localized** text keyed by `errorCode`, with the API's English `message` as the fallback. The lookup is fallback-first, so this is safe to ship incrementally:

```
errorCode → look up `error.code.<errorCode>` key
  ├─ key exists  → translate (resource fills {{vars}})
  └─ key missing → fall back to the API's English message
```

_**Note**: Any BE code without a FE key still renders the backend's English text — nothing breaks, nothing shows blank._

**Scope (round 1 = the error path):**
- `<Trans>` exposed from the `uiSrc/i18n` barrel (keeps mid-sentence links translatable).
- `getTranslatedApiError` lookup + English fallback; `resource` carries interpolation vars.
- `parseCustomError` migrated case-by-case to `error.code.*` keys.
- Error notification factories (`error-messages.tsx`) + the 5 `*-error-content`
  components migrated to
- `error.code.*` / `error.*` keys added to `en.json` + `bg.json` (Bulgarian fully translated).

**Out of scope (follow-ups):**
- Success/notification toasts (`success-messages.tsx`, `toast.*`, `InfiniteMessages.tsx`) — round 2.
- Translations for the remaining BE error codes — additive, data-only (just `en`/`bg` key pairs), no code change.

# Testing

All migrated error toasts can be triggered locally without cloud/RDI/Azure setup.
Temporarily drop this dev helper into `App.tsx` (it routes through the real
`getAxiosError → parseCustomError → addErrorNotification` path), then call it from
the browser console:

```ts
// App.tsx — dev-only, do not commit. Requires imports:
//   import { getAxiosError } from 'uiSrc/utils'  (apiResponse)
//   import { addErrorNotification } from 'uiSrc/slices/app/notifications'
//   import { ApiEncryptionErrors } from 'uiSrc/constants'
//   import { store } from 'uiSrc/slices/store'
if (process.env.NODE_ENV === 'development') {
  const testError: any = (errorCode: number, message?: string) =>
    store.dispatch(
      addErrorNotification(
        getAxiosError({ response: { data: { errorCode, message } } } as any),
      ),
    )
  testError.codes = [
    11000, 11001, 11002, 11003, 11004, 11005, 11006, 11007, 11008, 11011,
    11022, 11024, 11108, 11401, 11404,
  ]
  testError.all = () =>
    testError.codes.forEach((c: number, i: number) =>
      setTimeout(() => testError(c), i * 1200),
    )
  // Toasts routed by name/flag, not errorCode:
  testError.encryption = () =>
    store.dispatch(
      addErrorNotification({
        response: { data: { name: ApiEncryptionErrors[0] } },
      } as any),
    )
  testError.persistent = (message = 'A persistent error example.') =>
    store.dispatch(
      addErrorNotification({
        response: { data: { message } },
        persistent: true,
      } as any),
    )
  ;(window as any).__testError = testError
  // eslint-disable-next-line no-console
  console.info(
    '[i18n] __testError(code[, msg]) | .encryption() | .persistent() | .all() | .codes',
  )
}
```

Then in the browser console:

```js
__testError(11022)      // a single code
__testError.encryption() // encryption toast (routed by error name)
__testError.persistent() // persistent toast (routed by flag)
__testError.all()        // fire every migrated code, 1.2s apart
```

Switch language with `?lang=bg` in the URL to verify the Bulgarian copy; unmapped
codes should still fall back to English.

---

### Cloud / OAuth errors

| Code | Case | EN | BG) |
|------|------|-------------|------------|
| 11000 | Server error | Screenshot 2026-06-26 at 17 45 42|Screenshot 2026-06-26 at 17 45 54|
| 11001 | Session expired |Screenshot 2026-06-26 at 17 46 36|Screenshot 2026-06-26 at 17 46 24| 
| 11002 | Access denied | Screenshot 2026-06-26 at 17 46 54|Screenshot 2026-06-26 at 17 47 03| 
| 11003 | Bad request |Screenshot 2026-06-26 at 17 47 36|Screenshot 2026-06-26 at 17 47 25 | 
| 11004 | Resource not found |Screenshot 2026-06-26 at 17 47 56|Screenshot 2026-06-26 at 17 48 08 | 
| 11005 | Misconfiguration |Screenshot 2026-06-26 at 17 49 16|Screenshot 2026-06-26 at 17 49 07 | 
| 11006 | GitHub email permission |Screenshot 2026-06-26 at 17 49 21 |Screenshot 2026-06-26 at 17 50 40 | 
| 11007 | Unknown authorization request |Screenshot 2026-06-26 at 17 51 16|Screenshot 2026-06-26 at 17 53 38 | 
| 11008 | Unexpected error |Screenshot 2026-06-26 at 17 51 32|Screenshot 2026-06-26 at 17 53 13 | 
| 11011 | Invalid email (SSO) |Screenshot 2026-06-26 at 17 57 28|Screenshot 2026-06-26 at 17 57 02 | 

### Errors with actions / links

| Code | Case | EN | BG |
|------|------|-------------|------------|
| 11022 | Invalid API key (Go to Settings / Remove API key buttons) | Screenshot 2026-06-26 at 18 04 15|Screenshot 2026-06-26 at 18 03 47 |
| 11024 | Azure session expired (Sign in to Azure button) |Screenshot 2026-06-26 at 18 04 22|Screenshot 2026-06-26 at 18 07 28 |
| 11108 | Database already exists (Cloud console link) |Screenshot 2026-06-26 at 18 09 00|Screenshot 2026-06-26 at 18 08 47 |
| 11401 | Pipeline not deployed (Review / Download log) |Screenshot 2026-06-26 at 18 09 18|Screenshot 2026-06-26 at 18 14 20 |
| 11404 | Validation error |Screenshot 2026-06-26 at 18 15 03|Screenshot 2026-06-26 at 18 14 54  |

### Toasts routed by name / flag

| Case |EN |BG |
|------|-------------|------------|
| Encryption (Disable Encryption / Cancel buttons) |Screenshot 2026-06-26 at 18 26 59|Screenshot 2026-06-26 at 18 28 16|
| Persistent (Copy button) |Screenshot 2026-06-26 at 18 29 05|Screenshot 2026-06-26 at 18 28 47|

---

Ref: RI-8268

Related community requests: #605, #1317, #1883, #2040, #2137, #2181, #2213, #2417, #2550, #2590, #2695, #2715, #2716, #2727, #2828, #2912, #2943, #2945, #3081, #3141, #3142, #3319, #3411, #3601, #3625, #3628, #3663, #3671, #3812, #3815, #4150, #4182, #4408, #4664, #4672, #4744, #5052, #5967

---

> [!NOTE]
> **Low Risk**
> User-facing copy and localization only, with explicit fallbacks to API English when keys are absent; no auth or data-path behavior changes beyond how error text is resolved.
> 
> **Overview**
> This PR **localizes the error notification path** by keying API failures on stable `errorCode` values (`api.error.code.<code>.*`) with English API text as fallback when a key is missing.
> 
> **Core plumbing:** `getTranslatedApiError` / `getTranslatedApiTitle` in `apiResponse.ts` resolve message and title from i18n (with `resource` interpolation); `addErrorNotification` uses them instead of raw response fields. `parseCustomError` replaces hard-coded Cloud/OAuth/RDI/Azure copy with i18n keys, adds a generic coded-error default, and uses **`Trans`** for the Cloud console link (11108). `CustomError` gains optional `resource`.
> 
> **UI:** Error toast factories and five `*-error-content` components (Azure, CAPI, encryption, persistent copy, RDI deploy) move button/label strings to `useTranslation` / `i18n.t`. `en.json` and `bg.json` add the new `api.error.*` and `notification.error.*` keys.
> 
> **Tests:** `errors.spec.tsx` switches from deep JSX equality to rendered substring checks so i18n-driven markup stays stable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e56fbf27d4339341237fad54bde9a10519b16c5d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redisinsight/ui/src/utils/errors.tsx
```diff
@@ -269,17 +230,21 @@ export const parseCustomError = (
       break
 
     case CustomErrorCodes.AzureEntraIdTokenExpired:
-      title = 'Azure session expired'
-      message =
-        err?.message ||
-        'Azure Entra ID token expired. Sign in to Azure again to continue.'
+      title = i18n.t('error.code.11024.title')
+      message = err?.message || i18n.t('error.code.11024.message')
```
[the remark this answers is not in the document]
valkirilov: Good catch on the symptom, but translation-first would regress a real case. errorCode 11024 carries two different backend messages — the default one and a more specific autodiscovery message (azure-autodiscovery.service.ts:361). The err?.message || fallback is what surfaces that specific text; preferring the translation would flatten both into the generic string. Same goes for 11401 (RDI), where err.message is the actual deploy-error log.

The root issue is one errorCode mapping to multiple English messages, which a single translation key can't localize. That needs distinct codes on the backend — I'll track it as a follow-up to RI-8267 rather than fold it into this PR. Until then it falls back to the backend's English, so nothing breaks.

redisinsight/ui/src/utils/errors.tsx
```diff
@@ -54,209 +55,169 @@ export const parseCustomError = (
...
       )
       break
 
     case CustomErrorCodes.RdiDeployPipelineFailure:
-      title = 'Pipeline not deployed'
-      message =
-        err?.message ||
-        'Unfortunately we’ve found some errors in your pipeline.'
+      title = i18n.t('error.code.11401.title')
+      message = err?.message || i18n.t('error.code.11401.message')
```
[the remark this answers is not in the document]
valkirilov: Good catch on the symptom, but translation-first would regress a real case. errorCode 11024 carries two different backend messages — the default one and a more specific autodiscovery message (azure-autodiscovery.service.ts:361). The err?.message || fallback is what surfaces that specific text; preferring the translation would flatten both into the generic string. Same goes for 11401 (RDI), where err.message is the actual deploy-error log.

The root issue is one errorCode mapping to multiple English messages, which a single translation key can't localize. That needs distinct codes on the backend — I'll track it as a follow-up to RI-8267 rather than fold it into this PR. Until then it falls back to the backend's English, so nothing breaks.

redisinsight/ui/src/utils/apiResponse.ts
```diff
@@ -43,6 +45,23 @@ export function getApiErrorName(error: AxiosError): string {
...
+  const key = `notification.error.code.${data?.errorCode}.message`
+
+  if (data?.errorCode && i18n.exists(key)) {
+    return i18n.t(key as never, { ...(data.resource ?? {}) })
+  }
+
+  const apiMessage = getApiErrorMessage(error)
+  return apiMessage === DEFAULT_ERROR_MESSAGE
+    ? i18n.t('notification.error.default')
+    : apiMessage
```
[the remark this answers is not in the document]
valkirilov: By design. The contract is: BE provides a stable `errorCode`, we translate by that code (with `resource` interpolation when needed), and fall back to the response `message` only when the code is unmapped. Preferring the backend message for mapped codes would invert that contract.

For the handful of codes that carry request-specific detail (RDI deploy log, Azure autodiscovery reason), the right fix is on the backend — distinct codes or structured `resource` params — not a per-code exception on the frontend. Tracking that as an RI-8267 follow-up.

redisinsight/ui/src/slices/app/notifications.ts
```diff
@@ -58,7 +58,7 @@ const notificationsSlice = createSlice({
       const { instanceId } = errorPayload
       const title = errorPayload?.response?.data?.title
       const errorName = getApiErrorName(payload)
-      const message = getApiErrorMessage(payload)
+      const message = getTranslatedApiError(payload)
```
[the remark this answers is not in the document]
valkirilov: Fixed in a7eb83f0b. Added `getTranslatedApiTitle` — translate the title by `errorCode`, else fall back to the response title (mirroring `getTranslatedApiError`) — and localized the default heading via `notification.error.title.default`. Raw coded errors no longer render a translated body under an English "Error" heading.

redisinsight/ui/src/utils/apiResponse.ts
```diff
@@ -43,6 +45,36 @@ export function getApiErrorName(error: AxiosError): string {
...
 
+// Translate a backend error by its stable `errorCode` (`notification.error.code.<code>`),
+// filling interpolation vars from `resource`. Falls back to the API's English
+// `message` when the code is unmapped.
+export function getTranslatedApiError(error: AxiosError): string {
+  const data = error?.response?.data as CustomError | undefined
+  const key = `notification.error.code.${data?.errorCode}.message`
+
+  if (data?.errorCode && i18n.exists(key)) {
+    return i18n.t(key as never, { ...(data.resource ?? {}) })
```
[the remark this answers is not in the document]
valkirilov: Ignoring this for now — it's the same multi-message-code limitation tracked as a backend follow-up (specifics should come as distinct codes or resource params). Falls back to the localized generic copy until then.

pawelangelow (dismissed):

redisinsight/ui/src/i18n/locales/en.json
```diff
@@ -1,11 +1,59 @@
 {
+  "notification.error.button.copied": "Copied",
```
KrumTy, line 37: I think this json will get very large; we could split it if needed later
But it's possible that the agent adds duplicated keys, especially if there's no proper order. 
Maybe as a side task it's worth formatting these files somehow (keys ordered alphabetically) and also validating - no duplicate keys allowed; ci check

redisinsight/ui/src/utils/apiResponse.ts
```diff
@@ -43,6 +45,36 @@ export function getApiErrorName(error: AxiosError): string {
   return get(error, 'response.data.name', 'Error') ?? ''
 }
 
+// Translate a backend error by its stable `errorCode` (`notification.error.code.<code>`),
+// filling interpolation vars from `resource`. Falls back to the API's English
+// `message` when the code is unmapped.
+export function getTranslatedApiError(error: AxiosError): string {
+  const data = error?.response?.data as CustomError | undefined
+  const key = `notification.error.code.${data?.errorCode}.message`
```
KrumTy: I wonder if "notification" is the right namespace for all api related errors; it will probably be mixed with other (ui) errors.

redisinsight/ui/src/utils/apiResponse.ts
```diff
@@ -43,6 +45,36 @@ export function getApiErrorName(error: AxiosError): string {
...
+
+  if (data?.errorCode && i18n.exists(key)) {
```
KrumTy, line 55: data?.errorCode check seems redundant but I get the order/point

redisinsight/ui/src/utils/errors.tsx
```diff
@@ -54,209 +55,169 @@ export const parseCustomError = (
...
       )
       break
 
     case CustomErrorCodes.RdiDeployPipelineFailure:
-      title = 'Pipeline not deployed'
-      message =
-        err?.message ||
-        'Unfortunately we’ve found some errors in your pipeline.'
+      title = i18n.t('notification.error.code.11401.title')
+      message = err?.message || i18n.t('notification.error.code.11401.message')
```
KrumTy: so skips translation if `err?.message` exists (same bellow)?

redisinsight/ui/src/i18n/locales/en.json
valkirilov: > I think this json will get very large; we could split it if needed later

+1, yesterday I was thinking the same, this file might get quite big very fast, so yes - such a split might be an option 👍 

> But it's possible that the agent adds duplicated keys, especially if there's no proper order.

Yes, that's true. For some cases, it might make sense to have duplicates for the keys, because in one language they might look the same, but in another language, to have differences in the translation, because of the context in which the string is used. 

But anyway, it's true that we should be careful with the duplication - such an automatic check might be a good idea, I'll take a look what can we do in this direction.

> Maybe as a side task it's worth formatting these files somehow (keys ordered alphabetically)

This one I didn't manage to find the proper balance, because:
1. I like the keys sorted, by their purpose/namespace.
2. But if I have this custom order, every time we run the command to extract the keys, they get reordered and the file shows a huge diff (which is wrong, because only the keys are reordered).

So, for now, I left the default order to be the one produced by the command, although it looks messy, but I'm open for ideas on handling this other way. 

Maybe, if we split it into multiple files, it might not be that big of a problem 🤔

redisinsight/ui/src/utils/apiResponse.ts
valkirilov: Good point. Initially, all these errors were namespaced without the `notification` prefix, but since we introduced such translations for the [success and infinite toast messages](https://github.com/redis/RedisInsight/pull/6126), I decided to move the errors under the same namespace for consistency. So in the end, we have:
- `notification.error.*`
- `notification.success*`
- `notification.infinity*`

But I'm ok to change the hierarchy for the errors, if you think it might cause confusion in the future :)

redisinsight/ui/src/utils/errors.tsx
valkirilov: Yes, but only temporarily. 

The agent found that we have BE errors with some parameters interpolated as part of the message (like URLs, Key Names, etc.). So in order to translate these, we should pass them as props (in the API responses), so we can interpolate them properly while translating. 

```
{ 
    "message": "The database already exists.", 
    "statusCode": 409, 
    "error": "DatabaseAlreadyExists", 
    "errorCode": 11200, 
    "resource": { 
        "databaseId": "a1b2c3d4-..." 
    } 
}
```

But turns out we need to apply this change to a lot of places, so I decided to leave it for a separate PR and keep this one clean. This way, the English version still works as expected (a.k.a we don't introduce regressions, hopefully 🤞)

redisinsight/ui/src/utils/apiResponse.ts
valkirilov: Correct, the agent went too cautious here. It doesn't bring any additional value, it just makes it easier to read and follow the logic 👍

redisinsight/ui/src/i18n/locales/en.json
valkirilov: As per your advice, here is a [pull request](https://github.com/redis/RedisInsight/pull/6131) with a CI check that will watch for duplicate keys

redisinsight/ui/src/utils/apiResponse.ts
valkirilov: Made the changes we discussed so all API messages are prefixed properly. For now, we'll have only the errors, but it can be easily extended with success messages in case we need that.

Changes in 779312a71fae7ff992a5e9a781470b5eef5e2252

KrumTy (dismissed):

pawelangelow (dismissed):

redisinsight/ui/src/components/notifications/hooks/useErrorNotifications.ts
```diff
@@ -45,7 +44,7 @@ export const useErrorNotifications = () => {
         message = DEFAULT_ERROR_MESSAGE,
         instanceId = '',
         name,
-        title = DEFAULT_ERROR_TITLE,
+        title = i18n.t('notification.error.title.default'),
```
KrumTy: You can use i18n from useTranslation() hook here so that if language is changed runtime this will retrigger (as opposed to using the global i18n instance)

There are other places in the PR that import and use i18n directly - worth checking them as well

valkirilov: Good call — applied: `useErrorNotifications` now uses `t` from `useTranslation()`. e56fbf27d4339341237fad54bde9a10519b16c5d

Checked the other direct-`i18n` spots in this PR: the error-content components already use `useTranslation()`, and `error-messages.tsx` / `apiResponse.ts` / `errors.tsx` use the global instance out of necessity — they are a factory object and utils invoked imperatively from the redux slice, not React components/hooks, so the hook is not usable there. Will keep this in mind for the rest of the PRs in the chain.

redisinsight/ui/src/slices/app/notifications.ts
```diff
@@ -56,9 +57,9 @@ const notificationsSlice = createSlice({
     ) => {
       const errorPayload = payload as IAddInstanceErrorPayload
       const { instanceId } = errorPayload
-      const title = errorPayload?.response?.data?.title
+      const title = getTranslatedApiTitle(payload)
       const errorName = getApiErrorName(payload)
-      const message = getApiErrorMessage(payload)
+      const message = getTranslatedApiError(payload)
```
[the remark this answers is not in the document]
valkirilov: Fair point, but in practice this is a non-issue for toasts. Non-persistent error toasts auto-dismiss (`autoClose` default), so they are gone well before a user can navigate to settings and switch language — they will never see a transient toast in a stale locale. The only ones that linger are the few persistent toasts (`autoClose: false` — persistent/Azure-token-expired/encryption errors), and a slightly stale label on one of those until it is dismissed/re-triggered is cosmetic. A runtime language switcher is coming, but it will need to touch this notification path anyway, so we will revisit re-translation of live toasts there rather than refactor the reducer now.

KrumTy (approved):

PR #6143, opened 2026-07-01
feat(ui): Add value decoder

saidake:
# What
Add **Value Decoders** for HASH value tables. The initial implementation only includes a **Binary Decoder**.

This PR focuses on solving a single problem: ***making complex Redis values human-readable***.

* Users can create Value Decoders for specific Redis keys by defining a **Key Pattern**.
* A Decode (eye) button is displayed in the `Value` column header when any `Key Pattern` matches.
* With the **Binary Decoder**, users can configure the data type of binary fields in advance. Redis Insight will then decode binary values and display them in a human-readable format in the value preview table.
* Redis Strings are binary-safe byte arrays, so storing binary data in Redis is a very common practice.
* The **Binary Decoder** is especially useful for games and other high-performance systems that store binary data in Redis, making it much easier to inspect and debug binary values directly in Redis Insight.
* Value Decoder data is stored in Local Storage using the key ·valueDecoderRules_<instance-id>·, allowing each Redis database connection to maintain its own isolated Value Decoders.  
  when you delete a Redis instance through the UI, its value decoders are removed from Local Storage 
Currently, this functionality is heavily used in my own system. In the future, the **Value Decoders** could be extended to support additional formats as needed, such as *Protobuf*, *MessagePack*, and *BSON*.

* Will this feature have any performance impact?
  No. `HashDetailsTable.tsx` uses `VirtualTable` (`react-virtualized`) to render items, so only visible items are rendered and decoded.
* Will this feature misdisplay binary data?
  No. The eye toggle icon allows you to switch between the decoded view and the original value at any time.
  
Usage Example:
image
image
image
Stored Local Storage Data Example:
image

# Testing

1. Start UI and API server.
  
  Start the Redis Insight UI:
   ```bash
   yarn dev:ui
   ```  
  Start the Redis Insight API server:
   ```bash
   yarn dev:api
   ```

2. Create test data.  

Connect to a Redis database A. In a Linux terminal, create a HASH containing binary values:
```bash
printf '\x01\x88\x77\x66\x55\x44\x33\x22\x11\x02\x00\x11\x11\x11\x11\x11\x11\x11\x11\x22\x22\x22\x22\x22\x22\x22\x22\x7B\x00\x00\x00\x33\x33\x33\x33\x33\x33\x33\x33\x44\x44\x44\x44\x44\x44\x44\x44\xC8\x01\x00\x00' | redis-cli -x HSET example:hash field1
```
Connect to a Redis database B. In a Linux terminal, create a HASH containing binary values:
```bash
printf '\x01\x05\x00\x52\x65\x64\x69\x73\x18\x2D\x44\x54\xFB\x21\x09\x40\x08\x00\xDE\xAD\xBE\xEF\xCA\xFE\xBA\xBE' | redis-cli -x HSET example:hash field1
```
3. In Redis Insight, connect to Redis database A, open the `example:hash` key, click **Value Decoders** and create a new decoder with the following configuration:
image

4. Click the **Decode** (eye) button for each HASH field. The decoded values should be:
image

5. In Redis Insight, connect to Redis database B, open the `example:hash` key, click **Value Decoders** and create a new decoder with the following configuration:
image

6. Click the **Decode** (eye) button for each HASH field. The decoded values should be:
image

---

> [!NOTE]
> **Medium Risk**
> Large new browser surface on hash value rendering and localStorage-backed config; decoding runs on visible virtualized rows with iteration caps, but wrong schemas could mislead debugging until users toggle raw view.
> 
> **Overview**
> Introduces a **Value Decoder** feature (behind the `valueDecoder` flag, off by default) so users can define per-database **binary layout rules** keyed by Redis key globs and read HASH field bytes in a structured way in the browser.
> 
> **Configuration & storage:** A modal configures multiple decoders (key patterns, binary field schema with repeats and dynamic sizes). Rules persist in local storage per connection (`valueDecoderRules_<instanceId>`), with legacy global key migration and cleanup when a database is deleted.
> 
> **HASH UI:** Hash key details wrap in `ValueDecoderProvider`; the header adds **Value Decoders** on HASH keys and an eye toggle on the Value column when a pattern matches. Visible rows can show decoded output via client-side parsing while keeping the existing formatted/raw view.
> 
> **Platform:** API/UI feature flags and config version bump register `valueDecoder`; unit tests cover parsing, validation, clipboard import/export, and storage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 48b072b52ad53726113f572081f34e614ae362aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pawelangelow (changes requested):
Thanks a lot for this, @saidake - really nice piece of work. 🙏 The design is careful in the ways that matter: it's off by default, the raw value is always one toggle away, the parser is defensive (bounds-checked, repeat caps, regex wrapped in try/catch), and it's cleanly isolated so it doesn't touch any other key type.

Before we can ship it, there are a few things that need to be in place. None of these are hard blockers on the idea - they're about fitting it into how RedisInsight works at large:

- Feature flag (required for shipping). Every other optional action in the key header is wrapped in a FeatureFlagComponent, and we need the same here so the feature can be rolled out gradually and disabled remotely if anything goes wrong in production. This is the one true blocker.

- Per-database storage, not a global key. Right now the rules live in a single global `valueDecoderRules` localStorage entry, so a decoder created for one database silently applies to every connection with matching key names. These need to be scoped per database so rules don't leak across connections.

- UI treatment. The current "Value Decoders" primary button shows for every hash key for every user, which is a lot of visual weight for an optional tool. We'd like to experiment here to match the existing header style - likely a lighter/secondary or icon button, and/or only surfacing it when the feature is enabled for that database.

In terms of next steps, there are two ways we can go:

- You keep ownership and work on these - in which case we'd just ask that you try to scope them into small, focused PRs. That speeds up the review process a lot and gets things merged faster.
- We handle them on our side - happy to, though we can't guarantee you an exact timeline for when we'd get to it.

Either way works for us. Really appreciate the contribution - the core is in good shape, this is mostly about integration. Let us know which direction you'd prefer!

saidake:
hi @pawelangelow 
I'm happy to keep ownership of these. I've updated the PR. Please take another look when you have a chance.

pawelangelow:
Thanks a lot for the idea and the contribution, @saidake! 🙏 We're going to merge this in.

As next steps, we're planning to refactor FieldsSchemaEditor into smaller components, add tests, and then enable the feature flag for rollout. If you'd like to keep contributing, feel free to open PR(s) for any of those - happy to have you involved.

pawelangelow (approved):

saidake:
hi @pawelangelow.

Sure, I'll check it out. Thanks.

PR #6148, opened 2026-07-02
feat(connection): allow editing host when editing a connection

rokn:
Previously the host field was hidden entirely when editing a saved connection (shown only when adding or cloning). Show it in edit mode so the host can be changed. Cloud (Connect redirect) still hides host/port, and Azure-sourced databases keep host disabled via AZURE_READONLY_FIELDS.

# What

# Testing
Manual tested and verified.


valkirilov:
Thanks for your contribution @rokn. 

We're going to add a few more minor tweaks with https://github.com/redis/RedisInsight/pull/6155, but generally, we'll make the Host field editable. Thanks!

PR #6181, opened 2026-07-09
RI-8315: Edit array element values in a Monaco drawer

pawelangelow:
# What

- **Monaco value editor** — adds an "expand" affordance that opens the element
  value in a right-side **drawer** with a plaintext Monaco editor, giving large
  values much more room than the inline editor. Save reuses the existing ARSET
  apply path, behind the production-write confirmation. The inline textarea edit
  stays.
- **Grouped row actions** — moves the per-row **edit** and **expand** triggers
  out of the value cell and into the actions column next to **delete**, so long
  values are no longer hidden behind hover icons. A shared
  `getArrayElementEditState` helper backs both the value cell (display + inline
  editor) and the actions cell (trigger gating + drawer seed).
- **Truncation guard** — editing (inline + drawer) is disabled for
  backend-truncated values, mirroring the String details guard, so a truncated
  copy can't be saved back over the real element.
- Adds an opt-in `hideEditButton` to the shared `EditableTextArea`; default off,
  so list/hash/other consumers are unchanged.

# Preview

## Action buttons

Screenshot 2026-07-09 at 9 40 59

## Editor

Screenshot 2026-07-09 at 9 41 44

## What happens with very long values

Screenshot 2026-07-09 at 9 42 11

---

> [!NOTE]
> **Medium Risk**
> Touches production ARSET writes and connection/key-switch guards, but behavior is heavily covered by new tests and mirrors existing string-details patterns.
> 
> **Overview**
> Adds a **Monaco drawer** for array element values (expand in the actions column), with save going through the existing **ARSET** path and production-write confirmation. **Inline textarea editing** remains; inline and drawer are **mutually exclusive** and share refresh-pause, session, and abandon-on-tab/key/format rules.
> 
> **Edit** and **expand** move from the value cell into the **actions column** beside delete (`hideEditButton` on shared `EditableTextArea`). Shared **`getArrayElementEditState`** drives display, inline edit, trigger gating, and drawer seeding, including **blocking edits on backend-truncated** values.
> 
> **Safety around writes:** `updateArrayElementAction` accepts optional **`startInstanceId`** and skips ARSET if the connected database changed since the editor opened; key abandonment uses **live app-context selection** (not lagging `selectedKeyData`). Bulk delete and per-row delete are **frozen** during inline edit, drawer open, or in-flight ARSET.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9485e35de68ac4a500303f3ab404034c534a622e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redisinsight/ui/src/pages/browser/modules/key-details/components/array-details/array-details-table/components/ArrayValueEditorDrawer.tsx
```diff
@@ -0,0 +1,89 @@
...
+          <SecondaryButton
+            size="l"
+            onClick={onClose}
+            data-testid="array-value-editor-cancel-btn"
+          >
+            Cancel
+          </SecondaryButton>
+          <PrimaryButton
+            size="l"
+            onClick={() => onSave(value)}
```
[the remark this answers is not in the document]
pawelangelow: That's by design - Array element values are arbitrary Redis bulk strings - any byte string is valid, so there's nothing to "validate" for correctness.

redisinsight/ui/src/pages/browser/modules/key-details/shared/editable-textarea/EditableTextArea.tsx
```diff
@@ -105,7 +113,7 @@ const EditableTextArea = (props: Props) => {
         >
           {children}
         </Text>
-        {isHovering && (
+        {!hideEditButton && isHovering && (
```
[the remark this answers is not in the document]
pawelangelow: Thanks - acknowledged, but we're intentionally not changing this in this PR.

The close-on-confirm behaviour lives in EditableTextArea, which is shared by every key-type editor (hash, list, set, zset, string), so altering it has blast radius well beyond arrays and is out of scope here.

redisinsight/ui/src/pages/browser/modules/key-details/components/array-details/array-details-table/components/RowActionsCell/RowActionsCell.tsx
```diff
@@ -2,51 +2,110 @@ import React from 'react'
...
+                e.stopPropagation()
+                editConfig?.onOpenValueEditor(index)
+              }}
+              data-testid={`array-expand-btn-${index}`}
+            />
+          </RiTooltip>
+        </>
+      )}
+
+      {deleteConfig && !editConfig?.isValueDrawerOpen && (
```
VaskoAtanasovRedis: Deletes are frozen only while the drawer is open, not during an inline edit, so deleting the inline-edited element races its pending Save I think - so if you're editing a value inline the delete might be clickable and not while you're in the drawer

redisinsight/ui/src/pages/browser/modules/key-details/components/array-details/array-details-table/ArrayDetailsTable.tsx
```diff
@@ -85,13 +96,44 @@ const ArrayDetailsTable = memo(
...
     }
+    // The live selection — updated on key click, before fetchKeyInfo. Unlike
+    // `keyName` (from selectedKeyData) it doesn't lag a switch.
+    const liveSelectedKey = useAppSelector(appContextSelectedKey)
+
+    const { requestConfirmation } = useProductionWriteConfirmation()
 
     // Index of the row currently being edited; only one row edits at a time.
     const [editingIndex, setEditingIndex] = useState<Nullable<string>>(null)
+    // Index whose value is open in the Monaco drawer, plus its seed captured at
```
VaskoAtanasovRedis: nit: too much comments here I think

redisinsight/ui/src/pages/browser/modules/key-details/components/array-details/array-details-table/components/ArrayValueEditorDrawer.tsx
```diff
@@ -0,0 +1,91 @@
...
+            size="l"
+            onClick={onClose}
+            data-testid="array-value-editor-cancel-btn"
+          >
+            Cancel
+          </SecondaryButton>
+          <PrimaryButton
+            size="l"
+            disabled={isSaveDisabled}
+            onClick={() => onSave(value)}
```
VaskoAtanasovRedis, line 84: this directly saves without checking formatting of values, while inline does a `.isValid` check I think

pawelangelow: Good catch on the asymmetry, but I don't think it's worth aligning. Array element values are free-form — they can hold anything, so there's no real "valid/invalid" for the value itself.

The inline isValid check isn't actually validation either: it's a soft, overridable warning that only shows up when you're viewing in a structured format (JSON/Msgpack/PHP) and type something that doesn't parse as that format. Even then it just pops an "are you sure?" and lets you save anyway — it never blocks. For the default Unicode view (the common case) it never fires at all.

VaskoAtanasovRedis (approved):

PR #6192, opened 2026-07-13
Fix jest watch scoping and add typeahead filtering

pawelangelow:
# What

Improves the frontend jest watch-mode developer experience:

- **Scope jest via `roots` instead of a positional path arg.** The npm
  scripts passed `./redisinsight/ui` as a test-path pattern, which in
  `--watch` forced a full-suite run on startup. Moving the scoping into
  `jest.config.cjs` (`roots`) means `yarn test:watch` now idles on a clean
  tree and only reruns tests related to what you edit. Same 858 tests are
  discovered — no change to `yarn test` / `test:cov`.
- **Stop crawling `redisinsight/api/dist`** as a side effect, which removes
  the duplicate-manual-mock haste-map warnings on startup.
- **Add `jest-watch-typeahead`** (pinned `^2.2.2` for jest 29) for fuzzy
  filename / test-name filtering via the `p` / `t` prompts in watch mode.

# Preview

## Before (lots of warnings and you can't type `p` to select a particular test)

https://github.com/user-attachments/assets/dd83ee3f-1102-47e1-a05c-e78a345e0254

## After

https://github.com/user-attachments/assets/6681d29f-0e99-4ca8-9a80-e58cfc97e36d

---

> [!NOTE]
> **Low Risk**
> Dev-only Jest configuration and script changes; no production runtime impact, with the main caveat being accidental narrowing of test discovery if `roots` is wrong.
> 
> **Overview**
> **Jest watch and discovery** are tightened so local frontend test runs behave better without changing which tests CI runs.
> 
> Test scoping moves from passing `./redisinsight/ui` on the CLI into **`roots`** in `jest.config.cjs` (`ui`, `__mocks__`, `api-client`). Root `test`, `test:watch`, and coverage scripts now invoke plain `jest`, relying on that config. **Watch mode** should idle on a clean tree instead of running the full suite on startup, and discovery should avoid crawling **`redisinsight/api/dist`**, which removes duplicate manual-mock haste-map warnings.
> 
> **`jest-watch-typeahead`** is added with `watchPlugins` for fuzzy **filename** (`p`) and **test name** (`t`) filtering in watch mode.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1c6f5dee8907b9bece86b8ea5f125f60f961819. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


valkirilov (dismissed):

jest.config.cjs
```diff
@@ -2,6 +2,15 @@ require('dotenv').config({ path: './redisinsight/ui/.env.test' });
 
 /** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
 module.exports = {
+  // Scope test/haste discovery to the UI workspace (plus shared manual mocks)
+  // via `roots` rather than a positional path arg, so `--watch` idles on a
+  // clean tree and jest doesn't crawl redisinsight/api/dist.
+  roots: ['<rootDir>/redisinsight/ui', '<rootDir>/redisinsight/__mocks__'],
```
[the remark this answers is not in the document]
KrumTy: I think this is valid. If there are changes in api (e.g. enum value of some type changed), then api-client needs to be regenerated and it could cause regression

valkirilov (dismissed):

jest.config.cjs
```diff
@@ -1,7 +1,30 @@
...
+  '<rootDir>/redisinsight/ui',
+  '<rootDir>/redisinsight/__mocks__',
+];
+
+// Also watch the generated API client so watch reruns dependent tests when it
+// changes. It can be missing (generation skipped or failed), and jest aborts
+// on a missing root, so add it only when present.
+if (fs.existsSync(path.join(__dirname, 'redisinsight/api-client'))) {
+  roots.push('<rootDir>/redisinsight/api-client');
+}
```
KrumTy: why dynamic push? If api-client is missing then unit tests will fail anyway (components import enums from there)

pawelangelow: actually, you're right - these should be both available on dev environment and CI, so no need to dynamically import it.

KrumTy (approved):

PR #6217, opened 2026-07-16
Split FieldsSchemaEditor into smaller components

saidake:
# What

Refactors the value decoder `FieldsSchemaEditor` into smaller components.
- Extracted field editing UI into `FieldRow` (including size-source handling)
- Extracted repeat-block chrome into `RepeatBlockEditor`
- Kept `FieldsSchemaEditor` as the orchestration layer for add/remove/reorder and nested editing

Increase the timeout of a test in `redisinsight/ui/src/pages/browser/BrowserPage.test.tsx`, as I encountered a UI CI timeout error:
```
FAIL redisinsight/ui/src/pages/browser/BrowserPage.test.tsx (72.576 s)
  ● KeyDetailsHeader › Verify that user cannot rename key name with unprintable characters and check tooltip

    thrown: "Exceeded timeout of 5000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      184 |   })
      185 |
    > 186 |   it('Verify that user cannot rename key name with unprintable characters and check tooltip', async () => {
          |   ^
      187 |     const { queryByTestId } = render(<BrowserPage />)
      188 |
      189 |     const afterRenderActions = [...store.getActions()]

      at it (redisinsight/ui/src/pages/browser/BrowserPage.test.tsx:186:3)
      at Object.describe (redisinsight/ui/src/pages/browser/BrowserPage.test.tsx:162:1)
```

# Testing

The Value Decoder field schema continues to work as before:
image

image

---

> [!NOTE]
> **Low Risk**
> UI-only refactor of the value decoder schema editor with extracted pure mutation helpers and new unit tests; no API or security surface changes.
> 
> **Overview**
> **Refactors** the value decoder `FieldsSchemaEditor` by pulling UI and state logic out of one large file while keeping behavior the same.
> 
> Field editing moves to **`FieldRow`** and **`FieldSizeEditor`** (name, type, fixed vs custom size, size-from-field refs). Repeat blocks use **`RepeatBlockEditor`**, which renders nested schema editing via **`children`** instead of calling `FieldsSchemaEditor` from inside the repeat component. Add/remove/reorder and field/repeat updates go through **`useSchemaEditor`** and exported pure **`apply*`** helpers (covered by **`useSchemaEditor.spec.ts`**).
> 
> **`toNumericOptions`** (numeric ref select labels, including duplicate-name disambiguation) is centralized in **`schemaUtils`** with unit tests. **`BrowserPage.test.tsx`** gets a **10s** timeout on one tooltip test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c91ce39dafbf39e5e9a736d227e5b5d8d43fa0b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redisinsight/ui/src/pages/browser/components/value-decoder/FieldsSchemaEditor.tsx
pawelangelow: The `handle*` functions are all schema-mutation logic - I'd extract them into a `useSchemaEditor` hook so the mutation logic lives separately from the rendering. The nice side effect is that the hook can then be unit-tested directly (add/remove/reorder/nested-repeat) without rendering the editor, which is hard to do while the logic sits inside the component.

redisinsight/ui/src/pages/browser/components/value-decoder/RepeatBlockEditor.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }, {})
+
+  return fields.map((item) => ({
+    value: item.id,
+    label:
+      nameCounts[item.name] > 1
+        ? `${item.name} (${item.dataType}) · ${item.id}`
+        : `${item.name} (${item.dataType})`,
+  }))
+}
```
pawelangelow: This looks like duplicated logic

redisinsight/ui/src/pages/browser/components/value-decoder/FieldRow.tsx
```diff
@@ -0,0 +1,183 @@
...
+          <S.SizeUnit
+            data-testid={`${VALUE_DECODER_TEST_ID}-field-size-unit-${field.id}`}
+          >
+            {getSizeUnit(field.size)}
+          </S.SizeUnit>
+        </S.SizeInputWrapper>
+      )}
+    </S.SizeSourceWrapper>
+  )
+}
```
pawelangelow: Nice to see this pulled out of the nested ternary. Since it's a self-contained component with the trickiest branching (fixed / custom / field-ref), could we give it its own `FieldSizeEditor.tsx` file? That matches the one-component-per-file split the rest of this PR is going for, and keeps `FieldRow` focused on the row layout.

redisinsight/ui/src/pages/browser/components/value-decoder/FieldRow.tsx
```diff
@@ -0,0 +1,183 @@
...
+        data-testid={`${VALUE_DECODER_TEST_ID}-field-type-${field.id}`}
+      />
+      <div>
+        <FieldSizeEditor
+          field={field}
+          sizeSource={sizeSource}
+          sizeRefs={sizeRefs}
+          onFieldChange={onFieldChange}
+        />
+      </div>
```
pawelangelow: This `<div>` is acting as the grid cell for the size column. Since `FieldSizeEditor` already renders a single root element in both branches, could it be the grid child directly and drop the wrapper? If a wrapper is needed, a plain <div> here trips the "layout components over div" convention - though `FlexGroup`/`Row` would force flexbox onto what's really just a grid cell, so removing it is probably cleaner than swapping it.

pawelangelow (changes requested):
Overall looks good - nice clean split, and simple is better IMO. The one thing I'd like before we merge is tests for the new components and the extracted logic; that's really the win from isolating them like this.

saidake:
hi @pawelangelow 
The PR has been updated, please take a look.

pawelangelow (approved):

PR #6225, opened 2026-07-17
RI-8321: Enforce x-window-id auth on WebSocket connections

pawelangelow:
# What

The local WebSocket layer did not enforce the `x-window-id` authorization
that the HTTP API requires. `SessionMetadataAdapter` was registered as the
WebSocket adapter unconditionally after `WindowsAuthAdapter`, and NestJS keeps
only the last adapter set — so the auth gate never ran and every socket was
granted the default session. Unauthenticated cross-origin clients could reach
the `monitor`, `pub-sub`, and `bulk-actions` namespaces, observe live commands,
and trigger destructive bulk actions against saved databases.

`WindowsAuthAdapter` now extends `SessionMetadataAdapter`, so an authorized
socket gets both the window-id check and its session metadata, and it is the
only adapter wired for the desktop build. Web/dev builds keep the
metadata-only adapter, matching the Electron-only scope of the HTTP
`WindowAuthMiddleware`.

---

> [!NOTE]
> **High Risk**
> Changes WebSocket authentication and connection teardown for the desktop build, closing a path where unauthenticated clients could use monitor/pub-sub/bulk-actions namespaces.
> 
> **Overview**
> **Fixes a WebSocket auth bypass** where registering `SessionMetadataAdapter` after `WindowsAuthAdapter` in bootstrap meant only the metadata adapter ran, so every socket got default session without an `x-window-id` check.
> 
> For **Electron** builds, bootstrap now registers **only** `WindowsAuthAdapter` (with an inline comment). **Web/dev** builds use `SessionMetadataAdapter` alone, matching HTTP window-auth scope.
> 
> `WindowsAuthAdapter` **extends** `SessionMetadataAdapter` and calls `super.bindMessageHandlers` after a successful window-id check, so authorized sockets still get session metadata and user-room join. **Unauthorized** sockets are **`disconnect(true)`** so they cannot keep receiving namespace broadcasts.
> 
> Adds **unit tests** for authorized vs unauthorized `bindMessageHandlers` behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b87342f388d67333c760470ac9857ba3581f025b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisinsight/api/src/main.ts
+++ b/redisinsight/api/src/main.ts
@@ -95,12 +95,13 @@ export default async function bootstrap(apiPort?: number): Promise<IApp> {
         },
       },
     );
+
+    app.useWebSocketAdapter(new SessionMetadataAdapter(app));
   } else {
     app.setGlobalPrefix(serverConfig.globalPrefix);
+    // Must be the only web socket adapter here or the window-id auth gate is lost.
     app.useWebSocketAdapter(new WindowsAuthAdapter(app));
   }
-
-  app.useWebSocketAdapter(new SessionMetadataAdapter(app));
 
   const logFileProvider = app.get(LogFileProvider);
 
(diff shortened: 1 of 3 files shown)


pawelangelow:
@codex review

valkirilov (approved):

PR #6230, opened 2026-07-17
fix(instance): connect on deep-link to refresh modules

### LINKED ISSUE #6080: [Bug]: Docker Redis Insight does not detect Redis Search when opening a preconfigured database through a direct route
## **Preconditions**

Docker Redis Insight is started as a fresh instance.

The Redis database connection is preconfigured through Docker environment variables:

`RI_REDIS_HOST=redis`

`RI_REDIS_PORT=6379`

`RI_REDIS_ALIAS=Workshop Redis`

`RI_ACCEPT_TERMS_AND_CONDITIONS=true`

The target Redis database has Redis Search enabled.

## **Steps to reproduce**

1. Start Docker Redis Insight with a preconfigured Redis database connection.

2. Open Redis Insight directly at:

`http://localhost:5540/0/vector-search`

3. Observe the Search page.

4. Go back to the database picker:

`http://localhost:5540/`

5. Click the same preconfigured database.

6. Open the Search tab again.

## **Actual behavior**

When Docker Redis Insight is opened directly at `/0/vector-search`, the Search page says:

`Redis Query Engine is not available for this database`

After entering the same preconfigured database through the database picker, the Search page loads correctly.

## **Expected behavior**

Docker Redis Insight should detect Redis Search for the preconfigured database when opening `/0/vector-search` or `/0/browser` directly.

## **Additional context**

Host OS: macOS 15.7.4

Redis Insight distribution: Docker version

Redis Insight version: 3.6.0

Redis Insight build type: Docker on premise

Redis Insight image: `redis/redisinsight:latest`

Image SHA: `sha256:aa21bbd198455b4ad964f76782db951155aa0d712321f599972d1525f031f0e6`

Redis database version: 8.8.0

Database type: Standalone

Connection type: Standalone

Redis Search module version: 80800

Redis JSON module version: 80800

Redis TimeSeries module version: 80800

No browser console errors were observed during reproduction.

trippyogi:
## Summary
- Call `checkConnectToInstanceAction` when mounting `InstancePage`, matching the DB picker / Redis Stack entry path.
- Refresh modules via `GET /databases/:id/connect` before fetching instance data used by Vector Search guards.
- Fixes Docker preconfigured databases (`modules: []`) opened via `/0/vector-search` incorrectly showing Redis Search as unavailable.

Fixes #6080

## Test plan
- [x] `InstancePage.spec.tsx` — 9/9 passed
- [x] Fresh preconfigured DB with `modules: []`; deep-link `/0/vector-search` triggers connect and populates `search` module
- [x] UI: deep-link Search shows indexes (e.g. `idx:trie`), not the RQE unavailable empty state
- [x] `/0/browser` still loads after deep-link / picker entry
- [x] Regression: open DB from picker, then Search — still works

---

> [!NOTE]
> **Medium Risk**
> Changes core instance mount/connect flow and when Redux state is reset; mistakes could cause double connects, missed module refresh, or routing bounce on DB switch, though behavior is heavily covered by new tests.
> 
> **Overview**
> **Fixes deep-linked Vector Search** when preconfigured DBs have empty `modules` by aligning `InstancePage` with the DB picker path: it calls `checkConnectToInstanceAction` ( `GET /databases/:id/connect` ) before loading instance data, unless Redux already has the same `connectedInstance.id`.
> 
> When the route id matches the connected id, it **skips connect** and only runs the instance fetch/config/info/recommendations bundle. On connect success or failure it still runs that bundle so `connectedInstance.loading` clears and Vector Search does not spin forever. **Reset** (`resetConnectedInstance` + context wipe) runs only when switching from a different connected DB, not on fresh deep links or Redis Stack pre-set id.
> 
> `fetchConnectedInstanceAction` now dispatches **`setConnectedInstanceFailure`** on API errors. MSW exposes **`connectDatabaseApiSpy`**; `InstancePage.spec.tsx` adds coverage for same-id skip, connect success/fail, DB switch reset, and single connect on deep link.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89397ce3338fc5f6dcc7b717cf15c3d987d66a6d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisinsight/ui/src/pages/instance/InstancePage.spec.tsx
+++ b/redisinsight/ui/src/pages/instance/InstancePage.spec.tsx
@@ -39,10 +39,8 @@ import {
   setInitialRecommendationsState,
 } from 'uiSrc/slices/recommendations/recommendations'
 import {
-  getDatabaseConfigInfo,
   loadInstances,
-  setConnectedInfoInstance,
-  setConnectedInstance,
+  resetConnectedInstance,
   setDefaultInstance,
 } from 'uiSrc/slices/instances/instances'
 import * as rdiInstanceSlice from 'uiSrc/slices/rdi/instances'
@@ -157,10 +155,7 @@ describe('InstancePage', () => {
       loadRdiInstances(),
       getAllPlugins(),
       setDefaultInstance(),
-      setConnectedInstance(),
-      getDatabaseConfigInfo(),
-      setConnectedInfoInstance(),
-      getRecommendations(),
+      resetConnectedInstance(),
       ...resetContextActions,
       clearExpertChatHistory(),
       setConnectivityError(null),
--- a/redisinsight/ui/src/pages/instance/InstancePage.spec.tsx
+++ b/redisinsight/ui/src/pages/instance/InstancePage.spec.tsx
@@ -121,15 +121,27 @@ describe('InstancePage', () => {
       contextInstanceId: 'prevId',
     })
 
+    // Seed a different already-connected DB so InstancePage resets on switch
+    // (Redis Stack keeps the same id and must not reset).
+    const initialState = set(
+      cloneDeep(initialStateDefault),
+      'connections.instances.connectedInstance.id',
+      'prevId',
+    )
+    const testStore = mockStore(initialState)
+
     // Flush pending async thunks leaked from previous test renders
     await act(async () => {})
-    store.clearActions()
+    testStore.clearActions()
 
     await act(() => {
       render(
         <BrowserRouter>
           <InstancePage {...instance(mockedProps)} />
         </BrowserRouter>,
+        {
+          store: testStore,
+        },
       )
     })
 
@@ -163,9 +175,39 @@ describe('InstancePage', () => {
       setDbConfig(undefined),
     ]
 
-    expect(store.getActions().slice(0, expectedActions.length)).toEqual(
+    expect(testStore.getActions().slice(0, expectedActions.length)).toEqual(
       expectedActions,
     )
+  })
+
+  it('should not reset connected instance when already connected to same id', async () => {
+    ;(appContextSelector as jest.Mock).mockReturnValue({
+      contextInstanceId: '',
+    })
+
+    const initialState = set(
+      cloneDeep(initialStateDefault),
+      'connections.instances.connectedInstance.id',
+      INSTANCE_ID_MOCK,
+    )
+    const testStore = mockStore(initialState)
+
+    await act(async () => {})
+    testStore.clearActions()
+
+    await act(() => {
+      render(
+        <BrowserRouter>
+          <InstancePage {...instance(mockedProps)} />
+        </BrowserRouter>,
+        {
+          store: testStore,
+        },
+      )
+    })
+
+    expect(testStore.getActions()).not.toContainEqual(resetConnectedInstance())
+    expect(testStore.getActions()).toContainEqual(setDefaultInstance())
   })
 
   it('should call databases list api', async () => {
--- a/redisinsight/ui/src/pages/instance/InstancePage.spec.tsx
+++ b/redisinsight/ui/src/pages/instance/InstancePage.spec.tsx
@@ -34,10 +34,7 @@ import {
 import { setClusterDetailsInitialState } from 'uiSrc/slices/analytics/clusterDetails'
 import { setDatabaseAnalysisInitialState } from 'uiSrc/slices/analytics/dbAnalysis'
 import { setInitialAnalyticsSettings } from 'uiSrc/slices/analytics/settings'
-import {
-  getRecommendations,
-  setInitialRecommendationsState,
-} from 'uiSrc/slices/recommendations/recommendations'
+import { setInitialRecommendationsState } from 'uiSrc/slices/recommendations/recommendations'
 import {
   loadInstances,
   resetConnectedInstance,
--- a/redisinsight/ui/src/pages/instance/InstancePage.spec.tsx
+++ b/redisinsight/ui/src/pages/instance/InstancePage.spec.tsx
@@ -2,6 +2,8 @@ import { cloneDeep, set } from 'lodash'
 import React from 'react'
 import { BrowserRouter } from 'react-router-dom'
 import { instance, mock } from 'ts-mockito'
+import { faker } from '@faker-js/faker'
+import { http, HttpResponse } from 'msw'
 
 import { waitFor, within } from '@testing-library/react'
 import {
@@ -11,7 +13,9 @@ import {
   act,
   mockStore,
   initialStateDefault,
+  getMswURL,
 } from 'uiSrc/utils/test-utils'
+import { mswServer } from 'uiSrc/mocks/server'
 import { resetKeys, resetPatternKeysData } from 'uiSrc/slices/browser/keys'
 import { setMonitorInitialState } from 'uiSrc/slices/cli/monitor'
 import { setInitialPubSubState } from 'uiSrc/slices/pubsub/pubsub'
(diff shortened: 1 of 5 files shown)


trippyogi:
Addressed Codex feedback: `resetConnectedInstance` now only runs when switching away from a different already-connected DB (`connectedInstanceId !== connectionInstanceId`).

Redis Stack sets `connectedInstance.id` before mounting `InstancePage`, so we no longer clear it mid-route (which was bouncing `ProtectedRoute` back to home).

Added a unit test covering the same-id / no-reset case. `InstancePage.spec`: 10/10 passed.

trippyogi:
Addressed Bugbot findings on `2ac110e`:

1. **Connect failure / loading** — `loadInstanceData` now runs on both connect success and failure so `connectedInstance.loading` settles (Vector Search no longer spins forever).
2. **Unused import** — removed `getRecommendations` from `InstancePage.spec.tsx`.

`InstancePage.spec`: 10/10 passed.

redisinsight/ui/src/pages/instance/InstancePage.tsx
```diff
@@ -75,12 +78,29 @@ const InstancePage = ({ routes = [] }: Props) => {
...
+    // connectedInstance.loading settles and Vector Search does not spin forever.
+    dispatch(
+      checkConnectToInstanceAction(
+        connectionInstanceId,
+        loadInstanceData,
+        loadInstanceData,
+        Boolean(connectedInstanceId) &&
+          connectedInstanceId !== connectionInstanceId,
+      ),
+    )
```
VaskoAtanasovRedis: Making this unconditional regresses the existing entry paths though. The DB picker (`databases-list/methods/handlers.ts`), Redis Stack router (`RedisStackRoutes.tsx`), and the navigation popover (`InstancesList.tsx`) all dispatch `checkConnectToInstanceAction` *before* navigating and set `connectedInstance.id` in their success callbacks - so with this change every normal DB open issues a second `/connect`. That endpoint is not cheap on the backend (`DatabaseConnectionService.connect`): it re-probes modules and server version, rewrites the DB record, runs INFO + CLIENT LIST, re-runs recommendation checks, and emits a connection analytics event - all doubled per open, and the analytics get double-counted. It also serializes the four data fetches behind the extra round-trip on every navigation, not just deep links.

Since those paths all arrive with `connectedInstance.id` already equal to the route id, you can keep the fix and skip the redundant call:

```suggestion
    if (connectedInstanceId === connectionInstanceId) {
      loadInstanceData()
    } else {
      dispatch(
        checkConnectToInstanceAction(
          connectionInstanceId,
          loadInstanceData,
          loadInstanceData,
          Boolean(connectedInstanceId),
        ),
      )
    }
```

redisinsight/ui/src/pages/instance/InstancePage.spec.tsx
```diff
@@ -157,22 +164,49 @@ describe('InstancePage', () => {
...
+      resetConnectedInstance(),
       ...resetContextActions,
       clearExpertChatHistory(),
       setConnectivityError(null),
       setAppContextConnectedInstanceId(INSTANCE_ID_MOCK),
       setDbConfig(undefined),
     ]
 
-    expect(store.getActions().slice(0, expectedActions.length)).toEqual(
+    expect(testStore.getActions().slice(0, expectedActions.length)).toEqual(
```
VaskoAtanasovRedis, line 206: The reset / no-reset branches are well pinned (nice negative test for the Redis Stack case), but the behavior this PR exists to create ended up unasserted: the old expected-actions list contained `setConnectedInstance`, `getDatabaseConfigInfo`, `setConnectedInfoInstance`, and `getRecommendations`, and after this rework those only fire inside the connect success callback - which lands *after* the `slice(0, expectedActions.length)` window, so nothing checks them anymore. Deleting `loadInstanceData` (or passing `undefined` for both callbacks) keeps the whole suite green.

Could you add three tests to lock the new flow in:

1. **Success path** - render with the default store and `await waitFor(() => expect(store.getActions()).toEqual(expect.arrayContaining([setConnectedInstance(), getDatabaseConfigInfo(), setConnectedInfoInstance(), getRecommendations()])))`. The MSW handler for `/databases/:id/connect` already returns 200, so the callback runs in tests.
2. **Fail path** - override the connect handler with a 503 via `mswServer.use(...)` (same pattern as `instanceConnectionLost.spec.tsx`) and assert `setDefaultInstanceFailure` is dispatched *and* the four fetches still fire - this is the guarantee of commit 0ce6401, currently uncovered.
3. **Fresh deep link** (the literal #6080 state) - seed `connectedInstance.id = ''`, spy on the connect endpoint (same `jest.fn()` MSW-spy pattern as `getDatabasesApiSpy`) and assert it's called once and `resetConnectedInstance` is *not* dispatched.

Two smaller things while you're in here: the positive assertion in the same-id test (`toContainEqual(setDefaultInstance())`) doesn't actually prove the connect flow ran - `fetchConnectedInstanceAction` dispatches `setDefaultInstance()` too, so a full revert to the old direct-fetch code still passes it; asserting `setDefaultInstanceSuccess()` or the connect spy pins it properly. And `'prevId'` could come from faker per our testing conventions.

(If you adopt the same-id gate suggested on `InstancePage.tsx`, test 1 doubles as its coverage - same-id renders should then assert the fetches fire *without* a connect call.)

redisinsight/ui/src/pages/instance/InstancePage.tsx
```diff
@@ -75,12 +78,29 @@ const InstancePage = ({ routes = [] }: Props) => {
...
+    // Only reset when switching away from a different connected DB.
+    // Redis Stack already set connectedInstance.id before routing here;
+    // resetting would clear it and ProtectedRoute would bounce to home.
+    // Always load instance data after connect attempt (success or fail) so
+    // connectedInstance.loading settles and Vector Search does not spin forever.
+    dispatch(
+      checkConnectToInstanceAction(
+        connectionInstanceId,
+        loadInstanceData,
+        loadInstanceData,
```
VaskoAtanasovRedis: Running loadInstanceData on the fail path is a good catch, but the VS page can still load forever: if the GET /databases/:id inside fetchConnectedInstanceAction itself fails (e.g. a stale or invalid id in the url), its catch never sets connectedInstance.loading back to false.

I think we can add a one line guard against this inside that catch (slices/instances/instances.ts, ~line 759):
```
dispatch(setConnectedInstanceFailure())
```

redisinsight/ui/src/pages/instance/InstancePage.tsx
trippyogi: Good catch on the double /connect - adopted the same-id gate so picker / Redis Stack / nav keep a single connect, and deep links still refresh modules. Included in 2f3b6bf.

redisinsight/ui/src/pages/instance/InstancePage.spec.tsx
trippyogi: Added the three coverage cases (success / fail / fresh deep link), tightened the same-id path to assert no connect call + the four fetches, and switched prevId to faker. Also in 2f3b6bf.

redisinsight/ui/src/pages/instance/InstancePage.tsx
trippyogi: Agreed - i added setConnectedInstanceFailure() in the fetchConnectedInstanceAction catch so connectedInstance.loading settles on a bad/stale id. Also in 2f3b6bf.

VaskoAtanasovRedis:
Hey @trippyogi , thanks for addressing the comments so quickly! The lint checks seem failing, so I'll wait for you to resolve them, but generally the change now looks good to me, so once we resolve and CI is green I would be happy to get this merged in

trippyogi:
@VaskoAtanasovRedis No problem, i fixed the prettier failures in 89397ce, and it should be ready to run the lint check again shortly!

VaskoAtanasovRedis (approved):

VaskoAtanasovRedis:
Thank you for the time and effort here @trippyogi - changes look good to me, we'll get them merged and they'll be a part of our next release

Issue #5357, opened 2026-01-02
[Bug]: TS commands like RANGE, REVRANGE do not work in workbench with ACL


sagarringe:
**Preconditions** (Any important steps we need to know)
Create readonly user with access to +@read ~*  
Tried with explicit access too
+TS.RANGE, TS.INFO

**Steps to reproduce** (How to reproduce what you found step by step)

- Create a connection using a read access user
- Open Workbench
- RUN the command. eg `TS.RANGE "{FINNHUB_PRICES_20260102:RETURN1DAY}:LQ9WBB-R" - +`

**Actual behavior** (A short description of what you found)
Workbench shows the error "Time series data structure is not available for this database"
The command works in the CLI.
Same Command works for a full access user in the workbench.

**Expected behavior** (A short description of what you expected to find)
- The command should work as it does in a full access user (read/write) without any additional access.
- It should not show an incorrect error page.

**Screenshots** (Paste or drag-and-drop a screenshot or a link to a recording)

<img width="1419" height="652" alt="Image" src="https://github.com/user-attachments/assets/44ed67f2-2755-433c-a460-b49696a3c330" />

**Additional context** (Operating system, version of Redis Insight, Redis database version, Redis module version, database type, connection type, logs, or any other information)
Tested with Redis Cloud Database for versions 8.2 and 7.4
MacOS
RedisInsight Version 2.70.1

ArtemHoruzhenko:
@sagarringe Thank you for reporting this! This is a known issue and we will prioritize fixing it. In the meantime, there is a workaround: the user must have access to `info` (Modules section) or `module list` command, and everything should work as expected.

PR #6232, opened 2026-07-17
fix(workbench): do not block TimeSeries under ACL module detection failure

trippyogi:
﻿## Summary
- Skip Workbench ModuleNotLoaded when a module command already succeeded (ACL users can run TS.RANGE while module discovery fails).
- Fall back to HELLO for module detection when MODULE LIST / COMMAND INFO are unavailable; prefer HELLO over misparsed INFO Modules lines.
- Add ts.range / ts.revrange to TimeSeries probe commands.

Refs #5357

## Test plan
- [x] Unit: database-info.provider specs (HELLO fallback + collapsed INFO modules)
- [x] Local Bugbot: clean after HELLO/INFO and ModuleNotLoaded fixes
- [x] ACL Redis user (+@read + TS.RANGE, no MODULE LIST / COMMAND INFO): Workbench TS.RANGE returns data (not Time series unavailable)
- [ ] Full-access user: TimeSeries / Search Workbench still works
- [ ] Plain Redis (no modules): failed TS.* still shows ModuleNotLoaded guidance

---

> [!NOTE]
> **Medium Risk**
> Touches database module discovery (HELLO fallback) and Workbench error rendering; behavior changes for ACL-restricted connections but is guarded by success status and new unit tests.
> 
> **Overview**
> Fixes Workbench showing **Time series unavailable** for ACL users who can run `TS.RANGE` but whose connection has no detected modules.
> 
> **API:** After `MODULE LIST` fails, module discovery still tries `COMMAND INFO`; if that yields nothing, it now falls back to **`HELLO`** and maps modules from the reply (shared `mapToAdditionalRedisModules`). TimeSeries probe commands now include **`ts.range`** and **`ts.revrange`**.
> 
> **UI:** **`CommonErrorResponse`** skips **`ModuleNotLoaded`** when the execution already **succeeded**, so missing module metadata does not override a good reply. Group mode passes the full execution object (not only the raw Redis string) into that helper. Tests cover successful vs failed `TS.RANGE` in group mode.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 233b036e5538b77cd7c5fbda81c5cfd2ae0953ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisinsight/api/src/modules/database/providers/database-info.provider.ts
+++ b/redisinsight/api/src/modules/database/providers/database-info.provider.ts
@@ -137,7 +137,9 @@ export class DatabaseInfoProvider {
 
   /**
    * Determine database modules from HELLO or INFO response when MODULE LIST
-   * and COMMAND INFO are unavailable (e.g. restricted ACL users)
+   * and COMMAND INFO are unavailable (e.g. restricted ACL users).
+   * Prefer HELLO: getInfo()'s INFO parser collapses duplicate `module:` lines into
+   * a single { module: "name=...,ver=..." } entry, which is not a usable module list.
    * @param client
    * @private
    */
@@ -145,8 +147,17 @@ export class DatabaseInfoProvider {
     client: RedisClient,
   ): Promise<AdditionalRedisModule[]> {
     try {
-      const info = await client.getInfo();
-      const rawModules = this.normalizeModulesFromInfo(info?.modules);
+      let rawModules = await this.getModulesFromHello(client);
+
+      if (!rawModules.length) {
+        try {
+          const info = await client.getInfo();
+          rawModules = this.normalizeModulesFromInfo(info?.modules);
+        } catch {
+          // INFO may be denied
+        }
+      }
+
       if (!rawModules.length) {
         return [];
       }
@@ -168,20 +179,50 @@ export class DatabaseInfoProvider {
     }
   }
 
+  private async getModulesFromHello(
+    client: RedisClient,
+  ): Promise<{ name: string; ver?: number }[]> {
+    try {
+      const helloResponse = (await client.call(['hello'], {
+        replyEncoding: 'utf8',
+      })) as string[];
+      const helloInfo = convertArrayReplyToObject(helloResponse);
+      const modules = Array.isArray(helloInfo.modules)
+        ? helloInfo.modules.map((module) =>
+            Array.isArray(module)
+              ? convertArrayReplyToObject(module)
+              : module,
+          )
+        : helloInfo.modules;
+
+      return this.normalizeModulesFromInfo(modules);
+    } catch {
+      return [];
+    }
+  }
+
   private normalizeModulesFromInfo(modules: unknown): { name: string; ver?: number }[] {
     if (!modules) {
       return [];
     }
     if (Array.isArray(modules)) {
-      return modules;
+      return modules.filter(
+        (module) =>
+          module &&
+          typeof module === 'object' &&
+          typeof (module as { name?: unknown }).name === 'string',
+      ) as { name: string; ver?: number }[];
     }
     if (typeof modules === 'object') {
-      return Object.entries(modules as Record<string, unknown>).map(
-        ([name, ver]) => ({
-          name,
-          ver: parseInt(String(ver), 10) || undefined,
-        }),
-      );
+      const entries = Object.entries(modules as Record<string, unknown>);
+      // INFO # Modules collapses to { module: "name=...,ver=..." } — not usable
+      if (entries.some(([key]) => key === 'module')) {
+        return [];
+      }
+      return entries.map(([name, ver]) => ({
+        name,
+        ver: parseInt(String(ver), 10) || undefined,
+      }));
     }
     return [];
   }
--- a/redisinsight/api/src/modules/database/providers/database-info.provider.ts
+++ b/redisinsight/api/src/modules/database/providers/database-info.provider.ts
@@ -81,7 +81,11 @@ export class DatabaseInfoProvider {
           : undefined,
       }));
     } catch (e) {
-      return this.determineDatabaseModulesUsingInfo(client);
+      const fromCommands = await this.determineDatabaseModulesUsingInfo(client);
+      if (fromCommands.length) {
+        return fromCommands;
+      }
+      return this.determineDatabaseModulesUsingHelloOrInfo(client);
     }
   }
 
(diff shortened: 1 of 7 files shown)


pawelangelow:
@trippyogi I haven't had a chance to review the PR in detail yet, but I noticed that the `eslint` and `typecheck` checks are currently failing. Could you please address those issues while the PR is waiting for review? Thanks!

trippyogi:
> @trippyogi I haven't had a chance to review the PR in detail yet, but I noticed that the `eslint` and `typecheck` checks are currently failing. Could you please address those issues while the PR is waiting for review? Thanks!

Thanks @pawelangelow. fixed the eslint / typecheck failures in 6920994.

trippyogi:
@pawelangelow Follow-up: a couple more type-check issues showed up after 6920994. those are fixed in 75eae6c, and checks are green now. Ready whenever you have time to review.

redisinsight/api/src/modules/database/providers/database-info.provider.ts
```diff
@@ -131,6 +129,113 @@ export class DatabaseInfoProvider {
...
+    if (typeof modules === 'object') {
+      const entries = Object.entries(modules as Record<string, unknown>);
+      // INFO # Modules collapses to { module: "name=...,ver=..." } — not usable
+      if (entries.some(([key]) => key === 'module')) {
+        return [];
+      }
+      return entries.map(([name, ver]) => ({
+        name,
+        ver: parseInt(String(ver), 10) || undefined,
+      }));
```
pawelangelow: This `entries.map` is unreachable. Same point the Codex bot raised just above, I think.

The parser chain collapses the section before it ever gets here:

```ts
// src/utils/redis-reply-converter.ts:10
result[section.toLowerCase()] = convertMultilineReplyToObject(values[index].trim())

// src/modules/redis/utils/reply.util.ts:70
obj[key] = keyValuePair.join(valueSeparator)   // duplicate keys overwrite
```

Every line in `# Modules` is keyed `module`, so N modules collapse to a single entry and only the last survives. That entry always trips the `key === 'module'` guard above, so this `return` never runs. If anything ever did reach it, it would turn arbitrary INFO field keys into module names, and since they do not match `SUPPORTED_REDIS_MODULES` they pass straight through `mapToAdditionalRedisModules` (`name: supportedName ?? moduleName`) into the persisted module list.

The fix falls out of the restructure in my comment on `getModulesFromHello` above: `normalizeModulesFromInfo` disappears entirely, along with the INFO fallback at lines 146-153. HELLO already covers this case, so there is nothing left for the INFO path to add.

redisinsight/api/src/modules/database/providers/database-info.provider.ts
```diff
@@ -131,6 +129,113 @@ export class DatabaseInfoProvider {
...
+        ? helloInfo.modules.map((module) =>
+            Array.isArray(module) ? convertArrayReplyToObject(module) : module,
+          )
+        : helloInfo.modules;
+
+      return this.normalizeModulesFromInfo(modules);
+    } catch {
+      return [];
+    }
+  }
```
pawelangelow: Two small things here.

Use the existing DTO instead of the inline shape: `{ name: string; ver?: number }` is `RedisDatabaseModuleDto` (`dto/redis-info.dto.ts:125-137`). Same at lines 171 and 213.

And once `normalizeModulesFromInfo` goes (see my other comment), this method only needs the array branch: map each entry through `convertArrayReplyToObject`, then filter to the ones with a string `name`.

FYI, nothing to action here: this overlaps with `RedisClient.getRedisHelloResponse()` (`redis.client.ts:298-311`), which does the same parsing. We will consolidate that on our side rather than ask you to touch the shared client.

redisinsight/ui/src/components/query/query-card/QueryCardCommonResult/components/CommonErrorResponse/CommonErrorResponse.tsx
```diff
@@ -91,7 +91,19 @@ const CommonErrorResponse = (id: string, command = '', result?: any) => {
       CommandExecutionStatus.Fail,
     )
   }
-  const unsupportedModule = checkUnsupportedModuleCommand(modules, commandLine)
+
+  const isSuccessfulResult = Array.isArray(result)
+    ? result.length > 0 &&
+      result.every((item) => item?.status === CommandExecutionStatus.Success)
+    : result?.status === CommandExecutionStatus.Success
```
pawelangelow, line 97: `every` is stricter than the invariant you are asserting, and it reintroduces the reported bug in a narrower case.

For a repeat command such as `5 TS.RANGE ts:prices - +`, `result` has five entries. One `Fail` makes `every` false, so `isSuccessfulResult` is false and the card renders the `ModuleNotLoaded` upsell page instead of the four good results plus the real error. What you want to express is "this module demonstrably works", which is `some`.

Dropping the `result.length > 0 &&` guard is safe, since `[].some(...)` is already `false` and the empty-array case still falls through to the module check exactly as before.

```suggestion
  const isSuccessfulResult = Array.isArray(result)
    ? result.some((item) => item?.status === CommandExecutionStatus.Success)
    : result?.status === CommandExecutionStatus.Success
```

I ran this locally: `QueryCardCliGroupResult` and `QueryCard` suites pass (20 tests), and all of `components/query` passes (15 suites, 91 tests), eslint clean.

Second ask on the same lines: this `unsupportedModule` assignment used to be unconditional, and the PR turns it into a two-way branch. Your new test covers the branch that suppresses `ModuleNotLoaded`; the branch that still renders it has no test, and I checked that nothing else in the suite covers that decision either (`ModuleNotLoaded.spec.tsx` only exercises the component given props). So a later refactor that suppresses the page unconditionally would go green.

Could you add the other half to `QueryCardCliGroupResult.spec.tsx`, next to your new test? This version passes on your branch, and I confirmed it fails when the guard is made unconditional:

```tsx
it('should show ModuleNotLoaded for failed TS.RANGE in group mode', () => {
  const mockResult = [
    {
      response: [
        {
          id: 'id',
          command: 'TS.RANGE ts:prices - +',
          response: 'ERR unknown command',
          status: CommandExecutionStatus.Fail,
        },
      ],
      status: CommandExecutionStatus.Fail,
    },
  ] as unknown as Props['result']
  render(
    <QueryCardCliGroupResult {...instance(mockedProps)} result={mockResult} />,
  )

  expect(screen.getByTestId('module-not-loaded-content')).toBeInTheDocument()
})
```

pawelangelow (commented):
Thanks for the thorough work on this @trippyogi, and for chasing down the CI failures.

I verified the repro first: reverting the two UI files to `main` makes your new group-mode test fail with exactly the `module-not-loaded-content` node from the issue screenshot. Proper regression guard, appreciated. The two-layer approach is right too, since "the reply succeeded, so the module is loaded" holds even where module discovery cannot recover.

Three inline comments below. The two blocking ones are the unreachable INFO branch in `database-info.provider.ts` and the `every` check in `CommonErrorResponse.tsx`.

One non-blocking question: the issue was reported against Redis Cloud 8.2 and 7.4, where connections are proxied. Did you test HELLO against an actual Cloud database with the reporter's ACL? If Enterprise's HELLO returns an empty `modules` array there, the API half is inert for the original reporter and only the UI half fixes the symptom. Also worth closing out the two unchecked boxes in the test plan.

Nits, take them or leave them:

- A few of the new comments carry ticket references and change narration (`(see #5357)`, "previously fooled the success check"). Our convention keeps that context in the commit message or PR description instead. See `.ai/skills/code-quality/SKILL.md`.
- In `QueryCardCliGroupResult.tsx`, `item` already carries `{ id, command, response, status }`, so `CommonErrorResponse(item.id, item.command, item)` works without the wrapper array.

trippyogi:
Thanks @pawelangelow — addressed the blocking items in 233b036:

- `CommonErrorResponse` now uses `some()` so a mixed repeat (`5 TS.RANGE…`) keeps ModuleNotLoaded suppressed when any reply succeeded
- Added the group-mode fail-path test you suggested
- Removed the unreachable INFO/`normalizeModulesFromInfo` path; HELLO-only fallback remains, typed with `RedisDatabaseModuleDto`
- Applied the nits (pass `item` directly; drop ticket-narration comments)

On Cloud HELLO: we verified against local Redis Stack with a restricted ACL (`+@read` + `TS.RANGE`, no `MODULE LIST` / `COMMAND INFO`), not against a Redis Cloud DB with the reporter’s ACL. If Enterprise HELLO returns an empty `modules` array there, the UI success-gate is what fixes the symptom for that path. Happy to re-check on Cloud if you have a repro DB.

Test plan: the plain-Redis ModuleNotLoaded case is covered by the new group-mode fail unit test; I’ll close out the remaining manual boxes after one more full-access pass on Stack.

pawelangelow (approved):
Merging this - thanks @trippyogi! Nice work on the two-layer fix! Thanks for the patience, too.

PR #6505, opened 2026-09-15
RI-8398 Pass plugin visualization config via a JSON element

dantovska:
# What

Passes the Workbench plugin visualization config into the render iframe through a `<script type="application/json">` element read with `JSON.parse`, instead of interpolating it into inline script source. Generated attribute values (stylesheet `href`, body class, plugin script `src`) are quoted and escaped, and the plugin module is loaded from the parsed config. This keeps config values containing characters like backticks, `${`, or quotes as plain data rather than letting them break the generated document.

# Testing

Run `node node_modules/.bin/jest redisinsight/ui/src/plugins/pluginImport.spec.ts -c jest.config.cjs` — added coverage for config values with special characters. `lint:ui` clean.

---

Refs #RI-8398

---

> [!NOTE]
> **High Risk**
> Changes how untrusted Redis-sourced strings are embedded in plugin iframes; incorrect escaping could reintroduce XSS, though the approach is standard and heavily tested.
> 
> **Overview**
> **Hardens Workbench plugin iframe HTML** so server-supplied config (e.g. Redis module names) cannot break out of generated markup or run as inline script.
> 
> Plugin config is no longer embedded in template literals inside executable `<script>` blocks. It is written to a **`<script type="application/json">`** (`ri-plugin-config`), and `importPluginScript` reads it with `JSON.parse` at bootstrap. Stylesheet `href`, body `class`, and module `src` are **double-quoted and HTML-escaped**; JSON in the script tag escapes `<` as `\u003c` so `</script>`-style payloads cannot close the element.
> 
> The bootstrap loads the plugin via `import(riConfig.scriptSrc)` and builds error text from `riConfig.scriptPath` using values from parsed state, not string interpolation. **`pluginImport.spec.ts`** adds regression tests (malicious module names, attribute breakouts, sandboxed script evaluation) and drops the file from **`.tscheck.rec.json`** now that those specs type-check.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 95bf9b3a6915a920c9033b355060c39c6496fa9b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ArtemHoruzhenko (approved):
lgtm

